PR #122, opened 2014-12-17
Travis CI integration

TobiX:
Just having Travis CI build every commit is a huge plus, so I think we should add support. Here is a working configuration. Travis is able to create a binary. It probably wouldn't run, since it mixes different libav-versions in one binary (system OpenCV and newer libav from ppa), but since Travis uses Ubuntu 12.04, I couldn't come up with anything better at the moment.

Maybe I build my own ppa later, but don't hold your breath ;)


nieknooijens:
travis CI cannot find the libepoxy library, so that broke the build :-1:

TobiX:
I'll probably try another CI service that has newer base systems, I'm not comfortable with this. Closing for now.

gaborcsardi:
@TobiX https://semaphoreapp.com/ has recent ubuntu, two threads, good response times and you can ssh to the node. (I am not affiliated with them in any way, this is just my experience.)

PR #131, opened 2015-01-20
Tools rework (only a bit)

TobiX:
I wanted to change more with the tools, but I don't have that much time currently, so let's merge this back:

- Don't create notes.txt in ss_extract, since Performous doesn't need them.
- Replace ImageMagick with our own image saving code. This creates PNGs instead of JPEGs, if anyone cares enough, I'll add JPEG writing code


nieknooijens:
looks pretty OK to me.

tapio:
Nice to get rid of ImageMagick! I think png is fine for covers, but jpg would probably be preferable for larger backgrounds (no idea if backgrounds are even extracted though). I'm not a huge fan of the tiny common directory, but I'm fine with merge regardless.

TobiX:
Yeah, the tiny common directory is a bit ugly, but I had no other idea how to do it. And I really want to avoid cross-dependencies between tools and the main program, and I think this is the best way to keep a clean cut...

PR #134, opened 2015-02-20
Mingw build fix

TobiX:
Recently, building with MXE fails, because of name conflicts. I really don't know why this only surfaced recently:

* CALLBACK is a define deep in some windows includes, so you eventually get bitten by it
* Chord is an age-old Windows GDI function, don't know why this didn't triggered earlier (https://technet.microsoft.com/en-us/scriptcenter/dd183428%28v=vs.105%29.aspx)

I just renamed the symbols, there might be a better way (namespaces?) - Any thoughts?


nieknooijens:
finally it builds! and best of all, it builds WITH webcam! openCV is fixed in mxe ;-)

PR #136, opened 2015-02-23
Path to libboost_thread_win32-mt.dll fixed ~Windows Build

Baklap4:
When running make-performous.sh it was looking in the wrong folder. Changed it to look into the right folder where the file actually exists


nieknooijens:
I noticed that too, running cmake now works fine without any extra command-line parameters.

TobiX:
Oh, really? That's even better!

PR #137, opened 2015-03-09
Fatal error dialog

TobiX:
Show a meaningful error dialog to the user in case of fatal errors. This should stop errors of the type "Performous does not start" since the user gets some feedback...

Naturally we can only tell the user to look at the log file after logging is initialized ;)

### DIFF
--- a/game/log.cc
+++ b/game/log.cc
@@ -1,6 +1,7 @@
 #include "log.hh"
 
 #include "fs.hh"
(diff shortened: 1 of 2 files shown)


nieknooijens:
Nice! unfortunately performous doesn't crash here so I can't really test it ;-)

TobiX:
You can force the error by inserting errors in one of the shaders ;)

PR #143, opened 2015-04-13
Fix font loading

TobiX:
These fixes different font issues:

 - Restores loadFonts so all fonts from themes are loaded
 - Makes sure to use the freetype/fontconfig backend on all platforms (and not native rendering on Win32 and Mac OS X), so our "private" fonts are really used.
 - Fixes listFiles (which didn't do what it's docstring suggested)
 - Parse font-size from SVG CSS as integer instead of double, since the double parsing rules are bonkers.
 - Does some code cleanup and adds some logging to the font loading.

### DIFF
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -226,11 +226,14 @@ fs::path findFile(fs::path const& filename) {
 }
 
 Paths listFiles(fs::path const& dir) {
-	if (dir.is_absolute()) throw std::logic_error("findFile expects a folder name without path.");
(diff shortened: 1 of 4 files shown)


Lord-Kamina:
I have confirmed that:

a) The error is indeed related to libc++ (compiled trunk previous to this, using libstdc++ and it worked fine).
b) These commits fix the libc++ builds.
So I'll be merging this. 

Which means, we now have a 1.0+ functioning build, compiled with Clang (more modern, faster, much easier to setup in OSX, by virtue of being the default) and linked against libc++.

Yay!

PR #152, opened 2015-07-20
Fail on shader creation error

TobiX:
These changes seem to be reasonable ;)


nieknooijens:
these ARE completely reasonable....

PR #155, opened 2015-08-21
check for UTF8-BOM while guessing if type == TXT

frankhedecke:
This patch fixes an issue where a TXT-song-file with UTF-8 BOM will be ignored, because the check (type == TXT) fails.

Now there is a dedicated check for the UTF-8 BOM while guessing the type.

### DIFF
--- a/game/songparser-txt.cc
+++ b/game/songparser-txt.cc
@@ -15,7 +15,11 @@ namespace {
 
 /// 'Magick' to check if this file looks like correct format
 bool SongParser::txtCheck(std::vector<char> const& data) const {
-	return data[0] == '#' && data[1] >= 'A' && data[1] <= 'Z';
+	if (data[0] == '#' && data[1] >= 'A' && data[1] <= 'Z') return true;
+	// ignore UTF-8 BOM while checking
+	else if (data[0] == '\xEF' && data[1] == '\xBB' && data[2] == '\xBF' &&
+				data[3] == '#' && data[4] >= 'A' && data[4] <= 'Z') return true;
+	else return false;
 }
 
 /// Parse header data for Songs screen


yoda-jm:
Hi, shouldn't we check the size of data to make sure it contains enough data ?

frankhedecke:
Dear yoda-jm,
in game/songparser.cc : 63 there is a size check. Afterwards songparser is guessing the type (NONE, TXT, XML, INI, SM).

nieknooijens:
does this also fix those CP1220 encoded songs which have "?<<ï" in front of them and are therefore skipped?

frankhedecke:
I don't know. Never tried.

TobiX:
@nieknooijens No, it doesn't. I would suggest refactoring the reading logic in the SongParser::SongParser constructor: Currently it works like this:

1. Read file
22. Detect type
3. Convert to UTF-8
4. Parse content

IMHO, it should be changed to:

1. Read file
2. Convert to UTF-8
3. Detect type
4. Parse content

That would fix BOM problems in all file types (not only TXT)...

Bonus refactoring: SongParser into a proper class structure...

nieknooijens:
Extra bonus refactoring:
group all songfiles with the same name in the same directory which point to the same MP3 into a single song object, this would enable simultaneous DDR/Guitar/Duet-vocals in the future, I've been struggling with that for a while now (see songmerge branch) and while I can get it to merge a few files into a single object, the game crashes when trying to play that file, but in memmory it does the correct thing (according to GDB)

frankhedecke:
What will happen with this pull-request?

TobiX:
I will propose a more complete fix today or tomorrow.

TobiX:
@nieknooijens - Small steps please. Let's fix the BOM/UTF issue first and tackle the song merging later. (Actually, I don't think they even touch the same code at all.)

TobiX:
@frankhedecke Could you check if my new branch (pull request #158) works for you?

frankhedecke:
@TobiX I checked it and it works.

TobiX:
Obsoleted by #158

PR #156, opened 2015-08-25
Bunch of GL fixes

TobiX:
Some small GL fixes (mostly in respons to issues raised in #148):

- Check for GL extension for 3D stero shader instead of checking for the GL version. It seems some Intel cards say they are OpenGL 3.3 an still are missing this extension. The opposite should never happen (GL < 3.3 and support for this extension)
- This might be a bit controversial: Force high-performance GPU on dual-GPU-systems (Nvidia Optimus & AMD PowerExpress) - If anyone is interested, I can go into detail why I think this is a good idea.
- Do a compile check to only compile against libepoxy >= 1.2 - There is a nasty bug in libepoxy < 1.2 that makes Performous crash on Linux with nvidia or fglrx.

### DIFF
--- a/cmake/Modules/FindLibEpoxy.cmake
+++ b/cmake/Modules/FindLibEpoxy.cmake
@@ -25,5 +25,6 @@ find_library(LibEpoxy_LIBRARY
 
 set(LibEpoxy_PROCESS_INCLUDES LibEpoxy_INCLUDE_DIR)
 set(LibEpoxy_PROCESS_LIBS LibEpoxy_LIBRARY)
+set(LibEpoxy_VERSION ${LibEpoxy_PKGCONF_VERSION})
 libfind_process(LibEpoxy)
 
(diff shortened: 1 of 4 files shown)


yoda-jm:
LGTM, can you add little comment next to the changes in code (it will avoid someone else to revert this)

TobiX:
@yoda-jm For which changes? The libepoxy one?

yoda-jm:
I think the 3 changes can deserve a little comment to know why we are doing this

TobiX:
I added some comments to two of the changes. I couldn't find a short description for the dual-GPU changes, but I consider it the logical thing to do for a game. Please correct me if I'm wrong.

nieknooijens:
for the dual GPU stuff, I have one of those and on linux bumblebee works alot better than Nvidia's prime-switch since the nvidia-prime package just renders everything on the nvidia without any power management, basically frying your laptop battery. bumblebee runs a second x server on the discrete card and uses a ram-bridge to copy-over the frames to the intel's x-server.
bumblebee also put's the nvidia in power-save mode when on battery.

TobiX:
@nieknooijens Well, it won't affect you on Linux then. It currently only forces the discrete GPU on Windows, since there is a documented way for that. See the source on how that works.

nieknooijens:
well i'm working on my desktop for a change and guess what? I had issue #148! pulled this branch and it fixed it, so I propose we merge this one into master.

TobiX:
Nobody was against merging this, so I went ahead and merged ;)

PR #158, opened 2015-09-03
Songloading refactor

TobiX:
Do some light refactoring of the song loading. This unifies the fallback encoding and BOM handling code, so detection should be correct across all text file formats that we support.

I'm currently doing a test build for Windows, to see if the last commit breaks anything, but I don't expect it. As far as I understand it, this is only a problem when combining libraries compiled with different runtimes (for example, compiling Performous with Visual Studio and linking against a glibmm compiled with GCC)


TobiX:
Exception handling works fine on Windows. I consider this merge-able!

PR #163, opened 2015-09-25
Translation (90%)

skyli91:
Some minor things left, every visible thing is translated. Did not find "vocals" at scoring and "perfect!" at freestyle notes.


TobiX:
Thanks! Updated. It seems those strings might not be translatable... See for example game/dancegraph.cc, line 49ff and game/screen_sing.cc, line 637...

PR #164, opened 2015-09-25
Microphone colors update

skyli91:


TobiX:
Stupid question: Why? Since the colour names are used in the settings-file, this change might make users loose their settings for one microphone. (OTOH, you only changed colours that wer added after 1.0, so we might still use this...)

TobiX:
For comparison:
Old:
performous-colours-old
New: 
performous-colours-new

nieknooijens:
The new one has 2 colors that look alike, (4 and 6) so I prefer the old one, it's much more clear!

skyli91:
Hmm. I wanted to make an orange one from the 6. I was not able to test it (I tried hard to comply on linux with the tutorial but i failed :( ) so i just used photoshop. Sorry, for me, it looked like this in Ps: 

So what i noticed is the original 3 and 5 is too similar. If someone could change only the 6th to orange it would be perfect :)

What I wanted to do:
12026653_10208092821383477_605640883_n

skyli91:
Could someone check it now? :) (I'm not able to bould)

TobiX:
new-colors2
Like this?

skyli91:
Yes, this one looks great! :) :+1:

Baklap4:
When will this be merged? It's no biggy but as of Nov 23, 2015 it's been approved by requester. It's already April 11, 2016 now..

TobiX:
Could you rebase and squash your commits, please? Since I have no real objections against one or the other (except the config breakage..) I will merge when I get around to it...

skyli91:
Now i'm able to test on ubuntu so i've played with it again as the pink one is rather white. Everyona said "who is with the white?" etc :D so what about this? 
image
Player 9 and 10 could be white and grey by the way :D

PR #167, opened 2015-11-08
ffmpeg2.9 transition

apoleon:
This commit replaces deprecated ffmpeg 2.8 functionality which will be removed
in 2.9 and will thus cause a build failure. The patch works with 2.8 and 2.9.

This is Debian bug https://bugs.debian.org/803853

Patch was provided by Andreas Cadhalpun

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -116,7 +116,7 @@ void FFmpeg::open() {
 		height = cc->height;
 		m_swsContext = sws_getContext(
 		  cc->width, cc->height, cc->pix_fmt,
-		  width, height, PIX_FMT_RGB24,
+		  width, height, AV_PIX_FMT_RGB24,
 		  SWS_POINT, nullptr, nullptr, nullptr);
 		break;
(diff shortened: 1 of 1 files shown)


TobiX:
Do you know sine when AV_PIX_FMT_RGB24 is available in ffmpeg? I'd like to fix our cmake checks to require a minimal version... Since we use libavresample since a while, I suppose we require at least 0.11...

apoleon:
Am 10.11.2015 um 00:36 schrieb Tobias Gruetzmacher:
> Do you know sine when AV_PIX_FMT_RGB24 is available in ffmpeg? I'd like
> to fix our cmake checks to require a minimal version... Since we use
> libavresample since a while, I suppose we require at least 0.11...

Yes, you can use:
 * for AVPixelFormat, AV_PIX_FMT_*:
#if LIBAVUTIL_VERSION_MAJOR > 51
 * for av_frame_alloc/av_frame_free:
#if LIBAVUTIL_VERSION_MAJOR > 52

PR #175, opened 2016-01-14
Add picks for controller GUITAR_GUITARHERO5

arielenter:
I own this hardware and pick was reported as:
<hat hw="0" up="pick_up" down="pick_down" />

Before no pick was configured, but adding that to controllers.xml made pick work. Thank you.


nieknooijens:
Thanks for helping to improve performous!

PR #176, opened 2016-01-15
Add pick for GUITAR_GUITARHERO_XPLORER

arielenter:
I own this guitar too, and pick is register as:

hat hw="0" up="pick_up" down="pick_down"

Also hw="4" and hw="5" are very problematic for menu and song navigation. Both are motion sensor and are very sensitive. I tried to calibrate the controller using jstest-gtk but I was getting the same results. Every time I started the game it was impossible to move through the menu or the songs because the cursor started jumping all over the place.

axis hw="5" could be use for godmode but it is very inaccurate too. I suggest just disable them all along, since the D-pad and the pick can be use to navigate the menus and the songs anyway. If I found any other axis that can be use for godmod I let you know.

### DIFF
--- a/data/config/controllers.xml
+++ b/data/config/controllers.xml
@@ -98,11 +98,12 @@
 				<button hw="4" map="orange" />
 				<button hw="6" map="cancel" />
 				<button hw="7" map="start" />
+				<hat hw="0" up="pick_up" down="pick_down" />
 				<axis hw="1" negative="pick_up" positive="pick_down"/>
 				<axis hw="2" map="whammy" />
 				<axis hw="3" map="godmode"/>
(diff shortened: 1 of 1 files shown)


arielenter:
More than motion sensors, hw="4" and hw="5" register the angle of the guitar. They are analog. hw="4" registers the vertical angle of the guitar. hw="5" seems to register the horizontal one.

nieknooijens:
Well I don't have the hardware for it. I can only check whether there isn't any malicious code being pushed :stuck_out_tongue: Judging the discussion I think it would be justified to comment out those two axis but leave the comments in for those who do want to use those two axes.

PR #178, opened 2016-02-08
Dialog

nieknooijens:
I re-did the dialog to have a more clean look and made it so that it doesn't cover the whole screen anymore, which was particularly annoying when you had a song with broken notes, I know it's mandatory to let the user know, but most of the times it's not a big deal and I just want to continue singing, but the dialog was covering the text!


nieknooijens:
if no response this will be merged next sunday (28-02-2016)

Issue #182, opened 2016-04-11
View time till song number X


Baklap4:
With the webserver now being included in master i'd like to have another feature for the webserver.
Currently you can show the playlist but can't see how long it takes till your song requested is being played.
So within the playlist it'd be nice to see how much time till song number x is going to be played.

Example:

```
0. ABBA - Dancing Queen << current song being played with 5:23 minutes left of time (this one doesn't need to be on the playlist)
1. ABBA - Waterloo 5:23 << time of current playing song (mm:ss) which will count down.
2. ABBA - Mamma Mia 8:10 << time of the current playing song + playtime from previous songs
3. ABBA - Super Trouper 10:07 << current playing song + time of previous 2 songs 
```

Et cetera.

By adding this you can tell people you have to wait for X minutes instead of guessing the time.
It would be nice to have all the time values be counting down as the current song passes by.

If i have time I'll look into this.

nieknooijens:
this is a very often requested feature, unfortunately this isn't going to be an easy fix.
Basically we need to add a function to playlist.cc that checks the length of the songs.
You cannot use the notes for that since they're only loaded on play time so it needs to be extracted from the music file. the music file can be anything! ogg, mp3, wav, etc. currently we just dump the mp3 through libavcodec and it will provide us uncompressed audio, so I hope libavcodec has a way to get a songs duration.
for optimization I'd reccommend to store the duration in the song class so once it's extracted from the audio file it doesn't have to be checked again when it's still in the playlist.

nieknooijens:
ok in theory this should work:
https://github.com/performous/performous/tree/PlaylistDuration
in practice it doesn't... what am I doing wrong? the filepath is correct...

nieknooijens:
![schermafdruk van 2016-04-16 15-44-29](https://cloud.githubusercontent.com/assets/2794331/14581333/5cdb372e-03ea-11e6-9e38-2430afa40232.png)
good news! it's working, but when adding stuff to the playlist I need to decode the whole file to determine it's length, so it brings some lag, at this moment that's not a problem since it's loaded at the start of the playlist screen, and once it's loaded it's in memmory and can be accessed at any time but when you add this to the webserver interface we should make it load asynchroniously to avoid lockup on the main thread.

nieknooijens:
just realized something:
webserver is already a different thread, I can just dump a mutex in the song object and it'll work!

nieknooijens:
[...]

PR #183, opened 2016-04-16
Playlist duration

nieknooijens:
this branch allows you to view how long it takes until song X arrives as described in #182 
The only way to get the duration of an audio file is to completely read it, this puts stress on the filesystem so to make sure we access the filesystem as little as possible I designed it in the following way:
when screenPlaylist comes up (main thread) or the playlist is requested from the web frontend (webserver thread) they call getDurationSeconds() on the song object. this function checks whether the duration is already known and if so, it returns it, if not it opens up ffmpeg and decodes the music file, calculates the duration and stores it in memmory so it doesn't have to be read from disk again when the duration is requested next time.
this way we put as little stress on the system as possible, only fully loading the songs that are actually put in the playlist and only loading them once.

### DIFF
--- a/data/themes/default/index.html
+++ b/data/themes/default/index.html
@@ -260,7 +260,7 @@ filter: progid:DXImageTransform.Microsoft.gradient( startColorstr='#feffff', end
 				}
 				var minutestxt = ("00"+minutes).slice(-2);
 				if (minutes >= 100) {
-				("000"+minutes).slice(-3)
+				minutestxt = ("000"+minutes).slice(-3)
 					
 				}
 				$('#playListView').append('<li data-theme="c" data-icon="info"><a href="javascript:ViewInfo(' + i + ')" data-transition="slide"> #' + count + ' ' + songlist[i].Artist + ' - ' + songlist[i].Title + '      +' + minutestxt + ':' + ("0"+sec).slice(-2) + '</a></li>\n');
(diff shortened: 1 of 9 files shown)


Baklap4:
I've checked the enhancements you did make, it looks great!
A few things to make it look even better would be:

- Time in the format: hh:mm:ss (when an hour is reached. Else mm:ss)
- The next one in line currently shows 00:00 min. It'd be great to see this being the length of the current song being played counting down towards 00:00 min. Which also brings me to the 3rd point
- Counting down the time while the current song is playing.

The 3rd point: 
As the webserver view (within the browser) does can be set to refresh every 10 secs it would be nice to update the time aswell.

TobiX:
Congratulation, you mixed rebase and merge, now every commit is doubled.

nieknooijens:
@TobiX ok, well at least I went through the trouble of rebasing it.

TobiX:
Well, that obviously isn't useful if you then merge the old and new branches. Anyways, just ignore my bickering, I'll probably won't contribute much anymore, since I stopped using Performous almost a year ago.

nieknooijens:
@TobiX hey I'm still student for at least 6 months so I can still say I'm learning! (most of my classmates have never used git before... so me screwing up once in a while isn't that bad)

PR #191, opened 2016-06-17
Fixed the 4px side scrolling bug on the web interface

skyli91:
and made the code look a bit better (css)
also left aligned the headers of the database view table


nieknooijens:
schermafdruk van 2016-06-17 15-03-24

hmmm.... I don't like that gap on the right.

skyli91:
that was there before too but yeah it's a bug, let me see if i can fix :)

skyli91:
check now :)

nieknooijens:
yup that's perfect. thanks!
BTW: I'm a bit busy at work, got a few deadlines, so I'll not be programming for another 8 days, only accepting pull requests.
after that I'll try to get rid of some open issues.

PR #202, opened 2016-07-09
More color choices 

skyli91:
image
image
image

It also enables to play 11 players at the same time.

### DIFF
--- a/game/audio.hh
+++ b/game/audio.hh
@@ -13,7 +13,7 @@
 #include "pitch.hh"
 #include "util.hh"
 
-const unsigned AUDIO_MAX_ANALYZERS = 8;
+const unsigned AUDIO_MAX_ANALYZERS = 10;
 
 struct Output;
 
--- a/game/audio.hh
+++ b/game/audio.hh
@@ -13,7 +13,7 @@
 #include "pitch.hh"
 #include "util.hh"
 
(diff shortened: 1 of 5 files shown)


nieknooijens:
wouldn't it be better to merge it when the 11 players are all displayed correctly?
I'd rather not have half-features in the master.

skyli91:
Done! Fixed the scoring for 11 players! :)

nieknooijens:
perhaps we should refactor the code to allow for X number of players rather than a fixed number, because this is getting a bit out of hand. nex time someone needs 15 players, and then 20 and then 25! for now it's OK but for the next time serious refactoring is needed. although I can't think of a good reason to even allow more than 8 players. I regularly have groups of 25 people at my karaoke but together they sing loud enough to not need a microphone. I also cannot connect more than 8 mics to my PC since my mixer only has direct-outputs on the first 8 channels. (it's a 12 channel mixer).
so in the end, I'll accept the code for this time, but it smells a little.

Baklap4:
I agree with @nieknooijens Since there are now 11 players (Hating the fact it's a odd number).. The next time there'll be even more players needed.
It would be nice to re-write this so one could enter a number of players. Assign those players to specific colors (Preset (enum)? Dynamic? Random?)

PR #213, opened 2017-01-20
Added Language and Edition to search results

### LINKED ISSUE #209: [webserver] Add Language + Edition to search list
Add Language and Edition to the search list in the browser.

This will result in a list containing:
`Artist - Title | Language - Edition`
So as an example result:

`Frozen - Let It Go | Dutch - Disney`

This way it's easier to determine the language of the song if it's available in multiple languages.
If you want to see the language currently you have to add the song to the playlist and click on an item in the list to see its details. This should stay and be extended with #204 

image

Baklap4:
Added Language and Edition to search results.

# Acceptance
If Language is blank it'll result into:
`Artist - Title | Edition`

If Edition is blank it'll result into:
`Artist - Title | Language`

If Both Language and Edition are blank it'll result into:
`Artist - Title`

If Both Language and Edition are filled in it'll result into:
`Artist - Title | Language - Edition`

# Acceptance 2
Changed the playlist in the browser as well. Only adding the language. Not the edition. This way we can still have our Rick Roll work ;) And besides edition in the playlist just messes up the screen

Closes #209

### DIFF
--- a/data/themes/default/index.html
+++ b/data/themes/default/index.html
@@ -317,7 +317,7 @@ table#songtable {
 				} else {
 					durationtext = ("00"+minutes).slice(-2) + ':' + ("0"+sec).slice(-2);
 				}
-				$('#playListView').append('<li data-theme="c" data-icon="info"><a href="javascript:ViewInfo(' + i + ', false)" data-transition="slide"> #' + count + ' ' + songlist[i].Artist + ' - ' + songlist[i].Title + '      +' + durationtext + '</a></li>\n');
+				$('#playListView').append('<li data-theme="c" data-icon="info"><a href="javascript:ViewInfo(' + i + ', false)" data-transition="slide"> #' + count + ' ' + CreateSongString(songlist[i]) +' +' + durationtext + '</a></li>\n');
 				totalduration += songlist[i].Duration + playlistScreenTimeout;
 			}
 			$('#playListView').listview('refresh');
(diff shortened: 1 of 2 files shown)


Baklap4:
Could anyone test if this shows the search results as `Artist - Title | Language - Edition`: https://github.com/performous/performous/pull/213/commits/ad4de875d8abb5b31076a8b817b08a30fb673fd9

If so i can further implement this without worrying about this. Since i can't test on my hyper-v machine >.>"

game/songs.cc
```diff
@@ -61,7 +61,7 @@ void Songs::reload_internal() {
 	m_loading = false;
 	std::clog << "songs/notice: Done Loading. Loaded " << m_songs.size() << " Songs." << std::endl;
 	Game* gm = Game::getSingletonPtr();
-	gm->dialog(_("Done Loading! Loaded ") + boost::lexical_cast<std::string>(m_songs.size()) + " Songs.");	
+	gm->dialog(_("Done Loading!\n Loaded ") + boost::lexical_cast<std::string>(m_songs.size()) + " Songs.");
```
nieknooijens: this commit should actually go straight to master

data/themes/default/index.html
```diff
@@ -336,6 +336,25 @@
     }
 });
 
+function CreateSongString(song, showEdition) {
```
nieknooijens: moved common code to function? nice!

game/songs.cc
Baklap4: I wouldn't post code straight to master. Even though it's as small as changing a string.. First check than merge :P

data/themes/default/index.html
Baklap4: The whole `index.html` should be refactored imho. When i have time i might turn it into a nice project instead of 1 file ;) But that was not the case within this request hahah

PR #228, opened 2017-06-03
Offlinewebserver - Allow the web server to be used in a closed-network

nieknooijens:
do remember that we now have a lot of web files, so I moved them to their own "www" folder.
since cmake only overwrites/updates existing files during "make install" you should remove the old index.html from your install folder (usually /usr/local/share/games/performous)


data/themes/default/www/index.html
```diff
@@ -8,9 +8,9 @@
     <meta name="apple-mobile-web-app-status-bar-style" content="black">
     <title>Performous web frontend</title>
 
-    <link rel="stylesheet" href="http://code.jquery.com/mobile/1.3.2/jquery.mobile-1.3.2.min.css">
-    <script src="http://code.jquery.com/jquery-1.9.1.min.js"></script>
-    <script src="http://code.jquery.com/mobile/1.3.2/jquery.mobile-1.3.2.min.js"></script>
+    <link rel="stylesheet" href="jquery.mobile-1.3.2.min.css">
+    <script src="jquery-1.9.1.min.js"></script>
+    <script src="jquery.mobile-1.3.2.min.js"></script>
```
Baklap4: Maybe update to the new jquery/jquerymobile?

nieknooijens: that breaks the site and is not the purpose of this pull request. updating jquery mobile should be a seperate issue with a seperate pull-request

Baklap4: Will test it out myself since i have to prep my laptop for animecon anyway ;)

Baklap4:
Just tested it works in chrome and firefox so i think this can be merged :+1:

PR #234, opened 2017-06-09
Build with homebrew instead of MacPorts.

earlye:


Lord-Kamina:
IMO this is very much an issue of taste. Can you actually give any
arguments as to why Homebrew is "obviously better"? For starters, Macports
keeps everything (by default) separate from default user and system paths,
which is a very good thing; SIP notwithstanding.

Baklap4:
It's very much a matter of taste indeed, however won't it be nice to support both methods instead of choosing one over the other? Both methods are used quite a lot if i remember it correctly. This however creates another testing scenario we have to try before doing releases which can be troublesome

Lord-Kamina:
Yeah yeah, I was just kinda asleep when I replied, I thought you were
trying to replace one method for the other. Does homebrew actually need
something special to work for building? Considering the actual building
should be done with the auto build script that also takes care of bundling
and fixing dependencies. I've never actually tested the script with
homebrew but I don't see why it should not work.

Baklap4:
@Lord-Kamina I wouldn't know since i'm not running any macs anymore. Would be nice to give it a try when you or @earlye  have some time :)

earlye:
First, I closed this PR immediately after opening it specifically because I fat-fingered the opening of the PR and because it *is* a matter of taste.  I'd intended only to PR against my fork and didn't notice that github had defaulted to trying to push it here. I didn't want to push it to the main repo specifically because it creates a new testing face that I don't have the time or desire to support.

With that said, MacPorts and Homebrew are fairly mutually exclusive, and if you've already committed to Homebrew, it's annoying to switch over to MacPorts just for one project, which is why I needed to be able to build on homebrew. I wouldn't say that Homebrew is "obviously better," although in my experience it has been better maintained - that is, it is supported and used by a broader set of projects, which lowers the friction of using it. Right up until you find a project that only works with MacPorts. Obviously, others will have had the opposite experience.

To answer the other question, "Does homebrew actually need something special to work for building?" Yes. The osx.mk file here lays out the dependencies that one must install via homebrew in order to build. Or at least, the dependencies *I* had to install. There may be others that I already had, and therefore missed.  Further, homebrew has EOL'd one of Performous' dependencies, so I had to dig that out of the graveyard and host it myself - that's visible in the osx.mk as well. Once the dependencies are installed, cmake must be told where the SSL libraries are (also visible in osx.mk). Then you're home free and everything just works.

BTW, there's a segfault on exit that I had to fix in my fork, too. I'll try to find time to segment that into a PR that is suitable for pushing here, but it will require separating the git history so osx.mk doesn't try to weasel its way here. If you're industrious, feel free to take a look at this commit: https://github.com/earlye/performous/pull/3/commits/a83e82656664b46d48b987b31e40e99162db3d40

Most of it is me splitting things out for easier (for me) debugging; the actual fix is in webserver.cc.

Lord-Kamina:
I know macports and homebrew are usually mutually exclusive (although with
some tricks they can probably be made to work together). Something I did
back in the day with USDX comes to mind. For some reason, they used FINK.
So at some point when I moved to macports I couldn't build. Eventually I
realized if I just symlinked /sw to /opt/local, everything worked fine.

I don't really mind having the osx.mk if it won't interfere with the
current building process, just will make clear I likely won't be able to
test it, since I won't install homebrew.

Also, great re: segfault. I remember having noticed there being one on exit
but had no idea where.

earlye:
I understand, and I can't commit to testing `osx.mk` in a more rigorous manner than "if it's blocking my own development, I'll fix it," so I think not having it in github.com/performous makes the most sense for now. If anybody else complains about brew-vs-ports, feel free to point them at it, but I'd appreciate you warning them they're on their own if anything's broken in it. It's probably a moot point, because if you haven't seen it by now, I'm the complete edge case I normally find myself to be.

BTW, I've got another branch in my fork that has rudimentary support for pro drum scoring. If you hit a cymbal when the chart demands a tom, or vice-versa, it counts it as a miss. That feature is what really attracted me to looking at the code in the first place. I need to fix up my branch by letting the user select whether to use Pro Drums or not, and once I've done that, _then_ I'll have something to really push back here. Unless somebody wants to fork off my fork and finish up that effort for me. ;-)

In the end, I'd like to see Performous' drum support on par with Phaseshift, which goes beyond pro drum support by adding open/close hihat detection to enforce proper left foot technique, and I'm willing to put in the effort to make that happen although my availability is _super_ limited so progress will come in bits and dribbles unless somebody else is interested in that functionality, too.

Looking at the code, I think the guitar and drum controllers will need to be split into separate classes in order to support that effort, because the drum scoring is going to become enough different that it'll get hard to understand how _either_ instrument scores.

Tronic:
Current master works with Homebrew and all the deps are available (except optional feature webserver which needs cpp-netlib). Bundling still needs work, though, as that only works with Macports.

PR #260, opened 2017-06-28
[WIP] Toy around with resolution.

Lord-Kamina:
This is not even close to done, much-less actually mergeable, I just opened the PR to open a discussion.

See, here's the thing... this is something I realized while playing with the screen margins the other day... I found that one line

``` 	m_texture_width = std::min(0.96, text_x/800.); ```

in opengl_text.cc, and those arbitrary numbers seemed a bit odd to me. From what I can gather (or at least my theory) is that, 0.96 corresponds to the whole screen. minus .02 margin on each side... and, I think the "800." is actually a left-over from ages ago at a time when performous rendered in 800x600. If you check the history, you'll see that line has been on the repo roughly since the first commit.

But, if you take a look at 6a9fe6c7312d993a373f16666a30b780a4e5a993, you will see that it used to read:
```
-	double screen_width = 1.0;
 -	double screen_height = 600./800.;
  	double texture_ar = text_x/text_y;
 -	double texture_width = std::min(1.0, text_x/800.);
```
Now, with the change I just made, text looks _smaller_ on screen on higher resolutions, BUT for example you will notice that the text issue reported in #259 is actually better, even though the text looks much smaller, it's more readable. Granted, changing this will also need adjustment to placement (and perhaps sizes) of basically everything else, but I think I might be onto something here.

### CODE REVIEW

**Comment by Lord-Kamina** [comment]
I've realized  the error with the commit I made, and that is that it would produce variable placement with regards to resolution. Still, I believe we would probably benefit if we rework the UI with higher resolutions in mind. I would suggest 1366x768, that is one of the most common screen resolutions today.

**Comment by nieknooijens** [comment]
interesting... this might be what we're looking for to fix #259

File: game/opengl_text.cc [hunk 1]
```diff
@@ -262,16 +262,18 @@ void SvgTxtTheme::draw(std::vector<TZoomText> const& _text) {
...
 	m_texture_height = m_texture_width / texture_ar;
 
 	double position_x = dimensions.x1();
 	if (m_align == CENTER) position_x -= 0.5 * m_texture_width;
 	if (m_align == RIGHT) position_x -= m_texture_width;
 
-	if ((position_x + m_texture_width) > 0.48) {
+	if ((position_x + m_texture_width) > 0.48) { 
 		m_texture_width = (0.48 - position_x);
+		m_texture_height = m_texture_width / texture_ar; // Keep aspect ratio.
```
nieknooijens: that's at least one magic number less!

game/opengl_text.cc
```diff
@@ -262,16 +262,18 @@ void SvgTxtTheme::draw(std::vector<TZoomText> const& _text) {
 	}
 
 	double texture_ar = text_x / text_y;
-	m_texture_width = std::min(0.96, text_x/800.);
+	m_texture_width = std::min(0.96, text_x/1366); // Previous value of 800 appears to be a left-over from 800x600 times.
```
nieknooijens: Comment could be a little bit more clear, better would be to make a const int referenceScreenWidth = 1366;
otherwise it's ok.

basisbit: why was this change done in the first place? only thing this does is to increase visible text size,

basisbit:
maybe changing to a freetype lib based rendering approach for text would be a good permanent solution. Alternatively, change the rendering to cache the result in a glTexture and increase the svg rendering multiplier for these cached texts might be a possible approach, too.

basisbit:
on a side note, there is a nice write up about self-made vertices + shaders based text rendering available at https://web.archive.org/web/20160725213049/http://blogs.agi.com/insight3d/index.php/2008/08/29/rendering-text-fast/

game/opengl_text.cc
Lord-Kamina: Except it doesn't; it actually makes text smaller, but in most cases more readable.

game/video_driver.hh
```diff
@@ -8,6 +8,7 @@
 
 unsigned int screenW();
 unsigned int screenH();
+const unsigned int targetWidth();
```
Baklap4: i'm getting this warning:
`warning: type qualifiers ignored on function return type [-Wignored-qualifiers] const unsigned int targetWidth();`

Baklap4:
image
I think the text should be a little bit closer to eachother (probs the margin)

image

Nice work on the aspect ratio, however when changing resolution i think it's better to fall back to some lower resolution instead of keepin the old.. :o

Lord-Kamina:
Yeah, we probably need to edit the margins in the other screens, since I only made changes to the intro-screen and the audio devices screen so far.

Lord-Kamina:
New things in this branch:

Revamped lyrics highlighting; fixed line-spacing and positoning during singing, improved spacing of syllables while singing, commented-out (ideally we should delete it) a function that was apparently not being used at all. Menus now display 5 options instead of 4.

Opinions?

Baklap4:
Will check it out tomorrow since i gotta go get some sleep :+1: Feel free to rebase already if you want to :+1:

Lord-Kamina:
Both this branch and #270 are up-to-date with master; additionally I rebased this one to clean-up the history around where I broke it due to that merging conflict and such.

PR #262, opened 2017-07-02
Removed GetText using Boost::locale now

Baklap4:
# Changes

Removes GetText dependency. Using the boost module Locale now.

PR #263, opened 2017-07-02
Feature/localize webserver

Baklap4:
Added a localized webserver. see #236 for more details.

Closes #262


Lord-Kamina:
Had previously approved but I noticed something in #264  that made me change it. That request is missing commits. For example, my Spanish translation only has the last fixes I committed to it. Because I noticed that, I'd rather you re-check the three PRs to make sure nothing gets sucked into the void.

Baklap4:
@nieknooijens this is not the same as #262 ?

nieknooijens:
it has extra code not included in #262

Baklap4:
@nieknooijens isn't that the whole point of a merge request? :joy: I rebased it and it has currently everything in master.

Also i've reverted the change from `gettext_noop` to `translate_noop()` since i have no clue how poedit or any other software which uses xgettext finds this refference: https://www.gnu.org/software/gettext/manual/html_node/xgettext-Invocation.html See section `5.1.6 Language specific options` option `--keywords`

Baklap4:
Fixed @nieknooijens

PR #266, opened 2017-07-10
add missing libs to mxe settings file

nieknooijens:
add jsoncpp openssl and cpp-netlib to the settings.mk

### DIFF
--- a/win32/mxe/settings.mk
+++ b/win32/mxe/settings.mk
@@ -8,6 +8,6 @@ MXE_TARGETS := i686-w64-mingw32.shared
 #SOURCEFORGE_MIRROR := downloads.sourceforge.net
 # The three lines below makes `make` build these "local
 # packages" instead of all packages.
-LOCAL_PKG_LIST := gettext sdl sdl2 boost portaudio ffmpeg portmidi pango gdk-pixbuf librsvg libsigc++ libxml++ glew libcdio opencv libepoxy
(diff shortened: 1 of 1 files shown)


win32/mxe/settings.mk
```diff
@@ -8,6 +8,6 @@ MXE_TARGETS := i686-w64-mingw32.shared
 #SOURCEFORGE_MIRROR := downloads.sourceforge.net
 # The three lines below makes `make` build these "local
 # packages" instead of all packages.
-LOCAL_PKG_LIST := gettext sdl sdl2 boost portaudio ffmpeg portmidi pango gdk-pixbuf librsvg libsigc++ libxml++ glew libcdio opencv libepoxy
+LOCAL_PKG_LIST := gettext sdl sdl2 boost portaudio ffmpeg portmidi pango gdk-pixbuf librsvg libsigc++ libxml++ glew libcdio opencv libepoxy jsoncpp openssl cpp-netlib
```
Baklap4: openssl is not needed as package since [it's included within cpp-netlib](https://github.com/mxe/mxe/blob/master/src/cpp-netlib.mk#L12) if i'm correct. Also libcdio can be removed since we don't use it anymore. Same goes for sdl (removed) since this one is refferenced within ffmpeg and we don't use sdl anymore.

Baklap4: @nieknooijens you also do need to add the epoxy_shared flag to cmakelist.txt else you won't be able to compile

Baklap4:
Closed in favor of #280 : https://github.com/performous/performous/blob/870b403b2f69d4bf1dd83e383036e568373f2b08/win32/mxe/settings.mk

PR #270, opened 2017-07-26
[WIP] [Feature] Use a single audio backend at a time; define default preferred backends by platform.

Lord-Kamina:
Hey everybody, after some fighting with portaudio, static vars and some other stuff for a while, I think this is finally ready for merging, or at least at a point where you can check it out, test it, and request changes/say what doesn't work for you.

Please take a look at it and let me know.

Edit: Oh wait... one thing that's still missing is overhauling --audio and --audiohelp to support the new options.


game/libda/portaudio.hpp
```diff
@@ -117,6 +125,68 @@ namespace portaudio {
...
+			std::string name = convertToUTF8(info->name);
+			backends.push_back(BackendInfo(i, apiID, name, info->deviceCount));		
+		}
+	};
+		BackendInfos backends;
+		
+		std::string dump() const {
+		std::ostringstream oss;
+		oss << "audio/info: PortAudio backends:" << std::endl;
+		for (auto const& b: backends) oss << b.desc() << std::endl;
```
nieknooijens: please add curly brackets here as such
`for (auto const& b: backends) { oss << b.desc() << std::endl; } `

game/configuration.cc
```diff
@@ -352,6 +392,21 @@ void readConfigXML(fs::path const& file, int mode) {
...
 }
 
+int PaHostApiNameToHostApiTypeId (const std::string& name) {
+	if (name == "Auto") return 1337;
+	if (name == "Windows DirectSound") return 1;
+	if (name == "MME") return 2;			
+	if (name == "ASIO") return 3;
+	if (name == "Core Audio" || name == "CoreAudio") return 5;
+	if (name == "OSS") return 7; // Not an error, stupid PortAudio.
+	if (name == "ALSA") return 8;
```
nieknooijens: when I set the config to "auto" pulseaudio is detected as audio device, but it's not shown in the config option, is this correct?

Lord-Kamina: I'm not sure I'm following; can you be more precise? "Auto" is meant to fullfill two roles:

1) Fallback in case of invalid settings (for example, config file exported from one machine or platform to another), as well as a sensible default for casual end-users. i.e. the option that should work most of the time; the way it does this is simply by selecting (at runtime) an audio backend that should definitely be available depending on which platform we are running. This will be CoreAudio for macOS, ALSA for Linux (this might be subject to change since I'm not really all that familiar with different linux distros) and WASAPI for Windows.

The only thing this does is instruct performous how it should access audio devices; it should not really change which audio devices are selected by default. IIRC, in absence of a sensible choice, performous looks for the first audio device with 2 or more out channels and uses that for playing sound, correct?

So... "audio/backend" should be set to "Auto" most of the time, unless the user is sure of what he's doing (either way, the UI should never allow him to select an invalid configuration, manually editing the xml notwithstanding obviously.)

So, if you mean something related to what "audio/backend" displays in that case, odds are it's correct. If you are referring to something in the Audio Devices screen, it might be a bug that I missed.

Lord-Kamina: I just tested it on my ubuntu VM and it appears to work fine.
The default device wasn't pulse on my machine but it appeared to work with all possible valid output devices (on a VM, inside a hackintosh using HDMI... that's about as fringe as fringe-cases can get)

Still, portaudio itself has no pulseaudio api, so what happens is we use ALSA and ALSA uses PulseAudio.

game/configuration.cc
```diff
@@ -102,20 +103,44 @@ namespace {
...
+	else if (m_type == "int") {
 		int val = boost::get<int>(m_value);
 		if (val >= 0 && val < int(m_enums.size())) return m_enums[val];
 		return numericFormat<int>(m_value, m_multiplier, m_step) + m_unit;
 	}
-	if (m_type == "float") return numericFormat<double>(m_value, m_multiplier, m_step) + m_unit;
-	if (m_type == "bool") return boost::get<bool>(m_value) ? _("Enabled") : _("Disabled");
-	if (m_type == "string") return boost::get<std::string>(m_value);
-	if (m_type == "string_list") {
+	else if (m_type == "float") return numericFormat<double>(m_value, m_multiplier, m_step) + m_unit;
```
Baklap4: These `if` statements are changed to `else if`. This is not necesairy since each if returns. Please change back to if statements

game/configuration.cc
```diff
@@ -142,10 +167,12 @@ namespace {
 
 void ConfigItem::addEnum(std::string name) {
 	verifyType("int");
-	m_enums.push_back(name);
-	m_min = 0;
-	m_max = int(m_enums.size() - 1);
-	m_step = 1;
+		if (find(m_enums.begin(),m_enums.end(),name) == m_enums.end()) {
```
Baklap4: I'm wondering why the big gaps. Please reindent correctly

game/configuration.cc
```diff
@@ -155,8 +182,8 @@ void ConfigItem::selectEnum(std::string const& name) {
 }
 
 
-std::string ConfigItem::getEnumName() {
-	int val = i();
+std::string const ConfigItem::getEnumName() const {
+	int const& val = i();
 	if (val >= 0 && val < int(m_enums.size())) return m_enums[val];
```
Baklap4: Can be shortened with `val >= 0 && val < int(m_enums.size()) ? m_enums[val]; : std::string();` or does c++ not use this notation often? If not please add some braces.

game/configuration.cc
```diff
@@ -155,8 +182,8 @@ void ConfigItem::selectEnum(std::string const& name) {
...
 	return "";
```
Baklap4: Returning `std::string()` is the way to go when returning an empty string

game/configuration.hh
```diff
@@ -4,8 +4,10 @@
 #include <map>
 #include <string>
 #include <vector>
+#include <list>
 
 namespace xmlpp { class Element; }  // Forward declaration for libxml++ stuff
+class Audio;
 
 /// configuration option
```
Baklap4: For comments just use 2 `//` since we do it everywhere like that..

Lord-Kamina:
This should be ready for merging, if you don't have any issues with it.

Baklap4:
@Lord-Kamina I'll check this tonight 👍

Lord-Kamina:
The comment is made to easily document an enum, yes. But it's a portaudio enum and to find the names of what actually corresponded to what, I had to go into the portaudio source; so I thought it better to have it there, readily accessible for the future. Not to mention, not everybody uses an IDE.

Re: the 2 vs 3 slashes, I'm inclined to agree, at least in this case (in other languages/IDEs 3 slashes are using for markup documentation which is amazing), but we've actually got lots of /// comments spread around the code. IMO we can run a search&replace to normalize that but it'd be best to make a separate commit or this branch will suddenly touch almost every file in the code.

Baklap4:
LGTM however there's a merge conflict in game/configuration.cc

game/platform.cc
```diff
@@ -23,6 +23,19 @@ Platform::Platform() {
 
 const std::array<const char*,6> Platform::platformNames = {{ "Windows", "Linux", "MacOS", "BSD", "Solaris", "Unix" }}; // Relevant for debug only.
 
+int Platform::defaultBackEnd() {
+		switch (Platform::currentOS()) {
+			case windows: return 13; // WASAPI
+			case macos: return 5; // CoreAudio
+			case solaris:
+			case bsd: return 7; // OSS
+			case linux:
```
nieknooijens: I see what you're trying to do here but cascading cases are always a bit confusing, possibly add some comments here to explain things more clearly.

Lord-Kamina: Or I can just explicitly add the return for each case; frankly I doubt there's any difference from an efficiency standpoint and it will look better.

PR #272, opened 2017-07-31
Fix gettext (tools) detection and processing.

TobiX:
This restores correct translation processing by using cmake's FindGettext.

This detects translation _tools_ and not the gettext library (as our FindGettext did, the library is actually called libintl).

It would probably be a good idea to move most of the translation processing into the build system instead of relying on translators configuring poedit correctly. The cmake default macros always update all language files from the master pot on every build - If nobody objects, I would add that here, too. Then, the only thing missing would be an explicit maintainer target which updates the pot from the source code. Sounds good?


nieknooijens:
I thought @Baklap4 got rid of gettext so why is this still relevant?

TobiX:
Because there are no other command line tools to work with translations. We only use the boost translation libraries.

PR #273, opened 2017-08-01
Modifiers: Use Command instead of Control on Mac.

### LINKED ISSUE #94: OSX hotkeys
I've noticed that some of the hotkeys used in the game are a bit awkward (or, in some cases do not work, period) in mac.

In general, using F-keys is shaky terrain in mac so those should almost certainly be changed. On the other hand, if possible it wouldn't be at all bad if you could incorporate more-or-less standard mac shortcuts. (i.e. use Command-key instead of control, etc.)

For all intents and purposes Command key is "Meta"...

I'd try to remove anything assigned to F-keys in mac or things get funky quickly, additionally...

Some common mac shortcuts,

Global hotkeys:
Fullscreen:    Control+Command+F
Screenshot:    Command+Shift+3 Screenshot (4 instead of 3 allows for defining a rectangular selection) although in this case it probably doesn't hurt to keep Print Screen also)
End/Stop:    Command+.
Volume:    Volume keys.

Config screens:
Save:    Command+S
Reset:    Command+R

Synth:    Control+Option(Alt)+S
Toggle Voice:    Control+Option(Alt)+;
Karaoke:    Control+Option(Alt)+K
Pitch Wave:    Control+Option(Alt)+W
Webcam:    Control+Option(Alt)+A
Delay settings:    Control+Option(Alt)+1-6
Rewind Song: Command+Left Arrow
Back 5 sec/section: Left Arrow
Forward 5 sec/section: Right Arrow
Reload Song:    Command+R

Song Selection:
Jukebox mode:    Command+J
Filters:    Control+Option(Alt)+5-8/Tab?

Obviously feel free to tear me apart, not all of these have been very well thought-out yet.

Lord-Kamina:
Fixes #94.

I’d actually forgotten I’d ever opened this until I found it by accident. XD
From here, I'd just have to maybe update strings, but to avoid headaches for everyone it might just be better to make a single "Ctrl/Cmd+S (or whatever key) to..." rather than actually change the string according to the platform.

### DIFF
--- a/game/controllers-keyboard.cc
+++ b/game/controllers-keyboard.cc
@@ -46,7 +46,11 @@ namespace input {
 			event.hw = sdlEv.key.keysym.scancode;
 			event.value = (sdlEv.type == SDL_KEYDOWN ? 1.0 : 0.0);
 			// Get the modifier keys that we actually use as modifiers
+			#ifdef __APPLE__
+			unsigned mod = sdlEv.key.keysym.mod & (KMOD_LGUI | KMOD_LALT);
+			#else
 			unsigned mod = sdlEv.key.keysym.mod & (KMOD_LCTRL | KMOD_LALT);
+			#endif
 			// Map to keyboard instruments (sets event.button if matching)
 			if (!mod) mapping(event);
 			// Map to menu navigation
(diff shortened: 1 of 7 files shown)


Baklap4:
@nieknooijens i think it's best to just clean it up already. Else it'll be in code for the rest of time.
@Lord-Kamina I'd suggest a base class which will have a different implementation per os. This makes it extendable and gets rid of the `#ifdefs` as @nieknooijens mentioned. If you need help with it, feel free to ask 👍

Lord-Kamina:
@Baklap4 please check #277 and see whether you like it.

PR #276, opened 2017-08-22
Make screen_paths compile under MXE.

Lord-Kamina:
Haven't been able to test yet, but this at least makes it compile.
Binary appears to work fine under macOS at least. (Then again, there was never a problem with it in the first place)


game/screen_paths.cc
```diff
@@ -101,11 +101,10 @@ void ScreenPaths::generateMenuFromPath(fs::path path) {
 				std::clog << "screen_paths/notice: Ignoring hidden folder: ." << p.c_str() << std::endl;
 				continue;
 			} else {
-				m_menu.add(MenuOption(p.c_str(),_("Open folder")).call([this, p](){
-					generateMenuFromPath(p);
+				m_menu.add(MenuOption(boost::lexical_cast<std::string>(p),_("Open folder")).call([this, p](){ generateMenuFromPath(p);
```
Baklap4: Instead of using `boost::lexical_cast<std::string>(p)` which is a value that isn't translated. use `_(p)` i think this would work since `_()`-macro returns a string and other calls to `m_menu.add(MenuOption())` are done with two translateable strings.

nieknooijens: ah that was the problem? but why does linux just take it without cast and windows not?

nieknooijens: @Baklap4 "p" is the current path, it doesn't need to be translated

Baklap4: My bad translating paths is indeed rediculous 😂

Baklap4: would be nice to see the error/stackstrace of the windows build. Probably due to the fact `c_str()` vs `std::string`

Oh btw instead of doing `boost::lexical_cast` one can just run `p.string()`

Lord-Kamina:
What version of gcc are you using in Linux? Mxe uses 5.4.0. That might be it.

Baklap4:
@Lord-Kamina can't check my dev machine but checking on my work machine with ubuntu 16.04 installed it says 5.4. I don't have the issue with building master on this machine..

Lord-Kamina:
I've not the foggiest then, I guess. hahaha

PR #277, opened 2017-08-30
Issue/sanitize platforms

Lord-Kamina:
Things in this PR:

1) Created a Platform struct that can provide robust platform information for minor things that can be done at runtime.

2) Refactored platform-dependent implementations for filesystem tasks into independent include files, thus making code saner and providing a framework to do the same with the platform-dependent implementations we will be merging eventually.

3) Changed path for macOS config files and such to appropriate locations (~/Library/Preferences/Performous instead of ~/.config/performous). Also, modified pathInit so that it searches the old location and copies the settings over in case it finds them in there.
It would be great if somebody take a look at the windows implementation and checks double-checks whether we are actually storing things in the correct places.

4) Moved the dGPU enforcing code out of main.cc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ osx-utils/Performous.dmg
 osx-utils/CMakeCache.txt
 CMakeCache.txt
 CMakeFiles
+*.xcodeproj
--- a/boost_predef/os/haiku.h
+++ b/boost_predef/os/haiku.h
@@ -1,0 +1,46 @@
+/*
+Copyright Jessica Hamilton 2014
+Copyright Rene Rivera 2014-2015
+Distributed under the Boost Software License, Version 1.0.
+(See accompanying file LICENSE_1_0.txt or copy at
+http://www.boost.org/LICENSE_1_0.txt)
+*/
+
+#ifndef BOOST_PREDEF_OS_HAIKU_H
+#define BOOST_PREDEF_OS_HAIKU_H
+
+#include "../version_number.h"
+#include "../make.h"
+
+/*`
+[heading `BOOST_OS_HAIKU`]
+
+[@http://en.wikipedia.org/wiki/Haiku_(operating_system) Haiku] operating system.
+
+[table
+    [[__predef_symbol__] [__predef_version__]]
+
+    [[`__HAIKU__`] [__predef_detection__]]
+    ]
+ */
+
+#define BOOST_OS_HAIKU BOOST_VERSION_NUMBER_NOT_AVAILABLE
+
+#if !defined(BOOST_PREDEF_DETAIL_OS_DETECTED) && ( \
+    defined(__HAIKU__) \
+    )
+#   undef BOOST_OS_HAIKU
+#   define BOOST_OS_HAIKU BOOST_VERSION_NUMBER_AVAILABLE
+#endif
+
+#if BOOST_OS_HAIKU
+#   define BOOST_OS_HAIKU_AVAILABLE
+#   include "../detail/os_detected.h"
+#endif
+
+#define BOOST_OS_HAIKU_NAME "Haiku"
+
+#endif
+
+#include "../detail/test.h"
+BOOST_PREDEF_DECLARE_TEST(BOOST_OS_HAIKU,BOOST_OS_HAIKU_NAME)
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -31,7 +31,7 @@ namespace {
 
 	struct PathCache {
 		Paths paths;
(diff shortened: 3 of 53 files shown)


nieknooijens:
screenshot from 2017-08-30 09-02-50

maybe we're missing some dependency in cmakelist / or in the build script @ semaphore

Lord-Kamina:
How can I check that?

Lord-Kamina:
https://github.com/boostorg/boost/wiki/Getting-Started Leaving this here; since according to people on #boost (IRC) the issue is most likely because semaphore hasn't run

    ./bootstrap.sh
    ./b2 headers

.gitignore
```diff
@@ -11,3 +11,4 @@ osx-utils/Performous.dmg
 osx-utils/CMakeCache.txt
 CMakeCache.txt
 CMakeFiles
+*.xcodeproj
```
nieknooijens: does this need to be in the dmg package?

game/fs.cc
```diff
@@ -1,23 +1,18 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/range.hpp>
 #include <boost/thread.hpp>
+#include <cstdlib>
+
 #include <set>
 #include <sstream>
 #include <algorithm>
 
-#ifdef _WIN32
```
nieknooijens: kudo's for trying to remove platform-specific stuff

boost_predef/os/haiku.h
```diff
@@ -0,0 +1,46 @@
+/*
```
nieknooijens: apparently you dumped in the whole boost_predef directory in the repository. this means we also have to manage updates somehow which could be a problem. I suggest using a submodule or subtree for this since this can become quite a lot of clutter.

game/fs.cc
```diff
@@ -46,139 +44,55 @@ namespace {
...
-					used.insert(dir);
-				}
-			}
-			if (!bootstrapping) std::clog << logmsg << std::flush;
-		}
+		
+	#if (BOOST_OS_WINDOWS)
+	#include "platform/fs_paths.win.inc"
+	#else
+	#include "platform/fs_paths.unix.inc"
```
nieknooijens: in game/exename.cc you include a lot of platforms, but here you only specify windows and unix? this looks inconsistent to me.

game/platform.cc
```diff
@@ -0,0 +1,27 @@
...
+#include "fs.hh"
+
+Platform::platforms Platform::currentOS() {
+if (BOOST_OS_WINDOWS != 0) { return windows; }
+else if (BOOST_OS_LINUX != 0) { return linux; }
+else if (BOOST_OS_MACOS != 0) { return macos; }
+else if (BOOST_OS_BSD != 0) { return bsd; }
+else if (BOOST_OS_SOLARIS != 0) { return solaris; }
+else if (BOOST_OS_UNIX != 0) { return unix; }
+}
```
nieknooijens: boost also mentions amigaOS, IRIX or IOS. would it be worth it to also check for those? or should we add them in when needed?

.gitignore
Lord-Kamina: No; I don't usually even use xcode for coding, but I wanted to import the performous repo into it because it's re-indent feature is frankly really, really good.

game/fs.cc
Lord-Kamina: It's because execname.cc actually had that many different platform implementations already in the code.

For my version, I did only windows and unix because that's the only two meaningful differences. The further differences between linux and osx, for example, are actually determined at runtime instead of compile-time.

boost_predef/os/haiku.h
Lord-Kamina: No; I only added os.h and its explicit dependencies to the tree.
As for updates, perhaps but in reality I doubt it will be tremendously important (and in fact the version I added isn't even the latest boost because I didn't have it on hand). I don't think we need to worry too much about this because,

1) It's just a bunch of headers with mostly macro definitions.
2) It's only to be used on platforms that rely on obsolete versions of boost (<= 1.54), think of it as a polyfill.

game/platform.cc
Lord-Kamina: Maybe. I based those on the platforms that were already implemented in the execname.cc. If you think there might be important differences for one or another (and on which performous would actually run), certainly we could add them.

Lord-Kamina:
Closed via #279

Baklap4:
May this branch be deleted as its closed by #279?

Lord-Kamina:
Yes.

PR #278, opened 2017-08-30
Get pi from boost instead of defining ourselves.

Lord-Kamina:
We are using boost anyway, so I see no point in defining our own PI with an arbitrary level of precision (10 digits?)

### DIFF
--- a/game/libda/fft.hpp
+++ b/game/libda/fft.hpp
@@ -9,11 +9,9 @@
 #include <cstddef>
 #include <vector>
 
-#ifndef M_PI
-#define M_PI 3.141592653589793
-#endif
+extern const double m_pi;
 
-#define M_TAU (2.0 * M_PI)
+const double m_tau = (2.0 * m_pi);
 
 namespace da {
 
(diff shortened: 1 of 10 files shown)


game/libda/fft.hpp
```diff
@@ -9,11 +9,9 @@
 #include <cstddef>
 #include <vector>
 
-#ifndef M_PI
-#define M_PI 3.141592653589793
-#endif
+extern const double m_pi;
 
-#define M_TAU (2.0 * M_PI)
```
nieknooijens: yeah, macros are ugly and give really weird error codes since the preprocessor fills them in before any error is thrown, so it's best to get rid of them!

game/main.cc
```diff
@@ -26,6 +27,7 @@
 
 #include <boost/bind.hpp>
 #include <boost/format.hpp>
+#include <boost/math/constants/constants.hpp>
```
nieknooijens: doesn't including pi.hh already include boost/math/constants?

nieknooijens:
build also passed on semaphore

Baklap4:
LGTM

game/main.cc
Baklap4: It does, therefore should be fixed.

Lord-Kamina: I was going to say it's good practice for headers to not rely on indirect includes but then I realized main is not a header ¯\_(ツ)_/¯

So, fixed.

PR #285, opened 2017-09-04
Fix IP detection on macOS.

Lord-Kamina:
1) Changed google.com to google’s name server.
2) Added a port number, which effectively fixes this approach for macOS.

Can you guys please test whether it still works fine on the remaining platforms? (I don't see why it wouldn't)

### DIFF
--- a/game/screen_intro.cc
+++ b/game/screen_intro.cc
@@ -214,7 +214,7 @@ std::string ScreenIntro::getIPaddr() {
 	try {
 		boost::asio::io_service netService;
 		boost::asio::ip::udp::resolver resolver(netService);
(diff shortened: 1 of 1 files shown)


Lord-Kamina:
I don't think it's terribly more likely that the nameserver will break than that google.com will break. Either way, the best idea would be to either not hardcode an address, or have a list of addresses in case one fails. Also, most explanations I've seen of the hack you implemented use the nameserver. IMO it's not really relevant whether it's google.com or 8.8.8.8; if you want added reliability we should have more than one.

Lord-Kamina:
Something else I forgot to add; this relies on the server being available _most_of_the_time but not always or even especially at the time; as far as I know, this relies on the route existing but not actually being able to reach the host. You might notice for example, that even if your computer has no internet access, this approach will usually work.

PR #289, opened 2017-09-11
[WIP] Bug fix/misc windows bugs

### LINKED ISSUE #200: Missing characters from the font?
image

### LINKED ISSUE #256: config.xml integer values with thousands separator cause crash on load
### Actual behaviour

Start the game, see a fatal error messagebox appear stating `"[...]\config.xml:11 Error while reading entry: bad lexical cast: source type value could not be interpreted as target More details might be available in [path to performous\caching\infolog.txt]"`.

image

Game can be played again after modifying the line `<entry name="graphic/window_width" type="int" value="1.235"/>` and removing the thousands separator.

### Expected behaviour

Start the game and be able to play even after saving the settings in-game and later restarting it.
Settings should be saved without thousands separator. When reading values and an int is expected, any . or , should be removed from the string to support updating from old performous installs or the user doing manual changes.

### Steps to reproduce

1. Fresh install of current Performous
2. Start the game, modify + save settings in game
3. Close the game
4. Start the game, see a fatal error messagebox appear stating `"[...]\config.xml:11 Error while reading entry: bad lexical cast: source type value could not be interpreted as target More details might be available in [path to performous\caching\infolog.txt]"`.
5. After clicking OK, the game closes and you can not play.

### Details

Provide some additional information:

- performous version: current master, roughly 590fe62
- operating system + version: Windows 10 x64 with all current updates as of today
- typical German localisation settings: numbers format like 1.234,56 or 1234,56
- confirmed by @Baklap4 on linux

Lord-Kamina:
This is not done yet, as I need to fix the code to make sure Fontconfig is actually loading the config files in windows; and I haven't been able to test the latest revisions yet.

Closes #200
Closes #256

### DIFF
--- a/game/configuration.cc
+++ b/game/configuration.cc
@@ -259,9 +259,9 @@ void writeConfig(bool system) {
 		entryNode->set_attribute("name", name);
 		std::string type = item.get_type();
 		entryNode->set_attribute("type", type);
-		if (type == "int") entryNode->set_attribute("value",boost::lexical_cast<std::string>(item.i()));
+		if (type == "int") entryNode->set_attribute("value",std::to_string(item.i()));
 		else if (type == "bool") entryNode->set_attribute("value", item.b() ? "true" : "false");
-		else if (type == "float") entryNode->set_attribute("value",boost::lexical_cast<std::string>(item.f()));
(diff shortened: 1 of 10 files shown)


Lord-Kamina:
This would close #200

Issue #259, opened 2017-06-28
Performous text rendering barely readable


Baklap4:
As mentioned [here](https://github.com/performous/performous/issues/244#issuecomment-311563627) and [here](https://github.com/performous/performous/issues/244#issuecomment-311574328) and [here](https://github.com/performous/performous/issues/244#issuecomment-311574748)

With the default textsetting set to 2.0 the drawn text for example within screen_sing is barely readable on various resolutions:
![image](https://user-images.githubusercontent.com/3918517/27633484-b8f946c2-5bff-11e7-9522-a5af49c86e2c.png)
![image](https://user-images.githubusercontent.com/3918517/27633486-bd798716-5bff-11e7-98f9-feb1fa073ce3.png)
![image](https://user-images.githubusercontent.com/3918517/27633491-c6c2dd22-5bff-11e7-9a2c-edef127b83a0.png)

@basisbit @Stefan1200-de Please post the resolution your pc is using so we can further investigate this issue.

# Related issues
#10 
#260

basisbit:
also 1920x1080 for me. 
DPI zooming on windows seems to not affect this.
The above posted screenshot containing "Hey t... My name is R.." was done in sing-screen while the performous window was roughly at a size of 1800x1000 and all in-game settings on default.

Lord-Kamina:
@nieknooijens I think I'm at 2.0, but honestly I almost always play in fullHD and had never noticed that.

Stefan1200-de:
Maybe one solution could be to use a higher text quality only for the small text in the bottom left, maybe use a 50% higher text quality from the game settings. So if you set 2.0 in the game settings, use 3.0 for small text strings.

nieknooijens:
I think it's a combination of factors. system fonts get converted to SVG's which are then scaled and rendered. the text quality setting determines how much of the SVG should be calculated. it also has black outlines which become dominant at lower resolutions/svg quality and therefore some parts of letters become black. when this happens depends on the font, but eventually it will happen. so we could either calculate more points for small text (which in turn puts a heavier load on CPU) or shut off the black outline somehow.

Lord-Kamina:
There's something else I noticed the other day while playing with opengl_text.cc I noticed the other day. I want to play a bit with it later and will report with results.

Lord-Kamina:
So... call me crazy, but I believe we might actually be always rendering the text in 800x600...

Baklap4:
[...]

PR #297, opened 2017-10-31
made songinfo text larger

nieknooijens:
just a quick fix for @Baklap4 's problem in #259 .
since the text becomes even smaller in ubuntu 17.10 due to different default fonts I made the songinfo text slightly larger in the .svg.


Baklap4:
Looks great, thanks!

PR #302, opened 2017-11-12
Fix crash when showing “done loading” popup.

### LINKED ISSUE #271: Songs finished loading dialog causes a SIGSEGV crash.
### Actual behaviour

Performous crashes.

### Expected behaviour

Should show a message notifying all songs have been loaded.

### Steps to reproduce

Load Performous, wait.

### Details

Just something I noticed peripherally while working on my audio fixes (was following compiler warnings to see whether I could get rid of any) and I noticed an "unused variable" warning leading to songs.cc where a pointer to the Game singleton was being defined but not used. Immediately below was the line that would use it (commented out) saying it crashed ubuntu 16.10 but not 16.04.

I uncommented it to test and noticed it crashes on mac also. Following it around a bit, it appears the bug is here (surface.hh):

	OpenGLTexture(): m_id() { glGenTextures(1, &m_id); }

And googling for similar errors, I found this which seems to be an identical error (and we can probably derive the solution from it): https://stackoverflow.com/questions/8982472/sigsegv-error-while-using-c-on-linux-with-opengl-and-sdl

Lord-Kamina:
In short, the crash was due to the GL Context being on a different thread.

Fixes #271

### DIFF
--- a/game/main.cc
+++ b/game/main.cc
@@ -189,6 +189,10 @@ void mainLoop(std::string const& songlist) {
 		while (!gm.isFinished()) {
 			Profiler prof("mainloop");
 			bool benchmarking = config["graphic/fps"].b();
+			if (songs.doneLoading == true && songs.shownFinishDialog == false) {
+				gm.dialog(_("Done Loading!\n Loaded ") + std::to_string(songs.loadedSongs()) + " Songs.");
+				songs.shownFinishDialog = true;
+			}
 			if( g_take_screenshot ) {
 				try {
 					window.screenshot();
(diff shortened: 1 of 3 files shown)


game/surface.hh
```diff
@@ -124,7 +125,18 @@ template <GLenum Type> class OpenGLTexture: boost::noncopyable {
 		}
 		throw std::logic_error("Unknown texture type");
 	}
-	OpenGLTexture(): m_id() { glGenTextures(1, &m_id); }
+	OpenGLTexture(): m_id(0) { 
+		if (SDL_GL_GetCurrentContext() == NULL) { 
+			if (Window::stealGLContext() != 0) throw std::runtime_error(SDL_GetError()); // We need to bring the GL Context to the current thread in order to prevent a crash.
```
nieknooijens: window::stealGLContext? LOL...

Baklap4:
Lgtm and works on ubuntu 17.10
image

PR #311, opened 2017-12-10
Feature/fix warnings

nieknooijens:
fixed compile warnings, mostly case fall through ones (which require a comment to let the compiler know it was on purpose) and some ffmpeg stuff (they actually made the API easier to understand, but forgot to document it as always). there are still some std-auto-ptr deprecated errors, but these come from boost which is not our buisness. boost should fix those.

### DIFF
--- a/game/controllers-keyboard.cc
+++ b/game/controllers-keyboard.cc
@@ -1,4 +1,4 @@
-#include "controllers.hh"
+﻿#include "controllers.hh"
 #include "platform.hh"
 
 #include <set>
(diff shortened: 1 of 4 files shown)


nieknooijens:
build failed because semaphore uses ubuntu 14.04 which doesn't have the new ffmpeg API yet

nieknooijens:
@Baklap4 could be possible through PPA, but otherwise I could add in some macro-clutter that determines the ffmpeg version and changes the algorithm accordingly (as we have done in the past way too many times!)

Baklap4:
@nieknooijens i've updated the warnings i've seen just by recompiling. That's all :)

As for the ci, i'd add ifdefs if there's no other solution so we can still support a base 14.04 ubuntu although it's pretty old. Maybe extract those parts to different classes and load the correct class on runtime

nieknooijens:
@Baklap4 I'm not very fond of building in a full strategy-pattern, but i'm not a fan of macro clutter either...

Baklap4:
Strategy pattern's aint that bad imho better than ifdefs :joy: but yeah requires some more but once setup you can just hotplug another version basicly

Lord-Kamina:
I'd do it the same way as I did the platforms. I.e. Common headers, separate implementation.

nieknooijens:
@Baklap4 strategy pattern using preprocessor-macro's is a pain... (you need to use macro's since you don't want deprecated code to be compiled in to avoid errors)

nieknooijens:
#todo: move implementations to their own headers and preprocessor-macro the hell out of it...

nieknooijens:
replaced by #321

PR #312, opened 2017-12-11
[Webserver] Drag and drop issue #301

### LINKED ISSUE #301: Webserver Drag & Drop issues
While on mobile devices if one scrolls in the center items can be moved due to the drag and drop functionality. -> solution disable drag and drop on mobile

Also when drag and dropping it doesn't update the timestamps since the whole element is dragged, this fixes itself when the list is refreshed. -> Solution call refresh upon drop event

Also linked by #293

Baklap4:
Fixes and closes #301 

# Changes
* Added new div `playlist-songs-sortable` which hides on screen with bootstrap width < md and lg
* Added classes to `playlist-songs` to show when screen is < md and lg
* Clearing both divs now upon reload and refilling them both.
* While on md and lg scroll is set to true 
* One can scroll in the region of 150 pixels on both sides.
    * NOT IN THE MIDDLE (Therefore scroll is 'disabled' on mobile devices)
* Checked if we are refreshing the playlist after drag/drop
    * We already did.

### DIFF
--- a/data/themes/default/www/js/site.js
+++ b/data/themes/default/www/js/site.js
@@ -1,10 +1,12 @@
 "use strict";
 
 $('#refresh-playlist-toggle').bootstrapToggle('off');
-var list = document.getElementById("playlist-songs");
+var list = document.getElementById("playlist-songs-sortable");
 Sortable.create(list, {
     group: "words",
     animation: 150,
+    scroll: true,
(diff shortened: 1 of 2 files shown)


nieknooijens:
looks good so far, but I need to check it tonight.

PR #313, opened 2017-12-12
This lexical_cast when writing config.xml was tripping me up.

earlye:
I basically followed what was done in commit 6f4f5e7dfa3dcb63b89a82b53d791af282de8626.

Line 293 is the only consequential change; all of the other changes are because my editor removes trailing white space on save. I could have pared it down to just the lexical_cast -> to_string change, but trailing white space is kinda icky anyway.

### DIFF
--- a/game/configuration.cc
+++ b/game/configuration.cc
@@ -91,12 +91,12 @@ namespace {
         fmter % boost::io::group(std::setprecision(precision), double(m) * boost::get<T>(value));
         return fmter.str();
     }
-    
+
     std::string getText(xmlpp::Element const& elem) {
         xmlpp::TextNode const* n = elem.get_child_text();  // Returns NULL if there is no text
         return n ? std::string(n->get_content()) : std::string();
     }
-    
+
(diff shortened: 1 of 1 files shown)


earlye:
I did a quick grep, and it looks like the remaining lexical_cast's in that module only go the other way (string -> integer/float/etc).

Baklap4:
@nieknooijens aside from config there are still a few places where `boost::lexical_cast<std::string>` when `std::to_string()` can be used.  This can be fixed in another pr since this fixes something in configs.

Lord-Kamina:
Yeah, I hadn't bothered changing it because it was not causing an issue (and was unlikely to ever cause it) considering that item always will have predefined values.

earlye:
Yeah, for me it failed because it was storing the default value, and then blowing up trying to retrieve it. I'm not sure why it was storing the default value for me, or if it was supposed to, (I'm on MacOS) but if anybody ever had a valid reason to store a non-default value, it would've blown up for them.

PR #314, opened 2017-12-12
Drums: Pro Mode

earlye:
This is the work I've done on adding a pro mode for drums. It is enabled via config, and enforces cymbal/tom discrimination. Tested with keyboard (8,9,0 vs i,o,p).

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -98,6 +98,10 @@ to save the current settings to XML.
 		<short>Keyboard as drumkit</short>
 		<long>Enable keyboard as drumkit.</long>
 	</entry>
+	<entry name="game/drum_promode" type="bool" value="true">
+		<short>Drums use Pro mode</short>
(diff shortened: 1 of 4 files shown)


earlye:
I tried testing this w/ the Pro Midi USB drum controller from MadCatz yesterday. It looks like the way the controller subsystem is architected isn't very friendly to pro drums. In particular, it presumes that each event is independent, which isn't the case with pro drums. When you hit a tom, you get two events - the first indicates the color, and the second indicates that it _is_ a tom. When you hit a cymbal, you also get two events - (color, cymbal). I've got a hacky way to deal with this in my fork, and will be cleaning it up in the coming weeks. 

I don't think it's a show stopper for this PR, but be aware that in order to take advantage of pro mode drums, you probably need to be playing on a keyboard (or something that translates midi to keyboard), or using a direct midi->usb interface rather than the MadCatz controller. (I'm going to see if I can get a hold of one in order to test it out.)

PR #317, opened 2018-01-19
Feature/fix webserver searching and autocomplete

### LINKED ISSUE #293: webserver annoyances
while the new webserver is pretty fast and snappy I did encounter some minor annoyances using it last weekend @tomofair nijmegen.

first of all the Enter button behaviour is not what it's supposed to be, instead of doing the same as pressing "search"it selects the first entry of autocomplete, forcing you to press "search" with your mouse.
which brings me to the second issue: clicking an entry in the autocomplete list quite often fails. sometimes it works, but most often it fails with a message "no songs found using current filter".

it's not game-breaking, but it can become quite annoying when having to add a bunch of songs at once.

Baklap4:
Fixes and closes #293 

# Changes
* Fixed autocomplete functionality.
    * Due to malformed JSON and mixed JSON/url encoding this sometimes stopped working
* Fixed the dialog to contain info about the song on screens larger than bootstrap sm.


Baklap4:
@nieknooijens Docs are on its way, as is an even cleaner version of this branch, however i'm quite busy atm, so this will have to wait a bit more

nieknooijens:
@Baklap4 you're not the only busy person here 😫

Baklap4:
@nieknooijens I've documented the complete JS and HTML.
Shouldn't be that hard to read anymore ;)

Also i've managed to get the autocomplete working as we want it. Also the bug with the playlist modal dialog is solved.

Besides i've completely refactored the JS again to make it in even smaller parts available.

Feel free to try it out when you have time (maybe this weekend at dutch comic con ;)? )

nieknooijens:
@Baklap4 dutch comic con doesn't want karaoke anymore only gaming. but I will try it at tomocon the week after ;)

Baklap4:
I will run JSLint one more time, since i saw some mistakes myself. Then this will be merged into master

PR #318, opened 2018-01-19
Fix/PortInUse

### LINKED ISSUE #299: Webserver crash on start due to port already in use
I also experienced a webserver crash in a couple of occasions. Now, note... this might not be an actual issue because the internet and even the wifi in here is particularly shitty. Nonetheless, we should be able to gracefully handle this instead of producing a crash.

[performous_2017-11-12-022201_Anasas copy.log](https://github.com/performous/performous/files/1464867/performous_2017-11-12-022201_Anasas.copy.log)
[performous_2017-11-12-022211_Anasas copy.log](https://github.com/performous/performous/files/1464872/performous_2017-11-12-022211_Anasas.copy.log)

Baklap4:
Fixes and closes #299  

# Changes
* Added check if socket is available. 
* Trying 3 times with intervals of 20 seconds.
* Added fallback port to connect to if those 3 times trying failed.
* Trying the fallback port also 3 times with intervals of 20 seconds.
* After a total of 6 tries (normal and fallback port) we log that we couldn't start the webserver.

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -94,6 +94,12 @@ to save the current settings to XML.
 		<short>Webserver port</short>
 		<long>Port to use for the web server. Choose one that is not used by other applications.</long>
 	</entry>
+	<entry name="game/webserver_fallback_port" type="int" value="8001">
+		<!-- maximum of 16 bit is 65536, so 60000 should be a nice upper-limit -->
+		<limits min="0" max="60000" step="1" />
+		<short>Webserver fallback port</short>
+		<long>Port to use in case original webserver port is in use. Choose one that you know for sure is not used by other applications.</long>
+	</entry>
 	<entry name="game/keyboard_guitar" type="bool" value="true">
 		<short>Keyboard as guitar</short>
 		<long>Enable keyboard as guitar (Frets on Fire mode).</long>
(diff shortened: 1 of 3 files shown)


Baklap4:
It currently does not, it happily shows the server is running since we didn't have any properties exposing the webserver status.

Baklap4:
This just fixes performous not to crash whenever the choosen port is used.

https://github.com/performous/performous/issues/320 is here to resolve the text issue since some more changes have to be made to expose the status of the webserver (running or not)

nieknooijens:
maybe for future release: fallback port? if the first one doesn't start?

Baklap4:
@nieknooijens I've implemented a fallback port according to wishes.
The current behavior is as following:

1. Try to connect to default port `webserver_port`.
2. If it's crashing due to something try again in 20 seconds.
3. Keep doing this up till 3 times.

1. If after 3 times it still fails we're going to use the fallback port and try to connect to that one.
2. If it's crashing due to port in use try again in 20 seconds, with this fallback port.
3. Keep doing this up till 3 times.

If it failed after 3 times using the fallback port it kills the webserver with a log message saying that we couldn't start the server.

I think this is good to go

nieknooijens:
perfect! merged!

PR #322, opened 2018-02-24
fix: warning on non-explicit switch/call fallthrough

yoda-jm:


yoda-jm:
Don't think that clang understand this, yet I guess that it discard it nicely. However I'm not sure what is the minimal gcc version for supporting this feature (or at least discard it). I commented it to say it was c++11 compliant but I cannot find anything right now.

yoda-jm:
It seems that it only generates a warning : http://rextester.com/BKPM30072

Lord-Kamina:
I'll try it on mac when I've got the time.

PR #323, opened 2018-02-24
Use libxml++ 3.0 when available.

yoda-jm:
libxml++ 3.0 compiles with no warnings (2.6 contains a lot of auto_ptr).
First commits are refactors.

### DIFF
--- a/tools/ss_helpers.hh
+++ b/tools/ss_helpers.hh
@@ -16,6 +16,15 @@ std::string filename(boost::filesystem::path const& p) { return *--p.end(); }
 std::string filename(boost::filesystem::path const& p) { return p.filename().string(); }
 #endif
 
+namespace xmlpp {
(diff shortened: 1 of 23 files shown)


nieknooijens:
ok nice, I will try a build tonight to verify that it works (but since it passed on semaphore, I don't expect any problems)

tools/ss_helpers.hh
```diff
@@ -16,6 +17,22 @@ std::string filename(boost::filesystem::path const& p) { return *--p.end(); }
 std::string filename(boost::filesystem::path const& p) { return p.filename().string(); }
 #endif
 
+namespace xmlpp {
+#if LIBXMLPP_VERSION_2_6
```
nieknooijens: I like how all the macro clutter is quite isolated here.

PR #326, opened 2018-03-02
Fix deprecated auto keywords

Baklap4:
Fixed a few deprecated auto_ptr calls.
There's still one used in `Audio.cc` i somehow can't get rid of it.. keep getting errors would be kind if someone could help with the nasty bugger..

### DIFF
--- a/game/audio.cc
+++ b/game/audio.cc
@@ -12,6 +12,8 @@
 #include "libda/portaudio.hpp"
 #include "util.hh"
 
+#include <unordered_map>
+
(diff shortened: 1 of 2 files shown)


nieknooijens:
code is nice and clear, will do a compile run this week before approval.

Lord-Kamina:
@nieknooijens abusing auto is definitely not ideal and I DO try to avoid it, but as you say, _proper_ usage of auto is encouraged as good practice.

PR #329, opened 2018-03-09
Stdconversion

Tronic:
Replace boost::lexical_cast by C++1x equivalents std::to_string and std::stod etc. Doing so avoids dependency on that Boost library, removes compiler warnings with Boost prior to 1.61 and fixes at least one rather sloppy case of lexical_cast being used where no cast at all was needed.

This is a big change done mostly by sed and thus requires thorough testing before merging.

### DIFF
--- a/game/players.cc
+++ b/game/players.cc
@@ -6,7 +6,6 @@
 
 #include <algorithm>
 #include <boost/regex.hpp>
-#include <boost/lexical_cast.hpp>
 #include <unicode/stsearch.h>
 
 UErrorCode Players::m_icuError = U_ZERO_ERROR;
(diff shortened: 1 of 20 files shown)


Baklap4:
It seems the build on semaphore is failing:
```
CMake Error at CMakeLists.txt:39 (message):
  Your C++ compiler is too old, please upgrade.
```

game/players.cc
```diff
@@ -30,7 +29,7 @@ void Players::load(xmlpp::NodeSet const& n) {
 		xmlpp::Attribute* a_id = element.get_attribute("id");
 		if (!a_id) throw PlayersException("Attribute id not found");
 		int id = -1;
-		try {id = boost::lexical_cast<int>(a_id->get_value());} catch (boost::bad_lexical_cast const&) { }
+		try { id = std::stoi(a_id->get_value()); } catch (std::exception&) { }
```
nieknooijens: so you're basically catching every exception here?

game/screen_sing.cc
```diff
@@ -595,7 +594,7 @@ void ScreenSing::drawMenu() {
 	th.bg.draw();
 	// Loop through menu items
 	w = 0;
-	int player = 0;
+	std::size_t player = 0;
```
nieknooijens: this is actually a nice improvement, this way players cant go negative since a size_t is an unsigned 32 bit int.

game/players.cc
Baklap4: Better throw a http://www.cplusplus.com/reference/stdexcept/invalid_argument/ exception i think

PR #332, opened 2018-03-13
Issue/fix icon

Lord-Kamina:
We were calling both `SDL_SetWindowTitle` and `SDL_SetWindowIcon` before actually creating the window, which was raising an exception.
Besides that, I want to see whether `SDL_SetWindowIcon` is at all necessary nowadays. It only seems like it might be necessary in Windows and that implementation has a pretty nasty issue, it loads the image data directly on the stack, instead of using malloc.

In a nutshell, it'd be great if somebody could test how this behaves in windows and linux.


game/video_driver.cc
```diff
@@ -77,6 +70,12 @@ Window::Window(unsigned int width, unsigned int height, bool fs): m_windowW(widt
 		SDL_GL_CreateContext(screen);
 
 	}
+// 	{
+// 		SDL_Surface* icon = SDL_LoadBMP(findFile("icon.bmp").string().c_str());
+// 		SDL_SetWindowIcon(screen,icon);
```
nieknooijens: why comments? why don't we wrap it in #ifdefs?

Baklap4: Even better if there's no need for it anymore: delete it

yoda-jm:
Icon in title bar works on Linux, so LGTM.
Please remove code instead of commenting.

Baklap4:
Looks good to me. I removed the commented code.

Baklap4:
Currently building for windows.. Can take some time

Lord-Kamina:
@nieknooijens I just remade the glow in Illustrator because  Illustrator didn't like/understand the way it had been done originally. Also, I made sure the sizes are actually right, since the glow clipped the artboard, the png and bmp weren't actually square.

@yoda-jm I left it commented-out in case it became necessary to re-add that for windows. If the icon in the titlebar does not work as expected in windows, now we have two options: 
a) Stick with the SDL call and at least for now stick with a 32x32 bmp.
b) Use a native windows call and see what windows does when provided with a larger image.

Baklap4:
Someone on windows please test this build of this branch: http://asgardsings.nl/performous-installer-g4e372338.exe

nieknooijens:
@Lord-Kamina the original svg was made in inkscape, so using that should have no problems.

Baklap4:
The icon works on windows aswell!

PR #336, opened 2018-03-16
Catch crash related to locale.

Lord-Kamina:
We’ve had this uncaught exception for a while now, we can get rid of it by specifying a default locale.


yoda-jm:
LGTM

PR #339, opened 2018-03-19
Issue/std regex

Lord-Kamina:
If we're gonna get rid of boost::regex, might as well do it everywhere.

One thing I'm not sure about and would like input on is boost::regex::empty; that function never made it to the STL. Are you happy with my solutions to the issue?


game/controllers-midi.cc
```diff
@@ -3,22 +3,22 @@
...
-			boost::regex re(config["game/midi_input"].s());
+			std::regex re(config["game/midi_input"].s());
 			for (int dev = 0; dev < Pm_CountDevices(); ++dev) {
 				try {
 					PmDeviceInfo const* info = Pm_GetDeviceInfo(dev);
 					if (!info->input) continue;  // Not an input device
 					if (info->opened) continue;  // Already opened
 					std::string name = getName(dev);
-					if (!re.empty() && !regex_search(name, re)) continue;
+					if (!config["game/midi_input"].s().empty() && !regex_search(name, re)) continue;
```
Baklap4: If `config["game/midi_input"].s()` returns an empty string, than the check `!config["game/midi_input"].s().empty()` is not necesairy since `std::regex re` may be initialized with an empty string and than matching nothing. This goes for all these 'empty' checks

Baklap4: Nice work on getting it all replaced though! :+1:

Baklap4:
Build fails.. though :(

Lord-Kamina:
It's building in Clang and failing in GCC...

Lord-Kamina:
I just changed the build commands on semaphore to use gcc-4.9 (didn't know it was available) and it appears to be doing ok so far.

game/controllers-midi.cc
```diff
@@ -3,22 +3,22 @@
...
 		Midi() {
-			boost::regex re(config["game/midi_input"].s());
+			std::regex re(config["game/midi_input"].s());
 			for (int dev = 0; dev < Pm_CountDevices(); ++dev) {
 				try {
 					PmDeviceInfo const* info = Pm_GetDeviceInfo(dev);
 					if (!info->input) continue;  // Not an input device
 					if (info->opened) continue;  // Already opened
 					std::string name = getName(dev);
-					if (!re.empty() && !regex_search(name, re)) continue;
```
nieknooijens: isn't it a little dangerous to skip the empty check here?

game/controllers.cc
```diff
@@ -109,7 +109,7 @@ struct ControllerDef {
 		if (ev.source.type != sourceType) return false;
 		if (!deviceMinMax.matches(ev.source.device)) return false;
 		if (!channelMinMax.matches(ev.source.channel)) return false;
-		if (!deviceRegex.empty() && !regex_search(devName, deviceRegex)) return false;
```
nieknooijens: same here

game/songparser-ini.cc
```diff
@@ -1,7 +1,6 @@
 #include "songparser.hh"
 
 #include <boost/algorithm/string.hpp>
-#include <boost/regex.hpp>
```
nieknooijens: whut? these files didn't need it in the first place?

game/controllers-midi.cc
Baklap4: Like i said on discord. The std:basic_regex has no empty method anymore see this paper: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2003/n1507.htm

If a std::basic_regex (std::regex is derived from it) is filled with an empty string it will not match anything. So it wouldn't matter much

game/controllers.cc
Baklap4: Like i said on discord. The std:basic_regex has no empty method anymore see this paper: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2003/n1507.htm

If a std::basic_regex (std::regex is derived from it) is filled with an empty string it will not match anything. So it wouldn't matter much

game/songparser-ini.cc
Baklap4: correct.

game/controllers-midi.cc
nieknooijens: ok I wasn't able to read the discord due to MIA last weekend.

Baklap4:
@Lord-Kamina conflicts to be resolved the joy :joy:

PR #310, opened 2017-12-08
Implement a diacritic-agnostic searching algorithm.

Lord-Kamina:
This also includes @Baklap4's "copy-if" modification to the filter.

It closes #290 since it's now a duplicate of #293

PR #342, opened 2018-03-21
Revert UTF-8 Warning removal

Baklap4:
I'd love to revert this commit by @Tronic d7240e07accbf0a3dba126870c1761c3023b495b

Reasons:
1. Dev's can check their logs to see if they are dealing with UTF8+BOM or not
2. I used it for correctly formatting my txt files
3. Currently i have no option (within performous) to check if my files are UTF-8 with or without bom.
4. I'd love to revert this since the change made by @Tronic hasn't been peer reviewed by others. Comitting directly to master should be considered a bad practice imho.

As discussed at discord / github we implemented a bool to either turn it on or off as per #310

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -39,6 +39,10 @@ to save the current settings to XML.
 	</menu>
 
 	<!-- Gameplay preferences -->
+	<entry name="game/bom_warnings" type="bool" value="false">
+		<short>Warn about UTF-8 BOM.</short>
+		<long>Dump warnings to game log when a BOM is present.</long>
+	</entry>
 	<entry name="game/karaoke_mode" type="int" value="0">
 		<limits>
 			<enum>Off</enum>
--- a/game/unicode.cc
+++ b/game/unicode.cc
@@ -38,8 +38,10 @@ void convertToUTF8(std::stringstream &_stream, std::string _filename) {
         match = std::pair<std::string,int>("UTF-8",100); // If there's no filename, assume it's internal text and thus utf-8.
     }
     icu::UnicodeString ustring;
-	// Test for UTF-8 BOM (a three-byte sequence at the beginning of a file)
     if (data.substr(0, 3) == "\xEF\xBB\xBF") {
+        if (config["game/bom_warnings"].b()) {
+            std::clog << "unicode/warning: " << _filename << " UTF-8 BOM ignored. Please avoid editors that add a BOM to UTF-8 (e.g. Notepad)." << std::endl;
+        }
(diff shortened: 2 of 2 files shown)


Lord-Kamina:
We've discussed this before; I've never really agreed to this message. Although, something you might wish to consider; with out current algorithm, the BOM is actually beneficial because it will guarantee the files will be recognized and treated as unicode.

On the other hand, If you want to cling to it; fine as long as I can turn it off (the reason for having the switch in configuration in the first place)

yoda-jm:
I think that we should not have a configuration item for a single line of debug
either it is not needed (ever) and then we remove it and developper wanting more information adds it
or it is interesting at a level and we add this log to this level (for example debug)

I think that the message could be of some interest for some people, if one wants to keep it, I'm in favor of adding it at a debug level.

Lord-Kamina:
Honestly, I wouldn’t even want it at debug level; I run performous at that level like 80% of the time, and it clogs my log.
Initially I had proposed the idea to make it a hidden option (doesn’t appear in UI, needs to be set directly in config.xml). Maybe that will make everybody happy?

 

On March 21, 2018 at 21:07:33, Vincent Le Ligeour (notifications@github.com) wrote:

I think that we should not have a configuration item for a single line of debug
either it is not needed (ever) and then we remove it and developper wanting more information adds it
or it is interesting at a level and we add this log to this level (for example debug)

I think that the message could be of some interest for some people, if one wants to keep it, I'm in favor of adding it at a debug level.

nieknooijens:
wouldn't it be better to make this option hidden and enable it using a command line option?
because it can be useful to developers.

yoda-jm:
The problem is not having them displayed (yes it is, but it is not the main problem), it is just to have a configuration entry to know if a single log line should be displayed or not.
If this line is only useful for a developer while debugging, I would go for a complete removal and let:
- developer add the line when he/she wants to have it
- keep the line in a local patch/branch rebased for personal use.

Tronic:
Initially when I added that warning a few years ago Performous was the only singing game that even used UTF-8 encoded TXT files and it was a concern that BOMs would break USDX's or others' parsers.

Apparently BOMs are now so commonly used that they may be considered safe and not worth warning for, and this is the reason why I removed that warning. Granted, doing it straight on master was a bit lazy even though the change was really minor IMO.

There exist external tools for charset conversions and BOM handling for those who might be interested (e.g. recode on UNIX).

Baklap4:
Closing this due to everyone voting against it. I've made local rebased branch to keep it for myself..

PR #344, opened 2018-03-24
fix: detection of audio/video end of files

yoda-jm:
This should fix the CPU usage of ffmpeg decoding

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -7,6 +7,9 @@
 #include <iostream>
 #include <sstream>
 #include <stdexcept>
+#include <thread>
+#include <chrono>
+#include <system_error>
 
 extern "C" {
 #include AVCODEC_INCLUDE
@@ -16,10 +19,16 @@ extern "C" {
 #include AVUTIL_INCLUDE
 #include AVUTIL_OPT_INCLUDE
 #include AVUTIL_MATH_INCLUDE
+#include AVUTIL_ERROR_INCLUDE
 }
 
 #if (LIBAVCODEC_VERSION_INT) < (AV_VERSION_INT(52,94,3))
 #	define AV_SAMPLE_FMT_S16 SAMPLE_FMT_S16
+#endif
+
+// Some versions of libav does not contain this definition.
+#ifndef AV_ERROR_MAX_STRING_SIZE
+#	define AV_ERROR_MAX_STRING_SIZE 64
 #endif
 
 #define AUDIO_CHANNELS 2
@@ -138,7 +147,12 @@ void FFmpeg::operator()() {
 	m_duration = m_formatContext->duration / double(AV_TIME_BASE);
 	audioQueue.setDuration(m_duration);
 	int errors = 0;
+	bool eof = false;
 	while (!m_quit) {
+		if(eof) {
+			std::this_thread::sleep_for(std::chrono::milliseconds(100));
+			continue;
+		}
 		try {
 			if (audioQueue.wantSeek()) m_seekTarget = 0.0;
 			if (m_seekTarget == m_seekTarget) seek_internal();
@@ -146,7 +160,8 @@ void FFmpeg::operator()() {
 			errors = 0;
 		} catch (eof_error&) {
 			videoQueue.push(new Bitmap()); // EOF marker
-			boost::thread::sleep(now() + 0.1);
+			eof = true;
+			std::clog << "ffmpeg/debug: done loading " << m_filename << std::endl;
 		} catch (std::exception& e) {
 			std::clog << "ffmpeg/error: " << m_filename << ": " << e.what() << std::endl;
 			if (++errors > 2) { std::clog << "ffmpeg/error: FFMPEG terminating due to multiple errors" << std::endl; m_quit = true; }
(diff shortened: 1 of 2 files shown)


game/ffmpeg.cc
```diff
@@ -138,14 +138,22 @@ void FFmpeg::operator()() {
 	m_duration = m_formatContext->duration / double(AV_TIME_BASE);
 	audioQueue.setDuration(m_duration);
 	int errors = 0;
+	bool eof = false;
 	while (!m_quit) {
+		if(eof) {
+			// FIXME: use a boost::condition on m_quit
+			boost::thread::sleep(now() + 0.1);
```
Baklap4: I'd rewrite this with 
`std::this_thread::sleep_for(chrono::milliseconds(100));`

also include the header files: 
```
#include <thread>
#include <chrono>
```

game/ffmpeg.cc
```diff
@@ -138,14 +138,22 @@ void FFmpeg::operator()() {
...
 		try {
 			if (audioQueue.wantSeek()) m_seekTarget = 0.0;
 			if (m_seekTarget == m_seekTarget) seek_internal();
 			decodePacket();
 			errors = 0;
 		} catch (eof_error&) {
 			videoQueue.push(new Bitmap()); // EOF marker
+			eof = true;
+			std::clog << "ffmpeg/debug: done loading " << m_filename << std::endl;
 			boost::thread::sleep(now() + 0.1);
```
Baklap4: I'd rewrite this with 
`std::this_thread::sleep_for(chrono::milliseconds(100));`

also include the header files: 
```
#include <thread>
#include <chrono>
```

game/ffmpeg.cc
```diff
@@ -183,18 +191,39 @@ void FFmpeg::seek_internal() {
...
 #if LIBAVCODEC_VERSION_INT >= (AV_VERSION_INT(57, 37, 0))
 	AVPacket pkt;
-	while (av_read_frame(m_formatContext, &pkt) >= 0) {
+	while (true) {
+		int ret = av_read_frame(m_formatContext, &pkt);
+		if(ret == AVERROR_EOF) {
+			// End of file: no more data to read.
+			throw FFmpeg::eof_error();
+		} else if(ret < 0) {
+			// FIXME: we might want to return an error here.
```
Baklap4: Either throw the error, or remove the comment. OR remove the comment, and add a log line with the error.

game/ffmpeg.cc
```diff
@@ -183,18 +191,39 @@ void FFmpeg::seek_internal() {
...
-			// error
+		ret = avcodec_send_packet(m_codecContext, &pkt);
+		if(ret == AVERROR_EOF) {
+			// End of file: no more data to read.
+			throw FFmpeg::eof_error();
+		} else if(ret == AVERROR(EAGAIN)) {
+			// not enough data for decoder, read more
+			continue;
+		} else if(ret < 0) {
+			// FIXME: we might want to return an error here.
```
Baklap4: Either throw the error, or remove the comment. OR remove the comment, and add a log line with the error.

game/ffmpeg.cc
```diff
@@ -183,18 +191,39 @@ void FFmpeg::seek_internal() {
...
 			ret = avcodec_receive_frame(m_codecContext, frame.get());
-			if (ret == AVERROR(EAGAIN) || ret == AVERROR_EOF) {
+			if(ret == AVERROR_EOF) {
+				// End of file: no more data.
+				throw FFmpeg::eof_error();
+			} else if(ret == AVERROR(EAGAIN)) {
+				// not enough data to decode a frame, go read more and feed more to the decoder
+				break;
+			} else if(ret < 0) {
+				// FIXME: we might want to return an error here.
```
Baklap4: Either throw the error, or remove the comment. OR remove the comment, and add a log line with the error.

game/ffmpeg.cc
```diff
@@ -138,14 +138,22 @@ void FFmpeg::operator()() {
 	m_duration = m_formatContext->duration / double(AV_TIME_BASE);
 	audioQueue.setDuration(m_duration);
 	int errors = 0;
+	bool eof = false;
 	while (!m_quit) {
+		if(eof) {
+			// FIXME: use a boost::condition on m_quit
```
nieknooijens: if m_quit is rewritten as "if(eof) then the comment should be updated as well.

game/ffmpeg.cc
yoda-jm: Done

game/ffmpeg.cc
yoda-jm: Removed (already in the while loop

game/ffmpeg.cc
yoda-jm: now throw

game/ffmpeg.cc
yoda-jm: now throw

game/ffmpeg.cc
yoda-jm: now throw

game/ffmpeg.cc
yoda-jm: eof is meant to stay, we want to keep being here, we just want to be notified in the eof loop when quit is called, we might prefer a boost::condition instead of waking up.

Baklap4:
Approving changes when the build is ready to go again. Codewise my things have been picked up. Good to see boost::thread to slowly go away ;)

game/ffmpeg.cc
nieknooijens: the comment here is still wrong tough.

yoda-jm: No, this comment is correct : we do not want to pull m_quit comparison (with the sleep), but instead use a condition on the quit behavior. It avoids waking up the thread each 100ms

Baklap4: @yoda-jm instead of `boost::condition` use `std::condition_variable`

As for the comment, i have a pleasant hate regarding `FIXME` and `TODO` comments since most won't be done. Either fix it with this PR, or create a new issue discussing this (and remove the comment) so this pr can be closed.

yoda-jm:
ptal, removed the challenged log and created the issue #346 to solve the problem

Baklap4:
Awesome finally is performous useable again hahah...

PR #345, opened 2018-03-27
Issue/songparser cleanup

Tronic:
Maintenance of SongParser code; shouldn't affect functionality.

- Using C++1x features scoped enums and in-class init of variables
- Separating a few TXT-specific variables
- Reorganizing class SongParser so that functions and variables aren't mixed
- Fixing Doxygen comments
- Other minor cleanup


game/songparser.hh
```diff
@@ -10,105 +10,103 @@
...
+	void vocalsTogether();
+	void guessFiles();
+	bool getline (std::string& line) { ++m_linenum; return (bool) std::getline (m_ss, line); }
+	BPM getBPM (double ts) {
+		for (auto& itb : boost::adaptors::reverse (m_bpms)) {
+			if (itb.begin <= ts) return itb;
+		}
+		throw std::runtime_error("No BPM definition prior to this note...");
+	}
+	void addBPM (double ts, double bpm) {
```
nieknooijens: please move functions out of the header.

Tronic: Moved now.

Tronic:
I won't hack more on this today. Notes:

- At least one FoF song doesn't work (crashes Performous about missing BPM data when entering screen sing), but this is also broken on master
- I wonder what the idea of that mutex in Song was and whether one is actually needed

game/songparser.cc
```diff
@@ -308,4 +275,31 @@ void SongParser::finalize () {
 		// Add song beat markers
 		for (unsigned ts = 0; ts < m_tsEnd; ts += m_tsPerBeat) { m_song.beats.push_back (tsTime (ts)); }
 	}
-}
\ No newline at end of file
+}
+
+SongParser::BPM SongParser::getBPM(double ts) const {
+	for (auto& itb: boost::adaptors::reverse(m_bpms)) {
+		if (itb.begin <= ts) return itb;
```
Baklap4: In the original code of `getBPM` one had the line `if (itb.begin >= ts && ts != 0) continue;`
This crashed indeed telling the user no bpm was found.

@Lord-Kamina looked at this and noticed that changing the if to `if (itb.begin > ts && ts != 0) continue;` solved the problem.

Baklap4: Also as in current master this method looks as this
```
BPM getBPM (double ts) {
	for (auto& itb : boost::adaptors::reverse (m_bpms)) {
			if (itb.begin > ts) { 
				continue; 
			} else {
				return itb;
			}
		}
		throw std::runtime_error("No BPM definition prior to this note...");
	}
}
```

In yours (@Tronic ) i don't see any continue in the loop hmmm?

Lord-Kamina:
Actually, the first BPM might not be 0, so it should just be a "larger than" (or lesser than, depending on how you order the variables, I guess).

game/songparser.cc
Tronic: @Baklap4 The new code has that if clause reversed so that return comes first. After returning, no else block is needed, and continue does nothing because it would be the last statement of the loop.

The code in your message is misindented and has an extra } (master appears to be OK - throw is outside of the loop).

game/songparser.cc
```diff
@@ -82,39 +69,31 @@ m_tsEnd()
...
-		} else if (type == XML) {
-			xmlParseHeader();
-		} else if (type == SM) {
-			smParseHeader(); s.dropNotes();
-		}// Hack: drop notes here
+		if (type == TXT) txtParseHeader();
+		else if (type == INI) iniParseHeader();
+		else if (type == XML) xmlParseHeader();
+		else if (type == SM) {
+			smParseHeader(); s.dropNotes();  // Hack: drop notes here (load again when playing the song)
```
Baklap4: Newline between two function calls

PR #352, opened 2018-04-03
Changed throw error to log line.

### LINKED ISSUE #351: Bug/Feature Songs getting removed while trying to play.
As mentioned on Discord i've encountered an issue with some songs i've played last event. The message telling me: `Notes must have positive durations.`

This was added in commit: https://github.com/performous/performous/commit/eb0c4ed0fe884a1df058f52a2455426c18515827#diff-a74329148b8018fe77d1730d9c5cfe4f

image

Personally yes notes should have a positive duration, however some textfiles have `0` in their text files which used to work before this commit.

I think this check should be `if (length >= 0) { then keep executing } else throw error('notes should've positve duration')`

How is this currently done in usdx @basisbit ?

Baklap4:
# Issues

Closes #351 

# Changes
Instead of throwing an error we casually let this specific check slide through, as the same happens in USDX. The notes with length of 0 are displayed on screen as 'overlapping notes'.  

The song is still playable even though the length is 0

### DIFF
--- a/game/songparser-txt.cc
+++ b/game/songparser-txt.cc
@@ -168,7 +168,7 @@ bool SongParser::txtParseNote(std::string line) {
 		{
 			unsigned int length = 0;
 			if (!(iss >> ts >> length >> n.note)) throw std::runtime_error("Invalid note line format");
-			if (length < 1) throw std::runtime_error("Notes must have positive durations.");
+			if (length < 1) std::clog << "songparser/info: Notes must have positive durations." << std::endl;
 			n.notePrev = n.note; // No slide notes in TXT yet.
 			if (m_relative) ts += m_txt.relativeShift;
 			if (iss.get() == ' ') std::getline(iss, n.syllable);


Tronic:
Did you test this change with songs that have such notes? Any problems with note rendering, lyric display, scoring, etc?

Baklap4:
@Tronic Such notes now get ignored and a message on the screen will popup (as it always was) saying the note is overlapping with other notes

I didn't encounter any other problems, i could actually play some of the faulty songs again

game/songparser-txt.cc
```diff
@@ -168,7 +168,7 @@ bool SongParser::txtParseNote(std::string line) {
 		{
 			unsigned int length = 0;
 			if (!(iss >> ts >> length >> n.note)) throw std::runtime_error("Invalid note line format");
-			if (length < 1) throw std::runtime_error("Notes must have positive durations.");
+			if (length < 1) std::clog << "songparser/info: Notes must have positive durations." << std::endl;
```
nieknooijens: please add a return here, or the code below will have undefined behaviour.

PR #353, opened 2018-04-04
Replace all volatile variables with std::atomic

Tronic:
# Changes
Replace all volatile variables with std::atomic, as one is supposed t…o do for standards-compliance when threaded access is involved.


Tronic:
Okay to merge now? This one should be pretty straight-forward despite the large number of changes.

Baklap4:
Gonna check it out tonight 👍

PR #354, opened 2018-04-04
Feature/ffmpeg build fix + boost removals

Baklap4:
# Changes
* Fixes the broken build by including `<mutex>`
* Removed boost integers and replaced them by c++11 equivalents `<cstdint>`
* Removed ptr deque and replaced them by c++11 equivalent (`<deque>`)

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -156,7 +156,7 @@ void FFmpeg::operator()() {
 			decodePacket();
 			errors = 0;
 		} catch (eof_error&) {
-			videoQueue.push(new Bitmap()); // EOF marker
+			videoQueue.push(Bitmap()); // EOF marker
 			eof = true;
 			std::clog << "ffmpeg/debug: done loading " << m_filename << std::endl;
 		} catch (std::exception& e) {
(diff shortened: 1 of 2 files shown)


Baklap4:
Ahh i'll fix that tonight! thanks! Pointers and memory management are still quite new for me hmmm

Tronic:
The code could be refactored to make use of C++ move semantics and std::move. Then no pointers would be needed. The Bitmap type is already movable, so you could actually stick with std::deque<Bitmap> but change the code so that the original Bitmap is stack-allocated (not using operator new) and passed around as a rvalue reference (Bitmap&& bitmap).

nieknooijens:
cstdint is actually stdint.h from the c library. it's not a c++ lib.

Baklap4:
@Tronic i made the std::deque use a unique_ptr of bitmap. Did you mean something like that?

Tronic:
I changed it to use move semantics.

In ffmpeg.cc there is a Bitmap f (no pointers or dynamic allocation!) that is passed to videoQueue.push(), which moves (no copying) its data into another Bitmap inside the deque. 

std::move is needed on videoQueue.push(std::move(f)) to indicate that variable f (lvalue) won't be needed after this point (and its data can be stolen), and then VideoQueue::push uses std::move when placing its parameter f (which itself is an lvalue even though its type is Bitmap&&) into the container. Similarly, VideoQueue::tryPop uses std::move on m_queue.front() to move-assign it into Bitmap& f.

On the contrary, videoQueue.push(Bitmap()) doesn't need std::move because Bitmap() returns a temporary object (rvalue) that can always be moved from.

Tronic:
Tested with 1080p background video, runs real smooth. 4k is a bit too heavy for my laptop so it fades out after the decoding buffer is exhausted (probably because we convert it into RGB, using libavformat, instead of playing YUV video like VLC and others do).

Baklap4:
If i have internet tonight i'll try it out and merge it 👍

PR #355, opened 2018-04-05
Improved warning message and handling of non-stereo3d-capable GPUs.

Tronic:
Doesn't force the config setting anymore but instead displays a flash message whenever the config option is enabled on a system that doesn't support it.


Lord-Kamina:
Do you use stereo3d? For a while now I've been having an itch to do away with it since IMO it's mostly a gimmick and I haven't met a single person who uses it.

Lord-Kamina:
Other than that, I cannot test it since I'm vacationing in the US and didn't bring either my laptop or my external HD with dev stuff, but it looks good to me.

Tronic:
I do use stereo3d (when it works)... It's especially cool in band mode :)

PR #357, opened 2018-04-05
Issue/glcore

Tronic:
Change from OpenGL 2.1 compatibility/legacy profile to OpenGL 3.3 core profile.

### DIFF
--- a/game/glutil.cc
+++ b/game/glutil.cc
@@ -2,12 +2,19 @@
 
 namespace glutil {
 
-void VertexArray::generateVBO() {
-	if (m_vbo != 0) glDeleteBuffers(1, &m_vbo);
+VertexArray::VertexArray() {
+	glGenVertexArrays(1, &m_vao);
 	glGenBuffers(1, &m_vbo);
-	glBindBuffer(GL_ARRAY_BUFFER, m_vbo);
-	glBufferData(GL_ARRAY_BUFFER, sizeof(VertexInfo) * size(), &m_vertices.front(), GL_STATIC_DRAW);
-	glBindBuffer(GL_ARRAY_BUFFER, 0);
+}
+
+VertexArray::~VertexArray() {
+	clear();
+	glDeleteVertexArrays(1, &m_vao);
+	glDeleteBuffers(1, &m_vbo);
+}
+
+void VertexArray::clear() {
+	m_vertices.clear();
 }
 
 void VertexArray::draw(GLint mode) {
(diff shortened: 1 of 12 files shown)


nieknooijens:
so basically you rewrote the shaders?

Tronic:
In shaders "in" and "out" replace "attribute" and "varying". texture() is used instead of type-specific texture2D() and texture2DRect() of older GLSL versions, allowing removal of some code (especially since Rect is no longer used at all). Additionally a workaround for a bug in Nvidia drivers (for GPUs that were old even in year 2011) was removed (hopefully Nvidia fixed their drivers by now).

Tronic:
glGetAttribLocation (finding GLSL variables by name) is safer than integer values and performance cannot really be an issue for the (relatively low) number of calls that we make.

And no, I didn't know that anyone was already working on this :O

Tronic:
Regarding performance, the biggest issue with this pull request should be that it updates the VBOs every time they are drawn. I suppose this can affect guitar/drum gameplay when a lot of notes are visible. Didn't do any testing and didn't bother to make it only update when needed.

Lord-Kamina:
I agree re: name vs integers. The point is that the whole point is moot because GLSL allows you to explicitly define attribute locations now, so you don't need getLocation because they'll always be where you said they are. For the most part I get the idea your version is cleaner than mine, but do check mine out and see if you can merge the best of both.

Lord-Kamina:
Here: https://github.com/performous/performous/tree/issue/migrate_to_opengl_33

game/glutil.cc
```diff
@@ -40,13 +52,52 @@ void VertexArray::draw(GLint mode) {
...
+		// Prefix with all currently active GLErrorChecker contexts
+		std::string logmsg = "opengl/error: ";
+		for (auto s: stack) logmsg += s + ": ";
+		logmsg += msg(err) + "\n";
+		std::clog << logmsg << std::flush;
+	}
+	stack.back() = info + " after " + what;
+}
+
+/* static */ std::string GLErrorChecker::msg(GLenum err) {
```
Baklap4: Remove the static keyword if it's not needed

game/glutil.cc
```diff
@@ -40,13 +52,52 @@ void VertexArray::draw(GLint mode) {
...
+		case GL_INVALID_OPERATION: return "Invalid operation";
+		case GL_INVALID_FRAMEBUFFER_OPERATION: return "FBO is not complete";
+		case GL_STACK_OVERFLOW: return "Stack overflow";
+		case GL_STACK_UNDERFLOW: return "Stack underflow";
+		case GL_OUT_OF_MEMORY: return "Out of memory";
+		default: return "Unknown error";
+	}
+}
+
+/* static */ thread_local std::vector<std::string> GLErrorChecker::stack;
```
Baklap4: Remove the static keyword if it's not needed

Tronic: This is a definition for a static member variable. Due to C++ syntax, the static keyword can only be used inside class {} and not again when the variable is defined. The comment is there to make it clear to the reader that this is in fact a static member.

Baklap4: alright Granted 👍

PR #358, opened 2018-04-05
Issue/ffmpegfuture

Tronic:
Proof-of-concept, mostly -- using std promise/future for quiting of ffmpeg. This solves the issue of having to poll m_quit for true before exiting, once EOF is reached.

### DIFF
--- a/game/webcam.hh
+++ b/game/webcam.hh
@@ -45,8 +45,8 @@ class Webcam {
 	CamFrame m_frame;
 	Surface m_surface;
 	bool m_frameAvailable;
-	volatile bool m_running;
-	volatile bool m_quit;
+	std::atomic<bool> m_running;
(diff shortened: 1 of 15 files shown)


game/webcam.hh
```diff
@@ -12,16 +13,15 @@ namespace cv {
 }
 
 struct CamFrame {
-	int width;
-	int height;
+	int width = 0;
+	int height = 0;
 	std::vector<boost::uint8_t> data;
```
Baklap4: instead of `std::vector<boost::uint8_t>` use `std::vector<std::uint8_t>` Don't forget the include `#include <cstdint>`

Tronic:
I have merged this to audiohack branch and fixed the boost::uint8_t there. Closing and deleting branch.

Issue #356, opened 2018-04-05
Sometimes menu music plays instead of a song


Tronic:
Trying to play Defy the Machine, about 50 % of times when I start the song, clock stays at zero and main theme music plays instead of the actual song. Exit and try again, then it works normally. Try again and it fails again.

Baklap4:
I haven't encountered this yet

Tronic:
Quite possibly related: sometimes menu music doesn't play in menu. Enter some other screen (e.g. practice) and come back, then it plays. (with garbled audio at the beginning?)

Tronic:
Fixed in pull request #359.

PR #359, opened 2018-04-05
Audio cleanup

Tronic:
Rewriting audio code, mostly audio.cc, for clarity and modern C++. Added debug messages for playMusic processing. Moved m_pi to util.hh and made libda use its own.

### DIFF
--- a/game/libda/fft.hpp
+++ b/game/libda/fft.hpp
@@ -12,7 +12,7 @@
 
 namespace da {
 
-	constexpr inline double TAU = 2.0 * 3.141592653589793238462643383279502884;
+	constexpr double TAU = 2.0 * 3.141592653589793238462643383279502884;
 
 	// With g++ optimization -fcx-limited-range should be used for 5x performance boost.
 	
--- a/game/libda/fft.hpp
+++ b/game/libda/fft.hpp
@@ -9,11 +9,10 @@
 #include <cstddef>
 #include <vector>
 
-extern const double m_pi;
-
-const double m_tau = (2.0 * m_pi);
 
 namespace da {
+
+	constexpr inline double TAU = 2.0 * 3.141592653589793238462643383279502884;
 
 	// With g++ optimization -fcx-limited-range should be used for 5x performance boost.
 	
@@ -32,7 +31,7 @@ namespace da {
 				const std::complex<T> temp = data[i + M] * w;
 				data[M + i] = data[i] - temp;
 				data[i] += temp;
-				w *= std::polar<T>(1.0, -m_tau / N);
+				w *= std::polar<T>(1.0, - TAU / N);
 			}
 		}
 	};
--- a/game/util.hh
+++ b/game/util.hh
@@ -3,7 +3,6 @@
 #include <cstdint>
 #include <limits>
 #include <vector>
-#include <stdexcept>
 
 constexpr double TAU = 2.0 * 3.141592653589793238462643383279502884;  // https://tauday.com/tau-manifesto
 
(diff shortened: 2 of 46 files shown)


Tronic:
The improved debug shows me that sometimes songs get stuck on preloading. Next will need to look into ffmpeg.cc.

Also, a partially related problem, someone calls playMusic("menu.ogg") when a song is chosen and the user is moving from screen_songs to screen_sing, right before the actual song is started. Apparently configuration is also saved at this point. I did not yet look what exactly is going on but certainly we shouldn't be playing menu.ogg there.

Tronic:
Looks like the merge fixed the wrong song playing bugs. I cannot reproduce issue #356 anymore.

It still tries to play menu.ogg where it shouldn't.

game/libda/fft.hpp
```diff
@@ -9,12 +9,11 @@
 #include <cstddef>
 #include <vector>
 
-extern const double m_pi;
-
-const double m_tau = (2.0 * m_pi);
 
 namespace da {
 
+	constexpr double TAU = 2.0 * 3.141592653589793238462643383279502884;
```
Baklap4: Please define pi somewhere where all classes and namespaces can use it.
Now the code is 'cluttered' with 2 random pi digits. Effected place here and `game/util.hh#L8` als i wouldn't make it a fixed digit i'd calculate it like the following: `constexpr double pi() { return std::atan(1)*4; }`
`std::atan()` is defined in `#include <cmath>`

game/util.hh
```diff
@@ -5,6 +5,8 @@
 #include <vector>
 #include <stdexcept>
 
+constexpr double m_pi = 3.141592653589793238462643383279502884;
```
Baklap4: Please define pi somewhere where all classes and namespaces can use it.
Now the code is 'cluttered' with 2 random pi digits. Effected place here and `game/libda/fft.hpp#L15` als i wouldn't make it a fixed digit i'd calculate it like the following: `constexpr double pi() { return std::atan(1)*4; }`
`std::atan()` is defined in `#include <cmath>`

Baklap4:
Also a merge conflict

game/util.hh
Tronic: It is recommended to use plain digits because math functions might be slightly inaccurate for performance reasons. Also, atan is not constexpr so it cannot be evaluated at compile time (by c++ standard).

As for da::TAU - I want to keep libda a stand-alone library (that can be used by projects other than Performous). Also, including libda headers in parts of Performous that don't use libda doesn't seem like a good idea, so this why Performous uses the definition from util.hh.

I think that m_pi should be renamed to something that doesn't look like a member variable. Did you try "PI" already, or is that name already taken by other libraries? I suppose that we could also in Performous only define TAU, because most of the time we really want 2pi. See https://tauday.com/tau-manifesto

Tronic:
I found that commit 019216b2888c9950ebdefead05c7816cfb062a8b is the culprit for menu.ogg playing. Can someone explain how this audio backend switching is supposed to work? It looks a bit fishy (1) that configuration.cc function named writeConfig() switches audio things at all and (2) exiting songs screen would need to write config.

Tronic:
@Baklap4 I have removed m_pi. Now util.hh has TAU constant instead, and libda still keeps its own da::TAU that is not used by Performous directly.

Tronic:
I'm off for the weekend. Please merge this if it seems OK, don't need to wait for me. This branch is getting quite bloated so there are frequent conflicts if any other changes are done on master.

PR #360, opened 2018-04-06
Fix pause menu for instruments: no more duplicate players

Tronic:
This seems to fix the problem of instruments and dance players rejoining after pause menu is used. To reproduce:

1. Start guitar/drums/dance song
2. Join the game and wait until the song starts, then hit Esc
3. On the menu, choose Continue

Expected result: gameplay continues
Actual result: another player joins and the old one stays inactive; the controller controls the new one

This has been broken for years. This little change (not clearing existing devices/orphans) seems to fix it but more testing is needed to determine if that change also breaks something.

### DIFF
--- a/game/controllers.cc
+++ b/game/controllers.cc
@@ -267,10 +267,6 @@ struct Controllers::Impl {
 	void enableEvents(bool state) {
 		m_eventsEnabled = state;
 		Hardware::enableKeyboardInstruments(state);
-		if (!state) {
-			m_orphans.clear();
-			m_devices.clear();
-		}
 	}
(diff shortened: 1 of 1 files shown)


Tronic:
I cannot test with actual instruments or dance pads, or in multiplayer. So, please test properly before approving.

There doesn't appear to be any (open) issue about this -- which itself is surprising because the problem happens all the time at least with keyboard instruments.

Tronic:
Confirmed working with multiplayer, and also confirmed that the old code misbehaved with joysticks too.

PR #362, opened 2018-04-09
Fullscreen mode fixed on MacOS and cleanup of mainloop and video driver

Tronic:
Added Ctrl+Cmd+F shortcut for fullscreen toggle and made green button work better. After restoring via green button the window is borderless but there are no other known problems. Now window size is restored properly after coming out of full screen mode on HiDPI devices.

Needs testing on Windows and Linux to make sure that the resize() functionality is correct there too.

### DIFF
--- a/game/main.cc
+++ b/game/main.cc
@@ -66,42 +66,44 @@ static void signalSetup() {
 /// can be thrown as an exception to quit the game
 struct QuitNow {};
 
-static void checkEvents(Game& gm) {
+static void checkEvents(Game& gm, Time eventTime) {
 	if (g_quit) {
 		std::cerr << "Terminating, please wait... (or kill the process)" << std::endl;
 		throw QuitNow();
 	}
+	Window& window = gm.window();
 	SDL_Event event;
-	while(SDL_PollEvent(&event) == 1) {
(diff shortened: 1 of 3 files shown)


game/main.cc
```diff
@@ -66,43 +66,45 @@ static void signalSetup() {
...
-			break;
-		  case SDL_KEYDOWN: {
-			int keypressed  = event.key.keysym.scancode;
-			uint16_t modifier = event.key.keysym.mod;
-			if (((keypressed == SDL_SCANCODE_RETURN || keypressed == SDL_SCANCODE_KP_ENTER) && modifier & KMOD_ALT) || keypressed == SDL_SCANCODE_F11) {
+		auto type = event.type;
+		if (type == SDL_WINDOWEVENT) {
+			auto t = event.window.event;
+#ifdef __APPLE__
+			// Sent by MacOS green button (full screen mode)
```
Lord-Kamina: Please use the platform class that can even provide runtime checks to see whether we're running on mac.

Otherwise, I think it's ok but cannot test it for now.

Tronic:
There appear to be some bugs with this. I'll try to look into it tomorrow.

PR #363, opened 2018-04-10
Capture stderr output into logger system

Tronic:
This avoids spam from ffmpeg, alsa or other libraries that might write their own messages directly to stderr. We replace stderr with a pipe and rewrite everything as log messages to std::clog (using stderr/info log level). Additionally, std::cerr is changed to use normal stderr so that any messages printed to it actually go straight to stderr.

### DIFF
--- a/game/log.cc
+++ b/game/log.cc
@@ -5,6 +5,7 @@
 #include <boost/filesystem/fstream.hpp>
 #include <boost/iostreams/stream_buffer.hpp>
 #include <iostream>
+#include <memory>
 #include <mutex>
 #include <stdexcept>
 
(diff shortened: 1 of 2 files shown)


Tronic:
Tested working on Mac and Linux and it should change nothing on Windows. Logger still has some race conditions when multiple threads write to std::clog simultaneously but that is unrelated to my changes and beyond what I can fix.

game/log.cc
```diff
@@ -42,6 +45,46 @@
...
+#include <unistd.h>
+#include <future>
+struct StderrGrabber {
+	boost::iostreams::stream<boost::iostreams::file_descriptor_sink> stream;
+	std::streambuf* backup;
+	std::future<void> logger;
+	StderrGrabber(): stream(dup(STDERR_FILENO), boost::iostreams::close_handle), backup(std::cerr.rdbuf()) {
+		std::cerr.rdbuf(stream.rdbuf());  // Make std::cerr write to our stream (which connects to normal stderr)
+		int fd[2];
+		pipe(fd);  // Create pipe fd[1]->fd[0]
```
Baklap4: @Tronic getting a warning here that the result is not used. Is this desired behavior?

Tronic: @Baklap4 Fixed in commit 796c5b9c4cb0142e149b0ee6a8a6db6170b57065 of pull request #367.

Lord-Kamina:
I haven't looked at it in detail (much less tested since I still don't have
my new computer) but I'd really like to keep all ifdefs and similar
behavior either confined to platform.cc/hh or eventually separating
implementation as is done for execname.cc and fs.cc.

PR #367, opened 2018-04-11
Logger

Tronic:
A bigger reworking of logger. NOT READY FOR MERGE but here for comments. As can be seen from testing messages on exit, it is still not thread-safe (messages get repeated). Perhaps we need to give up using std::clog in order to fix it -- I honestly don't know at this time -- but for whatever reason random repeats seem to occur within cerr and/or boost streambuf whenever multiple threads write to clog at the same time, and I don't think this can be fixed by the mutex in log.cc.

### DIFF
--- a/game/audio.cc
+++ b/game/audio.cc
@@ -423,7 +423,12 @@ struct Audio::Impl {
 	std::string selectedBackend = Audio::backendConfig().getValue();
 	Impl() {
 		populateBackends(portaudio::AudioBackends().getBackends());
-		std::clog << portaudio::AudioBackends().dump() << std::flush; // Dump PortAudio backends and devices to log.
+		{
+			std::string logmsg = "audio/notice: Audio backends:\r";
+			for (char ch: portaudio::AudioBackends().dump()) logmsg += (ch == '\n' ? '\r' : ch);
+			logmsg.back() = '\n';
+			std::clog << logmsg << std::flush;
+		}
 		// Parse audio devices from config
 		ConfigItem::StringList devs = config["audio/devices"].sl();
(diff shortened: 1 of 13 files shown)


Tronic:
The beef is really in log.cc (everything else is about fixing log messages). That was quite an interesting exercise at mixing POSIX and modern C++.

game/audio.cc
```diff
@@ -423,7 +423,12 @@ struct Audio::Impl {
 	std::string selectedBackend = Audio::backendConfig().getValue();
 	Impl() {
 		populateBackends(portaudio::AudioBackends().getBackends());
-		std::clog << portaudio::AudioBackends().dump() << std::flush; // Dump PortAudio backends and devices to log.
+		{
```
Baklap4: Why the extra braces?

game/log.cc
```diff
@@ -37,159 +38,172 @@
...
-std::string target;
-int minLevel;
-
-void writeLog(std::string const& msg) {
-	std::lock_guard<std::mutex> l(log_lock);
-	std::cerr << msg << std::flush;
-	file << msg << std::flush;
-}
+static struct Impl {
+	const std::map<std::string, int> numeric = {
```
Baklap4: Personally i'd rename this to `loglevels` or something like that. Numeric is pretty much not saying anything

game/audio.cc
Tronic: To limit the lifetime of logmsg to that block.

Tronic:
Although this branch has relevant improvements, I doubt that anyone will have time to finish the work. or even bring it up to date. Closing.

PR #368, opened 2018-04-27
Remove hard paths from mac bundling script.

Lord-Kamina:
Replaced instead with variables defined at the top of the script. In theory this should make it easier to use with non-default paths or other package managers like Homebrew.

@Tronic care to give this a try? Change the prefix to whatever's appropriate for your Homebrew install. Note, I use 10.9 as a deployment target because I've also force-built all deps from source with this target, all this is because the script is intended to make a distributable bundle. If you want it just for yourself, you  can probably just use whatever version of Mac you're using.

### DIFF
--- a/osx-utils/performous-app-build.sh
+++ b/osx-utils/performous-app-build.sh
@@ -3,6 +3,13 @@ set -o errexit
 # the very first step is to check that dylibbundler exists,
 # without it the bundle would be broken
 
+PREFIXDIR="/opt/local" # By default, the default prefix for macports, change this if you're using a different path or package manager.
+DEPLOYMENT_TARGET="10.9" # Change this if you want to target a different version of macOS.
+MAKE_JOBS=8 # Change this to change the number of jobs spawned.
+CCPATH="/usr/bin/clang" # Path to system Clang, change if you want another compiler.
+CXXPATH="/usr/bin/clang++" # Path to system Clang, change if you want another compiler.
+
+
 args=("$@")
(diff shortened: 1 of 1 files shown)


osx-utils/performous-app-build.sh
```diff
@@ -3,6 +3,13 @@ set -o errexit
 # the very first step is to check that dylibbundler exists,
 # without it the bundle would be broken
 
+PREFIXDIR="/opt/local" # By default, the default prefix for macports, change this if you're using a different path or package manager.
+DEPLOYMENT_TARGET="10.9" # Change this if you want to target a different version of macOS.
+MAKE_JOBS=8 # Change this to change the number of jobs spawned.
```
Baklap4: You can change this into `MAKE_JOBS=$(cat /proc/cpuinfo | awk '/^processor/{print $3}' | wc -l)` which will count the logical cpu cores given by the proc system which is most likely the result you want to have.

Baklap4: Besides this it looks good to me

Lord-Kamina:
I see what you mean, that won't work on a Mac though. I can fix it, sure, but it's always had the -j8 since I originally thought I was most like gonna be the only one using that script. :P

Baklap4:
@Lord-Kamina it'd be nice to make it adept the logical cpu cores so it fits every system :+1:

Lord-Kamina:
@Baklap4 please re-check this?

PR #372, opened 2018-05-16
Restore webserver notice.

Lord-Kamina:


Lord-Kamina:
@nieknooijens the function checks the value of that variable to determine the text of the alert. Without that line (originally before playing notice.ogg) the value was always 0 so no alert was displayed

nieknooijens:
@Lord-Kamina yeah I was more referencing the fact that the fix was only a single line.

PR #382, opened 2018-06-13
Boost libraries required by cpprestsdk

Lord-Kamina:
CPPRestSDK depends on boost thread and chrono, so for all our effort to get rid of them, we need to include them if we’re building the webserver with it.

### More


Baklap4:
@Lord-Kamina as of https://github.com/Microsoft/cpprestsdk/releases/tag/v2.8.0 cpprest can be build without boost by setting `CPPREST_EXCLUDE_WEBSOCKETS` in CMake while building it.
We use the latest right now which is 2.10.1
> Merged #30 which adds a CPPREST_EXCLUDE_WEBSOCKETS option to CMake to allow removing all OpenSSL and Boost dependencies.

Lord-Kamina:
And performous is not affected at all by this? I'm asking because it was performous that gave me a linking error in the first place.

Baklap4:
Performous should not be affected by this as we don't use websockets. We use a http_listener and create our own threads which the server runs on (by using `std::thread`) but i haven't tried it with explicitly giving that option to the build process of cpprestsdk

Lord-Kamina:
@Baklap4 can you please re-approve this since we already established that even building with that options maintains boost dependencies?

Lord-Kamina:
@Baklap4 @nieknooijens please somebody approve this?

Issue #327, opened 2018-03-02
Loud pops when starting/restarting audio.


Lord-Kamina:
I don't know if anybody else has noticed this?

Sometimes when the audio starts or is destroyed, it makes a really loud pop... I was wondering whether it might be possible to fix this?

Baklap4:
Is this with current master?

Lord-Kamina:
Yup.

Lord-Kamina:
Check this, here's a few samples:

https://www.dropbox.com/s/zflbchxwtaqg9m4/popping.mp4?dl=1

Baklap4:
@Lord-Kamina is this still happening?

Baklap4:
Closing this due to inactivity and bugfix branch already being merged.

PR #384, opened 2018-06-29
Bug fix/audio fixes

Lord-Kamina:
### What does this PR do?

Chiefly, fixes the issue with Pa_OpenStream reporting invalid number of channels in windows. But it also fixes a couple other things I found along the way.

Succintly:

- Implement an additional member to the Devices to make sure we refer to them by their global index, instead of the one relative to their respective backend.
- Change the way the "mics" parameter is parsed from the schema.
- Correct a bug that would result in always using the first device when a schema or config entry did not have a "dev" parameter.
- Eliminate (I think) the audio artifacts when starting/restarting playback.

### Closes Issue(s)

#327 (at least, I think)

### More

### DIFF
--- a/game/CMakeLists.txt
+++ b/game/CMakeLists.txt
@@ -144,6 +144,7 @@ if(UNIX AND NOT APPLE)
 endif()
 
 if(WIN32)
+	add_definitions("-DEPOXY_SHARED")
 	option(MXE_HACK "Features horrible hacks, but is able to compile a static performous.exe (that may not work)." OFF)
 	mark_as_advanced(MXE_HACK)
 	if(MXE_HACK)
--- a/game/libda/portaudio.hpp
+++ b/game/libda/portaudio.hpp
@@ -121,12 +121,10 @@ namespace portaudio {
 			throw std::runtime_error("No such device.");
 		}
 		DeviceInfo const& findByChannels(bool output, unsigned num) {
-			// Try name search with full match
 			for (auto const& dev: devices) {
-			unsigned reqChannels = output ? dev.out : dev.in;
-			if (reqChannels >= num) { return dev;  }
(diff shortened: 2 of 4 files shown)


game/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ if(UNIX AND NOT APPLE)
 endif()
 
 if(WIN32)
+	add_definitions("-DEPOXY_SHARED")
```
Baklap4: Welcome back old friend <3

game/libda/portaudio.hpp
```diff
@@ -118,6 +120,15 @@ namespace portaudio {
 			}
 			throw std::runtime_error("No such device.");
 		}
+		DeviceInfo const& findByChannels(bool output, unsigned num) {
+			// Try name search with full match
+			for (auto const& dev: devices) {
+			unsigned reqChannels = output ? dev.out : dev.in;
```
Baklap4: Indents

game/libda/portaudio.hpp
```diff
@@ -223,6 +234,9 @@ namespace portaudio {
 		  PaStreamCallback* callback = NULL,
 		  void* userData = NULL)
 		{
+			if (output != nullptr) {
```
nieknooijens: little inconsistency here, you're using nullptr, but elsewhere NULL is used. might be a good idea to put these in the code standard.

Lord-Kamina: From what I understand, nullptr is generally to be preferred when working with pointers. So IMO we should update those other NULLs.

Lord-Kamina: P.S. I already did it but will open a new PR.

nieknooijens:
indents have been fixed so @Baklap4 you can approve now.

PR #385, opened 2018-06-29
Use nullptr instead of NULL.

Lord-Kamina:
### What does this PR do?

Replace all the remaining instances of NULL with nullptr.

### Motivation
From what I understand, from c++11 onwards, nullptr is the preferred method.

### More


nieknooijens:
seems ok to me, I' ll give @Baklap4 some time to see it and then merge.

PR #388, opened 2018-07-05
Overhaul UnicodeUtil, sorting, and a couple more things.

Lord-Kamina:
### What does this PR do?

Ok, this is a big one...

The most important change in this PR is an almost complete rewrite of the UnicodeUtil class:
- All methods are now static.
- Introduced Google's CED library (as a statically-linked submodule) exclusively for charset detection, since it's much more accurate than  ICU for that specific function.
- Implemented locale-aware sorting, the user should now be able to choose between secondary (takes into account accents, umlauts and such) and tertiary (same, plus also takes into account caps) collation for sorting.
- Rewrote the artist and title collation algorithms used for sorting; specifically, fields are now separated with "__" in the collation string, just to make sure not to sort across more than one field (which could be the case when using spaces or nothing)
- Articles such as "The", "A" at the beginning of an Artist or Title are now ignored when sorting: "The Police" would become "Police,The". I've added those two for English as well as four more for Spanish. They live in the schema, feel free to add more for more languages if you so desire.
   - This was actually implemented earlier, but it was broken. It's working properly now.
- Restored the signature of ```writeConfig``` to once be ```void writeConfig(bool system);```
- Added a hack so the audio subsystem is only reinitialized when saving settings if the value has actually been changed.
- Now sorting criteria for songs is saved immediately when it's changed in the song selection menu.

### Motivation

Most of the things I changed were clunky or just plain broken.

### More

### Additional Notes

TODO (hate me): We should update translation files.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "ced"]
+	path = ced
+	url = https://Lord-Kamina@github.com/performous/compact_enc_det
--- a/game/players.cc
+++ b/game/players.cc
@@ -1,4 +1,5 @@
 #include "players.hh"
+#include "unicode.hh"
 
 #include "configuration.hh"
 #include "fs.hh"
@@ -9,7 +10,6 @@
 #include <unicode/stsearch.h>
 
 UErrorCode Players::m_icuError = U_ZERO_ERROR;
-icu::RuleBasedCollator Players::icuCollator = icu::RuleBasedCollator(icu::UnicodeString(""), icu::Collator::PRIMARY, m_icuError);
 
 Players::Players():
 	m_players(),
@@ -126,7 +126,7 @@ void Players::filter_internal() {
 		else {
 			icu::UnicodeString filter = icu::UnicodeString::fromUTF8(m_filter);
 			std::copy_if (m_players.begin(), m_players.end(), std::back_inserter(filtered), [&](PlayerItem it){
-			icu::StringSearch search = icu::StringSearch(filter, icu::UnicodeString::fromUTF8(it.name), &icuCollator, nullptr, m_icuError);
+			icu::StringSearch search = icu::StringSearch(filter, icu::UnicodeString::fromUTF8(it.name), &UnicodeUtil::m_dummyCollator, nullptr, m_icuError);
 			return (search.first(m_icuError) != USEARCH_DONE);
 			});
 		}
--- a/game/songparser.cc
+++ b/game/songparser.cc
@@ -57,7 +57,7 @@ SongParser::SongParser(Song& s): m_song(s) {
 			type = XML;	// XMLPP should deal with encoding so we don't have to.
 		}
 		else {
-			convertToUTF8(m_ss, s.filename.string());
+			UnicodeUtil::convertToUTF8(m_ss, s.filename.string());
 			if (smCheck (m_ss.str())) {type = SM; } else if (txtCheck (m_ss.str())) {
 				type = TXT;
 			}
--- a/game/songs.cc
+++ b/game/songs.cc
@@ -281,18 +281,18 @@ void Songs::filter_internal() {
 		SongVector filtered;
 		if (m_filter == std::string() && m_type == 0) filtered = m_songs;
 		else {
-			MatchResult match = UnicodeUtil::getCharset(m_filter);
-			std::string charset = match.first;
+			std::string charset = UnicodeUtil::getCharset(m_filter);
 			icu::UnicodeString filter = ((charset == "UTF-8") ? icu::UnicodeString::fromUTF8(m_filter) : icu::UnicodeString(m_filter.c_str(), charset.c_str()));
+			UErrorCode icuError = U_ZERO_ERROR;
 			std::copy_if (m_songs.begin(), m_songs.end(), std::back_inserter(filtered), [&](std::shared_ptr<Song> it){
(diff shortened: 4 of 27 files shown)


Lord-Kamina:
Note: The CI is failing because it has an ancient version of ICU. I'm trying to fix that.

game/players.cc
```diff
@@ -126,7 +126,7 @@ void Players::filter_internal() {
 		else {
 			icu::UnicodeString filter = icu::UnicodeString::fromUTF8(m_filter);
 			std::copy_if (m_players.begin(), m_players.end(), std::back_inserter(filtered), [&](PlayerItem it){
-			icu::StringSearch search = icu::StringSearch(filter, icu::UnicodeString::fromUTF8(it.name), &icuCollator, nullptr, m_icuError);
+			icu::StringSearch search = icu::StringSearch(filter, icu::UnicodeString::fromUTF8(it.name), &UnicodeUtil::m_dummyCollator, nullptr, m_icuError);
```
nieknooijens: dummy collator?

game/songparser.cc
```diff
@@ -57,7 +57,7 @@ SongParser::SongParser(Song& s): m_song(s) {
 			type = XML;	// XMLPP should deal with encoding so we don't have to.
 		}
 		else {
-			convertToUTF8(m_ss, s.filename.string());
+			UnicodeUtil::convertToUTF8(m_ss, s.filename.string());
```
nieknooijens: this probably fixes the webserver text encoding issues, nice! :smiley:

game/songs.cc
```diff
@@ -381,7 +403,21 @@ void Songs::sortChange(int diff) {
 	if (m_order < 0) m_order += orders;
 	RestoreSel restore(*this);
 	config["songs/sort-order"].i() = m_order;
+	switch (m_order) {
+		case 1:
```
nieknooijens: add //fallthrough, or //falls through to indicate this was on purpose and to suppress -Wfallthrough from GCC

game/players.cc
Lord-Kamina: Yeah, it's a collator set to PRIMARY collating, with no rules of its own. Basically, it's only used for searching.

game/songparser.cc
Lord-Kamina: Hopefully.

Baklap4:
why are we still using this ancient version of ubuntu though? Personally i think the CI should have an up to date version so we can build the latest versions. And maybe a secondairy CI to check for compatibility as that's the only thing the CI is currently used for

game/songs.cc
Lord-Kamina: Should I use ```[[fallthrough]]```? (will that work, being c++17?) or literally ```//fallthrough```

Lord-Kamina:
@Baklap4 Because we'd have to move to a different CI I guess. Anyway, since ICU has virtually no dependencies I'm working on making a ppa to install it on trusty, but it's taking me a bit to learn the ropes since it's my first time doing it.

Baklap4:
@Lord-Kamina i'll create an issue for updating or moving the ci, so we can discuss things there 👍

game/songs.cc
nieknooijens: a simple comment is good enough (at least for GCC).
from the manual:
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

C++17 provides a standard way to suppress the -Wimplicit-fallthrough warning using [[fallthrough]]; instead of the GNU attribute. In C++11 or C++14 users can use [[gnu::fallthrough]];, which is a GNU extension. Instead of these attributes, it is also possible to add a fallthrough comment to silence the warning. The whole body of the C or C++ style comment should match the given regular expressions listed below. The option argument n specifies what kind of comments are accepted:

    -Wimplicit-fallthrough=0 disables the warning altogether.
    -Wimplicit-fallthrough=1 matches .* regular expression, any comment is used as fallthrough comment.
    -Wimplicit-fallthrough=2 case insensitively matches .*falls?[ \t-]*thr(ough|u).* regular expression.
    -Wimplicit-fallthrough=3 case sensitively matches one of the following regular expressions:
        -fallthrough
        @fallthrough@
        lint -fallthrough[ \t]*
        [ \t.!]*(ELSE,? |INTENTIONAL(LY)? )?
        FALL(S | |-)?THR(OUGH|U)[ \t.!]*(-[^\n\r]*)?
        [ \t.!]*(Else,? |Intentional(ly)? )?
        Fall((s | |-)[Tt]|t)hr(ough|u)[ \t.!]*(-[^\n\r]*)?
        [ \t.!]*([Ee]lse,? |[Ii]ntentional(ly)? )?
        fall(s | |-)?thr(ough|u)[ \t.!]*(-[^\n\r]*)? 
    -Wimplicit-fallthrough=4 case sensitively matches one of the following regular expressions:
        -fallthrough
        @fallthrough@
        lint -fallthrough[ \t]*
        [ \t]*FALLTHR(OUGH|U)[ \t]* 
    -Wimplicit-fallthrough=5 doesn’t recognize any comments as fallthrough comments, only attributes disable the warning. 

The comment needs to be followed after optional whitespace and other comments by case or default keywords or by a user label that precedes some case or default label.

switch (cond)
  {
  case 1:
    bar (0);
    /* FALLTHRU */
  default:
    …
  }

Lord-Kamina:
Semaphore's working again, we're now running the 16.04 platform with gcc7.3.0, clang-5.0 and most dependencies backported into a PPA I made.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "ced"]
+	path = ced
+	url = https://Lord-Kamina@github.com/performous/compact_enc_det
```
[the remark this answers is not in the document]
Baklap4: You're right its fixed in the gl branch

Issue #393, opened 2018-08-27
Hide audio devices in config. 


Lord-Kamina:
This should take 30 seconds to fix but I'm out of the house so I'll put it here before I forget (or if someone can do it)
------

In the schema, audio/devices should have the hidden attribute set to true, since it's not a list meant to be changed from the interface, and it can confuse users.

PR #397, opened 2018-09-07
schema.xml: Hide audio/devices from user interface

Lord-Kamina:
### What does this PR do?

Hides an entry that is used for internal guesses about devices we can use. It has no business being in the interface, as it only ends up confusing people.

### Closes Issue(s)

#393 

### Motivation

Somebody flustered about being unable to get performous to work because they were looking at this entry instead of the Audio Devices screen.

### More


Baklap4:
closes #393

PR #400, opened 2018-10-04
Use libswresample instead of libavresample, which is deprecated in FFmpeg

### LINKED ISSUE #396: libavresample is deprecated, switch to libswresample
libavresample is deprecated in current versions of FFMPEG and not built anymore by default. It should be replaced by libswresample, providing a very similar API.
If support for older Linux versions (like Ubuntu 14, which ships with libav) is required a wrapper could be easily built that supports both libraries, but usually libavresample can be simply replaced.

dos1:
This is an Arch Linux patch by alucryd adapted to current master.
https://git.archlinux.org/svntogit/community.git/tree/trunk/performous-libswresample.patch?h=packages/performous&id=bbef3c0770c2952cd0041592edaf87d6d3a31041

Closes #396


nieknooijens:
so.... what was the fight about again?
it looks almost exactly the same :confused: 
the code looks good but i have to chgeck whether this compiles and runs on ubuntu first before approving. (which might take a day or 3 since I have an event this weekend)

Lord-Kamina:
I'll see if I can test it later.

dos1:
From what I can see it should work since Ubuntu 16.04.

Lord-Kamina:
It's passing on the CI at least so that's something. I'll still try later to make a build and test that it actually works as expected.

Lord-Kamina:
Appears to be behaving on Mac as well, haven't had time to boot ubuntu yet though.

Lord-Kamina:
Haven't had the time to do this, got a trauma training-course mandated by work (yesterday and today) but I should get a minute to do it tonight or tomorrow at some point

PR #403, opened 2018-10-08
Removed boost::noncopyable and replaced with c++11 variant

Baklap4:
### What does this PR do?
Removes `boost::noncopyable` from our codebase since it can be replaced by `constructor = delete`

### Closes Issue(s)
N/A

### Motivation
It's good habbit to clean up some old code.

### DIFF
--- a/game/3dobject.hh
+++ b/game/3dobject.hh
@@ -2,7 +2,6 @@
 
 #include "fs.hh"
 #include "glutil.hh"
-#include <boost/noncopyable.hpp>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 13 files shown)


game/3dobject.hh
```diff
@@ -2,24 +2,26 @@
 
 #include "fs.hh"
 #include "glutil.hh"
-#include <boost/noncopyable.hpp>
 #include <memory>
 #include <string>
 
+//#include "noncopyable.hh"
```
yoda-jm: remove

Baklap4: Ahh forgot that one thanks!

PR #414, opened 2018-10-23
msvc build fixes

swurzinger:
### What does this PR do?
fix some issues that prevent successful compilation with MSVC.

### Motivation
build errors on MSVC.

### Additional Notes
Tested with Visual Studio 2017 15.8.7 (MSVC version 19.15.26730)


game/screen.hh
```diff
@@ -133,6 +133,7 @@ private:
 	std::unique_ptr<Dialog> m_dialog;
 };
 
+#ifndef _MSC_VER
```
Lord-Kamina: Why this?

game/platform.cc
```diff
@@ -39,7 +39,7 @@ int Platform::defaultBackEnd() {
 #if (BOOST_OS_WINDOWS)
 extern "C" {
 // For DWORD (see end of file)
-#include "windef.h"
+#include "windows.h"
```
Lord-Kamina: Is this a mingw vs msvc thing, or is windef deprecated or what?

CMakeLists.txt
```diff
@@ -38,6 +38,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 	else()
 		message(FATAL_ERROR "Your C++ compiler is too old, please upgrade.")
 	endif()
+else()
+  set(CMAKE_CXX_STANDARD 17)
```
Lord-Kamina: This is not really the way to do this (neither is the way we're using now though but I'm looking to fix it in a future PR)

game/screen.hh
swurzinger: seems to be unsupported or buggy in MSVC. see also https://stackoverflow.com/questions/32321006/resolving-definitions-of-specialized-static-member-variables-of-templated-classe

game/platform.cc
swurzinger: because a standalone include of windef.h throws an error (in the preprocessor):
#error "No Target Architecture"
This is avoided by simply including the whole headers as intended.

CMakeLists.txt
swurzinger: it's definitely better than setting compiler flags manually, which is currently done for gcc and clang

Lord-Kamina: Which is what I said. I've already made this somewhat better in s private branch but haven't committed it because it's sorta wrapped in a larger issue.

game/platform.cc
swurzinger: a more leightweight alternative would be to include wtypes.h or minwindef.h; but this is probably the most compatible one

game/glshader.cc
```diff
@@ -29,7 +29,7 @@ void Shader::dumpInfoLog(GLuint id) {
 	if (glIsShader(id)) glGetShaderiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 	else glGetProgramiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 
-	char infoLog[maxLength];
+	char* infoLog = new char[maxLength];
```
Tronic: Do not (ever) use new/delete[]. I suppose that std::vector should do fine here, if the buffer is too large for stack.

swurzinger: Putting the string on the stack would be fine, but the syntax is invalid for standard C++ (I'm aware that GCC allows this, but MSVC doesn't).

Lord-Kamina: What about a unique_ptr?

Tronic: It is unfortunate that VLAs (supported since C99 and planned for C++14) didn't make it into standard, and AFAIK there are no alternatives either. std::vector is the normal (least confusing) solution here. Unique ptrs only got array support in C++17 (do we still support C++14?), and it is hard to see any benefit over vectors.

game/platform.cc
Baklap4: I think this is resolved?

Baklap4:
Thanks for your contribution @swurzinger 
As of now this pull request is out of date. Also it has some comments on some issues mentioned by @Tronic  and @Lord-Kamina 

There's also a newer pull request by @allcreator https://github.com/performous/performous/pull/494 fixing the build within visual studio 2019 in combination with vcpkg and cmake

Can these be resolved? If not we'll close this pull for now

swurzinger:
I haven't taken a closer look on #494 but it seems to solve the same issue this PR intended to solve quite a while ago. So this has become obsolete (and the changes probably are as well as the base most likely has changed a lot since this was written) and therefore I'll close it.

Issue #389, opened 2018-07-05
Improve CI


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Our current CI (semaphore) is using ubuntu trusty (14.04). This is rather old and has lots of old packages.
In my opinion a CI server should be an up to date system to build your program with the latest and greatest of each of the packages. If you want to check for compatibility an older CI server can be used this is currently the main reason we're still using the current server.

I'd suggest adding an updated CI system and keep semaphore for compatibility checks.
This updated CI should

- [x] Build performous for Linux
- [x] Build performous for OSx
- [ ] Build performous for Windows

basisbit:
Maybe using docker + travis-ci you could test building with bunch of old and new environments, and also get to enjoy nice integration in github web ui, pull request requirements and so on?

Baklap4:
I was thinking of that yeah 👍 containers <3 Also makes CD available 👍

Baklap4:
We have recently merged the semaphore2 patch however we can still improve the ci:

I'd say the steps will be:

- [ ] Fetch dependencies (parallel for each system (linux, unix, win32 (cross/native)))
- [ ] Build (parallel for each system (linux, unix, win32 (cross/native)))
    - And parallel for gcc and clang (meaning this will be 6 jobs parallel)
- [ ] Deploy all versions (the previous 6) as release to github (when master is targeted) (not sure if this is nicely integrated in github tho)

lua86:
So, I've been wondering (and felt like this issue would be the best place to ask), is there interest to add AppImage artifact building to the CI build? I don't wanna read on semaphore 2 and go through the effort if you're unlikely to refuse my merge request.

Lord-Kamina:
I say yes, but I don't really use Linux so 🤷🏼‍♂️

Baklap4:
> AppImage artifact building

@lua86 We currently have an open PR to support appimage #642 

Feel free to leave suggestion in regard of that in that pr :)

Closing this issue since semaphore can't do windows builds as of yet.

PR #415, opened 2018-10-27
Semaphore 2.0 config file.

Lord-Kamina:
### What does this PR do?

Add the initial sempahore-2.0 config file.

### More

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -1,0 +1,37 @@
+version: v1.0
+name: Performous Build on Ubuntu 18.04 
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+blocks:    
+  - name: Ubuntu 18.04 Build
+    task:
+      prologue:
+        commands:
+          - echo "Installing Performous Dependencies..."
+          - sudo apt -y update
+          - sudo apt -y install cmake gettext help2man clang gcc-7 g++-7 libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libpng-dev libjpeg-dev libxml++2.6-dev portaudio19-dev libopencv-dev libportmidi-dev libqrencode-dev libicu-dev libcpprest-dev libglm-dev
+      jobs:
+        - name: GCC-7 Build
+          env_vars:
+             - name: CC
+               value: gcc-7
+             - name: CXX
+               value: g++-7
+          commands:
+             - checkout
+             - git submodule init && git submodule update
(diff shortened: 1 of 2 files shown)


Baklap4:
I do like the way it is defined though, very much gitlab alike

Lord-Kamina:
Yeah I was just reading the reference (took me a bit to find it). Looks nice.

Baklap4:
I'd say the steps will be:
Fetch dependencies (parallel for each system (linux, unix, win32 (cross/native)))
Build (parallel for each system (linux, unix, win32 (cross/native)))
Deploy all versions as release to github (when master is targeted) (not sure if this is nicely integrated in github tho)

Lord-Kamina:
That'd be great in the long run but I don't think it's possible for now at least. Only Ubuntu 18.04 is available for now.

Baklap4:
with ubuntu 18 its possible to at least do 4/6 iems by crosscompiling. however we need to get rid of cpprestsdk first then :o

.semaphore/semaphore.yml
```diff
@@ -12,9 +12,9 @@ blocks:
         - name: Get dependencies.
           commands:
              - checkout
-             - sudo apt-get -y update
+             - sudo apt -y update
              - sudo apt-get -y install cmake gettext help2man clang gcc-7 g++-7 libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libpng-dev libjpeg-dev libxml++2.6-dev portaudio19-dev libopencv-dev libportmidi-dev libjsoncpp-dev libqrencode-dev libicu-dev libcpprest-dev libglm-dev
```
Baklap4: Also use `sudo apt install -y package-1 package-2 package-3`

Lord-Kamina: Will change it later. Anyway this was made to try and get around the insufficient space issue which is completely a problem on semaphore's side.

Baklap4: I thought so hahah. Hmm is there an issue or such which we can follow?

Lord-Kamina: I've tracked the issue down to libsdl2, which requires udev and that in turn wants a new kernel, which causes the error with initramfs-tools. Still, that would be solved by making the /boot partition larger. I've made a thread on the semaphore Slack so if you've got Slack you can follow it there.

nieknooijens:
did @Lord-Kamina force-push or something? since the comments from @Baklap4 are outdated yet there's only 1 commit.

Lord-Kamina:
@nieknooijens yes I rebased everything into a single commit and force-pushed.

Baklap4:
Approving and linked #389 to this to upgrade our ci to be more useful for us :)

Issue #411, opened 2018-10-17
Text rendering: some relatively serious issues.


Lord-Kamina:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Sometime ago, people on Discord brought to our attention the fact they wanted to have highlighted lyrics change outline color instead of just fill color. This is a very reasonable request. This doesn't work and at the time I dismissed it as a relatively minor oversight but in trying to tackle the issue now I see (IMO) it's a pretty deep issue going to the core of how we render text.

From what I understand... we initialize OpenGLText objects, which constructs the actual text using cairo, then makes a bitmap and passes that to our surface loader. Now, when we sing, this doesn't happen again. Instead, we grab the bitmap, stretch it artificially at the syllable level (which is the cause of my having to implement a hacky mechanism to sort-of account for the increase in size and prevent syllables from colliding onto each other) and then use OpenGL to change its color. Since we already have our text made from before, we currently have no way to change the stroke color of the text.

I'm thinking maybe we need to refactor this so highlighted lyrics are actually rendered again with appropriate size, color and stroke, or if that would cause a a significant blow to performance, maybe find a better way to use OpenGL to accomplish this.

I'm not sure I'm savvy enough to get into this myself though... anybody care to weigh in and/or help?

nieknooijens:
text rendering code is indeed pretty messy, I once tried to add some asian style scrolling effect but I only got as far as filling an entire letter at a time in stead of gradually filling it like this:
https://youtu.be/mNjif02658w?t=333

Lord-Kamina:
Basically what I'm saying is, we need to figure out a better way to compose/render text.

Lord-Kamina:
This looks... Interesting? https://github.com/rougier/freetype-gl/blob/master/README.md

nieknooijens:
it does...

lol the japanese text on their example reads: "I can eat glass but I can't find it".

Lord-Kamina:
*summons @basisbit*
Do you have any experience/ideas from others projects you might be able to contribute to the discussion with?

Lord-Kamina:
Also... somebody pointed me to this: https://github.com/ocornut/imgui

Using that would be a MAJOR refactor of the code but it looks impressive IMO.

basisbit:
* I was summoned *
Not entirely sure what would be the best way for performous, but for another game I am working on, I was considering this:
[...]

PR #416, opened 2018-11-01
Here comes another monster: Text-rendering and more GL fixes.

Lord-Kamina:
### What does this PR do?

- Adds
  - Support for stroke-linecap, stroke-linejoin and stroke-miterlimit SVG attributes.
  - Revamps the code for recoloring text while singing, now enabling people to theme both fill and stroke and even changing opacity levels.
  - A handy new pair of functions in UnicodeUtil for converting text to uppercase and lowercase in an utf-8 friendly way. We should use those instead of the boost functions for other things now.
- Changes
  - Instead of iterating through every shader to update uniforms, we now use a Uniform Buffer Object to represent four different blocks of uniforms; we're using a bit more GPU memory this way (especially accounting for the idiotic offset aligning which means allocating 800 bytes for a 368 byte buffer) but we only need to update the values in one place and it will be reflected in all shaders.
  - Removed my hacky text-scaling code that added spaces to text in an attempt to account for the scaling during singing. Fixed by just making sure that the x_offset between syllables is enough to account for the size change.
  - Increased max mipmap level to 12 because... why not? I mean, it is possible to run the game in 4k now.
  - Use different texture min and mag filters for regular textures and text. For text, we'll never use bilinear (it looked awful with my recoloring code, plus from what I understand it's slower)
  - Changed the SVG lyric files a little bit; added a different stroke-color to the highlight and cleaned them up a little bit.
  - Removed some unused pieces of code from both the shaders and opengl_text.
- Fixes
  - (Hopefully) I finally got the idiotic index out of bounds error that would happen at the beginning of some songs (duets, usually). It was apparently the result of a race-condition between threads.
  - Fixes an ffmpeg deprecation warning (wonder if they'll ever get tired of trolling everybody else).
  - Fixes broken screenshots in high-dpi mode.
  - Some other minor changes, fixes and cleanup.

### Closes Issue(s)

#411 sorta.
This game would so very much benefit from some sprite batching; that should be a medium-term goal but I don't think I can do it alone.

### More

- [x] Added/updated documentation : Added some extra commenting here and there.

### DIFF
--- a/data/shaders/core.frag
+++ b/data/shaders/core.frag
@@ -14,7 +14,7 @@ out vec4 fragColor;
 layout (std140) uniform shaderMatrices {
 	mat4 projMatrix;
 	mat4 mvMatrix;
-	mat3 normalMatrix;
+	mat4 normalMatrix;
 	mat4 colorMatrix;
 };
 
--- a/data/shaders/core.frag
+++ b/data/shaders/core.frag
@@ -25,7 +25,7 @@ layout (std140) uniform lyricColors {
 	vec4 newStroke;
 };
 
-const vec4 epsilon = vec4(1e-10);
+const vec4 epsilon = vec4(1.96e-3);
 
 #ifdef ENABLE_TEXTURING
 uniform sampler2D tex;
--- a/data/shaders/core.frag
+++ b/data/shaders/core.frag
@@ -1,10 +1,6 @@
 #version 330 core
 
-out vec4 fragColor;
-
 //DEFINES
-
-uniform mat4 colorMatrix;
 
 in vData {
 	vec3 lightDir;
@@ -13,11 +9,29 @@ in vData {
 	vec4 color;
 } fragIn;
 
+out vec4 fragColor;
+
+layout (std140) uniform shaderMatrices {
+	mat4 projMatrix;
+	mat4 mvMatrix;
+	mat3 normalMatrix;
+	mat4 colorMatrix;
+};
+
+layout (std140) uniform lyricColors {
+	vec4 origFill;
+	vec4 origStroke;
+	vec4 newFill;
+	vec4 newStroke;
+};
+
+const vec4 epsilon = vec4(1e-10);
+
 #ifdef ENABLE_TEXTURING
 uniform sampler2D tex;
 #define TEXFUNC texture(tex, fragIn.texCoord)
 #else
-#define TEXFUNC vec4(1,1,1,1)
+#define TEXFUNC vec4(1.0)
 #endif
 
 #ifdef ENABLE_SPECULAR_MAP
@@ -29,6 +43,7 @@ uniform sampler2D emissionTex;
 #endif
 
 void main() {
+	mat4 colorMatrixInternal = colorMatrix;
 	vec4 frag = TEXFUNC;
 
 #ifdef ENABLE_VERTEX_COLOR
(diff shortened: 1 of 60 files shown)


data/shaders/core.frag
```diff
@@ -1,23 +1,37 @@
...
 } fragIn;
 
+out vec4 fragColor;
+
+layout (std140) uniform shaderMatrices {
+	mat4 projMatrix;
+	mat4 mvMatrix;
+	mat4 normalMatrix;
+	mat4 colorMatrix;
+} shaderMatrices;
```
nieknooijens: schermafdruk van 2018-11-01 19-31-46
this is "already defined"

data/shaders/dancenote.vert
```diff
@@ -1,19 +1,26 @@
...
+	mat4 normalMatrix;
+	mat4 colorMatrix;
+} shaderMatrices;
+
+layout (std140) uniform danceNote {
+	int noteType;
+	float hitAnim;
+	float clock;
+	float scale;
+	vec2 position;
```
nieknooijens: why do we use vec4's for nearly everything, but vec2's for dance? weren't dance notes supposed to be 3d models in the future?

game/color.cc
```diff
@@ -52,24 +52,24 @@ Color::Color(std::string const& str) {
...
-	else if (name == "fuchsia") return Color(1.0, 0.06, 127/255.0);
-	else if (name == "yellow") return Color(1.0, 1.0, 0.0);
-	else if (name == "green") return Color(0.0, 1.0, 0.0);	
-	else if (name == "red") return Color(1, 0.0, 0.0);
-	else if (name == "blue") return Color(0.0, 43.75/255.0, 1.0);
-	else return Color(0.5, 0.5, 0.5);
+	if (name == "black") return Color(3.0/255.0, 3.0/255.0, 3.0/255.0, 1.0);
+	else if (name == "gray") return Color(24.0/255.0, 24.0/255.0, 24.0/255.0, 1.0);
+	else if (name == "white") return Color(1.0, 1.0, 1.0, 1.0);
+	else if (name == "aqua") return Color(0.0, 1.0, 1.0, 1.0);
```
nieknooijens: so everything has alpha now? making it consistent with everything else. makes sense.

game/glshader.cc
```diff
@@ -49,6 +51,40 @@ void Shader::dumpInfoLog(GLuint id) {
 	std::clog << logmsg << std::flush;
 }
 
+void Shader::bindUniformBlocks() {
```
nieknooijens: We should add some comments here as I have no idea what an "uniform block" is.

data/shaders/core.frag
Lord-Kamina: What the blue fuck? Maybe I need to change the instance names between stages? 🤷🏼‍♂️

data/shaders/dancenote.vert
Lord-Kamina: Don't know honestly.

game/color.cc
Lord-Kamina: Yeah. Less margin for error IMO.

game/glshader.cc
Lord-Kamina: Uniforms organized in "interface blocks" which are essentially structs as you can see in the shaders. Cons: use more GPU memory, pros: shared state between all shaders so you only modify it once.

Lord-Kamina:
@nieknooijens Anything else you want to see changed?

nieknooijens:
not really, looks ok now.

Lord-Kamina:
I'll merge it then.

PR #421, opened 2018-11-06
Issues/tolower toupper

Lord-Kamina:
### What does this PR do?

Fixes the Unicode case-convertion functions I added in my last PR and uses them to replace Boost case-conversion functions throughout our code.

### Motivation

Unicode-awareness is always good.

### More

### Additional Notes

Note, I left a C `toupper ` in midifile.cc; this was not an oversight, but it seems to me that replacing that would be a lot more work, plus I don’t think that code runs very often and anyway I don’t know if midi can have unicode characters in section names.

### DIFF
--- a/game/controllers.cc
+++ b/game/controllers.cc
@@ -3,7 +3,7 @@
 #include "chrono.hh"
 #include "fs.hh"
 #include "libxml++-impl.hh"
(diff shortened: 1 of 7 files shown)


Baklap4:
My bad for not checking it again

PR #424, opened 2018-11-21
[WIP] Tweaking graphics

Lord-Kamina:
### What does this PR do?

For now, just replace some remaining pngs with svg. Eventually, I'm planning something bigger.

### More

### Additional Notes

Still WIP, but could you guys give it a run and give opinions?


Tronic:
It would be nice to see this merged.

Lord-Kamina:
> It would be nice to see this merged.

Yes. I have to check my branch to see again whether there was anything left to do. I was hoping in the meantime somebody might give it a test run.

Baklap4:
can we rebase this on current master to get it most up to date so we can check if it still works on current master?

Lord-Kamina:
I'll look into it when I've got time.
Spoiler: I Didn't have time during the weekend to do any coding.
Was busy translating some stuff.

game/svg.cc
```diff
@@ -34,19 +35,34 @@ void loadSVG(Bitmap& bitmap, fs::path const& filename) {
...
+	if (bitmap.ptr) throw std::logic_error("Bitmap shouldn't point to a foreign object.");
+	bitmap.copyFromCairo(surface.get());
+
 	// Change byte order from BGRA to RGBA
-	for (uint32_t *ptr = reinterpret_cast<uint32_t*>(&*bitmap.buf.begin()), *end = ptr + bitmap.buf.size() / 4; ptr < end; ++ptr) {
-		uint8_t* pixel = reinterpret_cast<uint8_t*>(ptr);
-		uint8_t r = pixel[2], g = pixel[1], b = pixel[0], a = pixel[3];
-		pixel[0] = r; pixel[1] = g; pixel[2] = b; pixel[3] = a;
+
+	for (uint8_t *ptr = &*bitmap.buf.begin(), *end = ptr + bitmap.buf.size(); ptr < end; ptr += 4) {
```
allcreater:
```suggestion
	for (uint8_t *ptr = bitmap.buf.data(), *end = ptr + bitmap.buf.size(); ptr < end; ptr += 4) {
```

Lord-Kamina: Re: moving to to dedicated function... Maybe, I don't know how the others feel about it?
Re: your proposed change, is this a change needed for MSVC? I assume you tested it and it works properly. Yes?

game/image.hh
```diff
@@ -43,7 +44,7 @@ struct Bitmap {
 	}
 	unsigned char const* data() const { return ptr ? ptr : &buf[0]; }
 	unsigned char* data() { return ptr ? ptr : &buf[0]; }
-
+	void copyFromCairo(cairo_surface_t* surface);
```
allcreater: Just a thinking: Looks like a bit of SRP violation because bitmap is an abstract image. In theory we could just move it to the separate method and all will be better. Or even change loadSVG so that it will create a new image and not try to reuse old. 

But I think it's ok to leave things as they are. It's just a refactoring proposal, not development.

game/svg.cc
allcreater: @Lord-Kamina To be honest, no, I still don't tested it yet, but vector::data is c++ 11 feature, should work almost everything: https://ru.cppreference.com/w/cpp/container/vector/data 
In theory data() method was designed specially to replace dereferencing of first element, and even replace undefined behavior when it is not present - the data will just return nullptr;

PS: let me try to build it with MXE :)

allcreater: @Lord-Kamina yes, it compiles and seems to work under MXE (after merging with master). 
So then I suggest to use vector.data() where it possible, especially if we plans to use new standards (seriously, today is already 2020, and we use 9 years-old standard :))

Lord-Kamina: I think technically we'd decided to target c++14 a while ago but still, point taken.

Tronic: Could very well go to C++17 already. Does the CI pipeline support that yet?

Lord-Kamina: I'm not certain. I think maybe semaphore 1 didn't ansd emaphore 2 did. The reason we were still using semaphore 1 is that pull requests didn't work properly on semaphore 2 but that has probably been fixed now.

Baklap4:
@Lord-Kamina can you update this to latest master so we can get this merged?

Lord-Kamina:
> @Lord-Kamina can you update this to latest master so we can get this merged?

Will do later today. I want to add and hopefully try @allcreater’s change above relating to the iterator dereferencing.

Lord-Kamina:
After merging this, we'll probably need to force-push master in order to get rid of those ridiculous amounts of empty merge commits. @performous/core-owners

Baklap4:
Conflicts currently, then we can get this merged :)

game/CMakeLists.txt
```diff
@@ -81,6 +81,9 @@ find_package(LibEpoxy 1.2 REQUIRED)
 include_directories(${LibEpoxy_INCLUDE_DIRS})
 list(APPEND LIBS ${LibEpoxy_LIBRARIES})
 
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(deps REQUIRED IMPORTED_TARGET glib-2.0 gio-2.0 gobject-2.0)
+
```
Baklap4: What for do we need pkgConf? This is mostlikely not installed within windows.

Lord-Kamina: Those find modules need pkgconfig, I believe.

Lord-Kamina: I haven't tried it again but when I originally made this PR, I had to get new find modules because not everything was being imported and these modules depende on PkgConfig.``

Baklap4:
For now we'll let it be since we don't have a native windows build right now.

Lord-Kamina:
> For now we'll let it be since we don't have a native windows build right now.

I'd be more than happy to remove it if you can find a way to do it properly without pkg-config (or even just an alternative approach to fix it on windows)

PR #440, opened 2018-12-24
Locale fix

### LINKED ISSUE #435: LANG envvar should be used to determine language, not LC_CTYPE

Bug

### What did you do?

Tried to run performous in my non-default locale (`sk_SK.utf8`).

Common way to switch program language on Linux is by setting `LANG` environment variable, e. g. `LANG=fr_FR.utf8 mc` starts Midnight commander in French. For some reason performous reads `LC_CTYPE` instead, so while `LANG="fr_FR.utf8" performous` does not switch the language to French, `LC_CTYPE="fr_FR.utf8" performous` does. `LANG` should be honoured as this is a standard.

### What did you see instead?
Looks like performous only checks for `LC_CTYPE` and ignores the rest of `LC_` variables or `LANG`.

### Output of `performous --version`:

```
Performous 1.1-766-gc4fcfecc
```

Arch Linux 64bit

Baklap4:
Fixes to use the current locale.

### Closes Issue(s)

* Closes #435

### DIFF
--- a/game/i18n.hh
+++ b/game/i18n.hh
@@ -14,7 +14,7 @@ public:
 		gen.add_messages_path(getLocaleDir().string());
 		gen.add_messages_domain(package);
 		try {
-			std::locale::global(gen(""));
+			std::locale::global(gen(boost::locale::util::get_system_locale(true)));
 		} catch (...) {
 			std::clog << "locale/warning: Unable to detect locale, will try to fallback to en_US.UTF-8" << std::endl;
 			std::locale::global(gen("en_US.UTF-8"));


Lord-Kamina:
@jose1711 Can you check whether this branch behaves as you'd expect it to?

Baklap4:
@Lord-Kamina as #435 is closed do we still need this? It'd be nice to pick the current system default

Lord-Kamina:
Yes we do need this still. Despite the fact @jose1711's issue was caused by that script and not really understanding what LANG does, our code is sort-of broken as illustrated by my comments earlier comparing both versions.

Baklap4:
Alright let me update it so it can be merged :+1:

PR #445, opened 2019-01-04
Fixes the maximize / fullscreen bug

### LINKED ISSUE #420: Fullscreen bug

Bug

### What did you do?

1. Build performous from scratch
2. Open performous (It opened in windowed mode)
3. Maximize performous window (middle button in the menu bar near the close button)
4. Performous goes into full screen mode
5. Go to settings, change full screen mode to disabled (this results in performous windowed mode being the size of desktop)
6. Save config
7. Close performous
8. Start performous
9. Close performous (essential)
10. Start performous
11. Performous is suddenly in full screen mode

I expected performous to be still in windowed mode.

### What did you see instead?
Performous was in fullscreen state needing to be changed to windowed mode again :/

Latest master

Ubuntu 1704

Baklap4:
### What does this PR do?
It restores the way maximize window was intended: to maximize a screen.

### Closes Issue(s)
* Closes #420 
* Fixes the issue with minimizing performous when another program is used on another screen

### Motivation
At the last few events i always had to go back to a previous commit to restore the feature that i can use performous on another screen and use a browser on my main laptop. Since performous otherwise would just minimize the application.

### DIFF
--- a/game/video_driver.cc
+++ b/game/video_driver.cc
@@ -96,7 +96,6 @@ Window::Window() {
 	createShaders();
 	resize();
 	SDL_ShowWindow(screen);
-	m_fullscreen = config["graphic/fullscreen"].b();
 }
 
 void Window::createShaders() {
@@ -339,14 +338,22 @@ void Window::swap() {
 void Window::event(Uint8 const& eventID) {
 	switch (eventID) {
 		case SDL_WINDOWEVENT_MAXIMIZED:
-			[[fallthrough]];
+			if (Platform::currentOS() == Platform::macos) {
+				config["graphic/fullscreen"].b() = true;
+				}
+			else { m_needResize = true; }
+			break;	
 		case SDL_WINDOWEVENT_RESTORED:
-			[[fallthrough]];
-		case SDL_WINDOWEVENT_RESIZED:
-			[[fallthrough]];
+			if (Platform::currentOS() == Platform::macos) {
+				config["graphic/fullscreen"].b() = false;
+				}
+			else { m_needResize = true; }
+			break;	
 		case SDL_WINDOWEVENT_SHOWN:
 			[[fallthrough]];
 		case SDL_WINDOWEVENT_SIZE_CHANGED:
+			[[fallthrough]];
+		case SDL_WINDOWEVENT_RESIZED:
 			m_needResize = true;
 			break;
 		default: break;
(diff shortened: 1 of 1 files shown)


Lord-Kamina:
Wait, actually it _might_ be working as intended but I'm not sure. It's really odd. Maximizing in mac looks exactly like full-screen except you still have a mouse that you can use to reveal the menu and the window titlebar...

Baklap4:
Maximize in windowed mode should just maximize the window. It shouldn't cause performous to go into fullscreen mode. Idk about osx maximize/zoom thingy since that were 2 different things if i'm correct?

Lord-Kamina:
In mac, the old "maximize" essentially fills the screen but respecting menubars and the dock; without going into actual full-screen so it essentially looks like a window resized to the whole screen (if the dock is hidden). The thing is however; to SDL, WINDOW_MAXIMIZED is triggered (on mac) when the newer full-screen button is clicked. And that results in full-screen but as I said, still showing the mouse pointer and retaining the ability to scroll to the top of the window in order to show the window titlebar. That might be intended behavior but if so, there is an issue; it doesn't save it for subsequent runs.

Lord-Kamina:
Arguably the new behavior plays nicer with the OS to be honest, but it bothers me that it's not getting saved.

Baklap4:
Hmm here on linux it does get saved o.o?

Lord-Kamina:
Maybe you need to make it check somewhere in the loop if it's Mac® maximized and if it is, set the config flag to true?

Lord-Kamina:
It seems in mac, SDL_WINDOW_MAXIMIZED is always true and both SDL_WINDOW_FULLSCREEN SDL_WINDOW_FULLSCREEN_DESKTOP are toggled by both our in-house full-screen AND the mac full-screen. Except, they behave differently.

Baklap4:
If i'm correct you can use the option keyboard button + the green window button to actually fill the screen as maximizing
This will also happen by double clicking the menu bar of the program itself

Then in regard to fullscreen or not, this should only imho be changed within the ingame settings or through some ingame keycombination.

Both above options will have the config item to be saved correctly am i right?

Lord-Kamina:
The alt version is what I was referring to as the "old maximize". It fills screen but only if the dock is set to hide, it fills width and sets height equal to (screen - dock). But to the OS that is just like resizing the window. I don't think that triggers any fullscreen or maximizing event in SDL. And when the screen is created using full-screen on mac, the behavior is different.

Baklap4:
Will check it tonight, but i think your right 👍  thanks!

Baklap4:
Alright this works 👍

PR #446, opened 2019-01-04
Feature - Open on last window position

Baklap4:
### What does this PR do?
This makes performous open at one specific coordinate.
It can be configured through the `config.xml` and the value will auto-update when the window of performous is moved. This especially is neat when having 2 or more screens.
Once the coordinate is bigger upon loading performous than the screen it'll default back to the screen which is attached.

### Motivation
To see a program open up on a place you exited is so much more userfriendly.

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -217,6 +217,18 @@ to save the current settings to XML.
 		<short>Windowed height</short>
 		<long>The height of the window.</long>
 	</entry>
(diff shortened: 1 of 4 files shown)


Baklap4:
@nieknooijens feel free to re-approve this pr. Rebased this upon current master, therefor your previous review got dismissed.

Baklap4:
I thought about that.. but since we don't need the extra data i rewrote it back to just reference values

Lord-Kamina:
Fair enough.

PR #477, opened 2019-07-08
Check for invalid window position and size.

### LINKED ISSUE #474: SIGABRT when starting, on fresh installation

Bug

### What did you do?

- sudo apt-get install performous
- performous

The GUI, at least to set up audio (as the manpage intimates)

### What did you see instead?

    $ performous                                                                                 
    logger/notice: Logging all notices, warnings and errors. Log file: /home/tglase/.cache/performous/infolog.txt
    core/notice: Performous 1.1+ starting...
      Internationalization: Enabled
      MIDI Hardware I/O:    Enabled
      Webcam support:       Disabled
    core/notice: Starting the audio subsystem (errors printed on console may be ignored).
    audio/error: Audio device 'dev="USBMIC" mics="blue,red"': No such device.
    audio/error: Audio device 'dev="Microphone" mics="*"': No such device.
    Aborted 

```
Performous 1.1+
```

Debian sid/amd64

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::unknown_option> >'
  what():  unrecognised option '--logLevel=DEBUG'
Aborted 
```

Lord-Kamina:
### What does this PR do?

Now performous should more intelligently handle multiple displays. If possible, the window will always be created at exactly the position and size it had on last launch; if not, it will try to determine which window corner(s) is/are outside the drawing space and correct them.

### Closes Issue(s)

Should close #474.

### Motivation

QA

### More


mirabilos:
Hi Arjan,

>Looks alright i'll. Can @mirabilos look if this works for him?

will have a look within the week, thanks!

mirabilos:
Erk… the Debian package is at a git snapshot of 2018-11-18, and the patch does not at all apply. This complicates testing. I guess I’ll have to {work with the maintainers,locally update} the package first…

Lord-Kamina:
Can't you just clone the PR and build it yourself?

mirabilos:
I can build and install Debian packages; anything else would be a lot of effort (set up a throw-away virtual machine first, etc.) and not be testable on my main laptop which has the issue in the first place.

Lord-Kamina:
Pretty sure you can install all the deps (most of which will already be installed if you'd installed performous before) then you just clone the PR branch and build manually. You don't need a virtual machine or anything for that! Debian packages are just a convenience but not at all necessary.

mirabilos:
Gregorio Litenstein dixit:

>Pretty sure you can install all the deps (most of which will already
>be installed if you'd installed performous before) then you just clone

Not the development packages.

>the PR branch and build manually. You don't need a virtual machine or

No, I don’t want to taint my system this way.

>anything for that! Debian packages are just a convenience but not at
>all necessary.

Debian packages make it reliable to build, install and remove software,
plus they actually make it easier(!) for me to build the software.

In addition to that, I’ll want to use the result, and others want it
as well, so, if I update the Debian package, others benefit from it
with no additional cost for me.

mirabilos:
Nope, not enough ☹

Screenshot:

[typescript.txt](https://github.com/performous/performous/files/3389213/typescript.txt)

Infolog:

[infolog.txt](https://github.com/performous/performous/files/3389214/infolog.txt)

Lord-Kamina:
I guess we need a stack trace then.

mirabilos:
Here you are.
[stacktrace.txt](https://github.com/performous/performous/files/3389225/stacktrace.txt)

mirabilos:
The latest edition of this PR (with two commits) fixes the crash for me when compiling with `-DNDEBUG` in `CPPFLAGS` *only*.

mirabilos:
Hmm, but the patch breaks starting with `LIBGL_ALWAYS_SOFTWARE=true`…

mirabilos:
You’ll need to move the `epoxy_gl_version()` check two lines down, *behind* `SDL_GL_CreateContext()` (and the corresponding `nullptr` check), otherwise it won’t work right: “You need to create a GL context and make it current before” using `epoxy_gl_version()` say the developers of the GL stuff.

Tronic:
I don't like the added complexity of making Window optional. This adds a lot of useless clutter and room for errors. If you must make it optional, keep the Window& reference elsewhere and handle optional context inside the Window class.

Lord-Kamina:
@Tronic making it optional is the only way to properly handle errors thrown by the constructor without a heavy rewrite of the class.

What do you mean handle optional context inside the Window class? In practice we never really have more than one instance of Window and it shouldn't ever actually be empty unless something goes very wrong.

The point is, we need a try block to catch the errors in the window constructor; using an optional allows me to declare the window outside the try block but call the constructor inside, where I can then handle exceptions.

data/config/schema.xml
```diff
@@ -219,13 +219,13 @@ to save the current settings to XML.
...
 		<short>X-Position of the window</short>
 		<long>The X-Position of the window.</long>
 	</entry>
 	<entry name="graphic/window_pos_y" type="int" value="0" hidden="true">
 		<ui unit=" pixels" />
-		<limits min="0" max="65000" step="256" />
+		<limits min="0" max="16384" step="8" />
 		<short>Y-Position of the window</short>
 		<long>The Y-Position of the window.</long>
 	</entry>
```
Tronic: Not sure how relevant this here is but a step size of 120 pixels instead of 256 would match all common resolutions (1280x720, 1920x1080, 1920x1200 etc).

game/main.cc
```diff
@@ -130,7 +132,15 @@ void mainLoop(std::string const& songlist) {
...
+	boost::optional<Window> window;
+	try {
+		window = boost::in_place();
+		} catch (RUNTIME_ERROR& e) {
+			std::cerr << "ERROR: " << e.what() << std::endl;
+		}
+		
+	if (!window) throw std::runtime_error("Unable to create window.");
+
 	TextureLoader m_loader;
```
Tronic: If window creation fails, you print a message and re-throw? This looks like something that could well be done within the Window constructor itself.

game/video_driver.cc
```diff
@@ -94,7 +150,6 @@ Window::Window() {
 	std::clog << "video/info: GL_RENDERER:   " << glGetString(GL_RENDERER) << std::endl;
 	// Extensions would need more complex outputting, otherwise they will break clog.
 	//std::clog << "video/info: GL_EXTENSIONS: " << glGetString(GL_EXTENSIONS) << std::endl;
-	if (epoxy_gl_version() < 33) throw std::runtime_error("OpenGL 3.3 is required but not available");	
 	createShaders();
 	resize();
 	SDL_ShowWindow(screen);
```
Tronic: Window constructor is getting huge. I'd suggest breaking it into smaller units. If you need some of them initialized in Window initlist, you may still use function-try-block to catch any errors within the constructor.

Lord-Kamina: That would probably be a better solution in the long-term. Perhaps we should make the default constructor just allocate resources but add a function that actually creates the window and GL context?

Tronic: As you mentioned earlier, that is usually messy. If at all possible, construct and fail in constructor, so that the object may only exist in completely valid state. The same goes for Game and optional Window, which would suddenly cause exception to be thrown when someone tries to use it.

I believe that in this case you can avoid any partial constructor or optional values, unless I misunderstood somehow how this patch works.

Lord-Kamina: The problem is the following: 
There are a number of reasons why the Window constructor can fail. However; if the Window is not allocated inside a try block, the exception goes unhandled and the user gets no feedback of what went wrong because the actual error is buried in the stack trace (which is what happened with the issue this PR originally meant to correct)

Options are either doing what I'm doing, doing what I'm proposing or I suppose eventually handling the exception inside the constructor, but I'm not sure whether we prefer adding try/catch blocks into individual classes rather than what I'm doing?

mirabilos:
L. Kärkkäinen dixit:

>Not sure how relevant this here is but a step size of 120 pixels
>instead of 256 would match all common resolutions (1280x720,
>1920x1080, 1920x1200 etc).

1024x600 (netbook)
1024x768 (laptop)
1280x1024 (desktop)

mirabilos:
Gregorio Litenstein dixit:

>The problem is the following: There are a number of reasons why the
>Window constructor can fail. However; if the Window is not allocated
>inside a try block, the exception goes unhandled and the user gets no
>feedback of what went wrong because the actual error is buried in the
>stack trace (which is what happened with the issue this PR originally
>meant to correct)

If -DNDEBUG is passed in CPPFLAGS, they are caught and displayed.

I wrote this in IRC ;-) the PR without those recent changes *almost*
works for me, it needs one change (move the epoxy_gl_version() check
to BELOW the SDL_GL_CreateContext() call, as it needs an initialised
OpenGL context “on the screen” or something like that, check the IRC
log from some days ago):

https://salsa.debian.org/games-team/performous/commit/d83e0791580ee7022292508eac5b3db10ca6a74e.patch

This is the full PR/patch in the version I’m using in Debian successfully:

https://salsa.debian.org/games-team/performous/blob/experimental/debian/patches/pr477.diff

Thanks,
//mirabilos
-- 
This space for rent.

https://paypal.me/mirabilos to support my work.

Lord-Kamina:
> L. Kärkkäinen dixit:
> Not sure how relevant this here is but a step size of 120 pixels instead of 256 would match all common resolutions (1280x720, 1920x1080, 1920x1200 etc).
> 1024x600 (netbook) 1024x768 (laptop) 1280x1024 (desktop)

Honestly not relevant because the resolution isn't really configurable in the UI; the resolution is chiefly changed when Maximized or when resizing the window. (my change is also irrelevant honestly for the same reason, although I guess we could leave it like you say in case we do implement resolution changing from the UI in the future)

Lord-Kamina:
So... please weigh in here @performous/core-owners re: how to solve this:

Options currently are:
1) Keep this PR.
2) Refactor the `Window` class so the constructor doesn't actually create the gl context; which would eliminate the issue (but apparently @Tronic isn't really fond of this solution either)
3) Add try...catch blocks INSIDE the `Window` constructor (I'm not fond of this because it'll only make it more messy IMO)
4) Maintain the idea from this PR but use maybe a unique_ptr instead of an optional.
5) ...I don't know, ideas?

Tronic:
I ask for someone else to take this; I won't have time to continue reviewing this week.

Lord-Kamina:
@Tronic still, what's your take on the options I laid out? (Or do you have another?). I think maybe n°4 is best and more idiomatic; I went for an optional chiefly because I've been doing a lot of Swift lately. 😂

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Lord-Kamina:
> This is not necessary. As I said in IRC, the problem with the constructor error not displayed was because I had not built with `-DNDEBUG`.
> 
> Please see https://salsa.debian.org/games-team/performous/blob/experimental/debian/patches/pr477.diff for a version of the _complete_ patch from this PR that is actually needed and that actually works, and apply that.

This IS a problem for me, because I don't expect people to use debug builds when we eventually release a new stable.

Lord-Kamina:
Anyway... I've just done the same now but in a much more idiomatic (and less complex) way, using `unique_ptr`.

mirabilos:
Gregorio Litenstein dixit:

>This IS a problem for me, because I don't expect people to use debug
>builds when we eventually release a new stable.

Hmm. I could see the error when running under gdb only, then.

Lord-Kamina:
Got rid of the "check OpenGL version before creating context" commit to avoid confusion; since that actually ended up being an error that was reverted.

Lord-Kamina:
Can we rebase and merge this?

Lord-Kamina:
This should be up-to-date now; it'd be great if we could get it merged.

PR #491, opened 2020-01-17
Fix audio device selection

Tronic:
Avoids errors about device not having enough channels when identically named but separate devices appear for audio input and output. Choose the first matching device *with the requested amount of channels*


Lord-Kamina:
Hadn't encountered this issue. Anyway, PR looks good on my side also

PR #492, opened 2020-01-26
Experimental support for Meson build system, code cleanup.

Tronic:
Minimal support for https://mesonbuild.com/ which might avoid some compile/configuration issues that some users have had. Limitations: no version number ~or optional features (like MIDI drums or webcam).~

Please verify that your CMake-based builds still work before merging this.

Drops support for ~libxml++-2.6 and~ some old libav/ffmpeg versions. Unfortunately it would appear that the CI pipeline still relies on libxml++-2.6. Any possibility of upgrading it to libxml++-3.0?


Lord-Kamina:
> Drops support for libxml++-2.6 and some old libav/ffmpeg versions. Unfortunately it would appear that the CI pipeline still relies on libxml++-2.6. Any possibility of upgrading it to libxml++-3.0?

This is a big issue. I'd definitely be up for dropping it but AFAIK Ubuntu doesn't have libxml++3 even in the latest versions. Same thing with debian. If either did, we could probably just make a backport and be done with it. I'm not really sure how to make a debian package from scratch though, so unless you know how to do this, we'll probably have to keeo 2.6 support in.

Lord-Kamina:
Also, I'm not really experienced at all with meson; but we could probably make optional features optional via meson_optiond.txt or whatever that's called?

Tronic:
Now webcam and midi are autodetected, which should be good enough. Webserver could probably be added in a similar fashion.

Lord-Kamina:
Fair enough. Say, is the idea to eventually drop CMake or do you just want the extra option?

Tronic:
No idea. I just heard of Meson today and wanted to give it a try. CMake is slowly falling apart though and Meson so far feels far more sensible.

Pros:
- Easiest to use build system I've ever used
- Mostly based on pkg-config so works with little hassle
- Build script language is similar to Python (i.e. far better than CMake)
- Faster than CMake

Cons: 
- No manual override when lib detection fails (like CMake GUI)
- Build script language is *not* Python but their own making...
- Is supposed to regenerate every time you run `ninja` but I had to manually delete build dir at least once to actually get an updated config (like `rm CMakeCache.txt`).

Tronic:
I'm quite done with this, feel free to merge. Issues to be resolved if someone else has time:
- cpprestsdk detection
- git version tag into project version (now hardcoded as `git-meson`)

Lord-Kamina:
> I'm quite done with this, feel free to merge. Issues to be resolved if someone else has time:
> 
> * cpprestsdk detection
> * git version tag into project version (now hardcoded as `git-meson`)

I'll try to test it properly during the weekend and merge if I find no issues.

Baklap4:
@Lord-Kamina did you have time to test this out?
I sure didn't yet. Aside from that the PR looks good

Lord-Kamina:
Yeah no I'm arse-deep in COVID-19 suspects 😰

Baklap4:
I'd like to get this merged, not so much for the meson support, but more for the cleanup of stuff. 
As of the meson support i didn't try it yet any other @performous/core-owners did?

If we're going to get this merged (with meson support) what would we want to do?
Remove current build and completely merge to meson?

I'm not feeling it to maintain multiple builds, so how are we going to 'force' this?

Lord-Kamina:
I still haven't tested it. I keep forgetting. Will definitely try today.
I really don't want to MOVE to Meson though. I thinl it's nice to support but I prefer CMake; and even if you just like Ninja better than Make; CMake can output Ninja makefile as well as Unix Makefiles.

Tronic:
I did not intend this as a replacement but as an option, at least until it becomes clear which one has brighter future, much like how projects migrated from autoconf to CMake a decade ago.

CMake has more powerful means of locating libraries, which just cannot be replicated on Meson yet. OTOH, Meson relies on pkg-config and thus it can handle the dependency hell and custom flags of libraries that our CMake scripts are struggling to handle (and the whole CMake ecosystem seems stagnated: no one else is going to keep Pango/harfbuzz/whatever modules working so we need to keep rolling our own). Meson's big problem is that not all libraries ship pkg-config files, and on Windows pkg-config might not be available at all, and there is no way (AFAIK) to do manual override like you can do with CMake GUI.

Tronic:
This builds OK now, with system-installed aubio.

game/ffmpeg.cc
```diff
@@ -24,11 +24,6 @@ extern "C" {
 #include <libavutil/error.h>
 }
 
-// Some versions of libav does not contain this definition.
-#ifndef AV_ERROR_MAX_STRING_SIZE
-#	define AV_ERROR_MAX_STRING_SIZE 64
-#endif
-
```
Baklap4: current version we're using/supporting does have this or do not need this definition anymore?

Tronic: This is not my change, but merge from master, which no longer has that code.

Baklap4: Ahh check

Baklap4:
@Tronic do we have documentation on how to run this build? If so where can we find it, if not let's create it then get this merged 👍

Tronic:
@Baklap4 Added to https://github.com/performous/performous/wiki/Building-and-installing-from-source#meson-build

There's also Compiling.txt, should that be updated as well?

Baklap4:
Yes if you have the time feel free to do so 👍

Tronic:
> Yes if you have the time feel free to do so 👍

I don't really...

Tronic:
Note: this doesn't have any install target yet. Main binary and static data files should be simple enough to add but app launcher and other things built by CMake might get complicated.

deuteragenie (commented):
I have successfully tested the experimental Meson build in this PR on Ubuntu 21.04 (very fast build !)

deuteragenie:
Hi,

The following compiler arguments can be used to reduce significantly the number of warnings, while enabling as much as possible.  Some of them really should not be disabled (narrowing, overflow, etc.) and probably pragmas should be added on the legacy files.  Tested with gcc and clang.

`
compiler=meson.get_compiler('cpp')
if compiler.get_id() == 'gcc'
  compiler_args=compiler.get_supported_arguments(
    ['-Wall','-Wextra','-pedantic','-Wc++20-compat',
    '-Wno-deprecated-declarations',
    '-Wno-mismatched-tags',
    '-Wno-narrowing',
    '-Wno-overflow'])
else
  compiler_args=compiler.get_supported_arguments(
    ['-Wall','-Wextra','-pedantic','-Wc++20-compat',
    '-Wno-c++11-narrowing',
    '-Wno-implicit-int-float-conversion',
    '-Wno-inconsistent-missing-override',
    '-Wno-deprecated-declarations',
    '-Wno-mismatched-tags'])
endif
`

Tronic:
This should probably be merged and then someone else should tune it further in a separate PR.

Tronic:
Closing in favour of #596.

Issue #476, opened 2019-07-01
Crash on Fedora 30


jorritjorritsma:
### Do you want to request a *feature* or report a *bug*?
Bug

### What did you do?
I started the application

### What did you expect to see?
the performous application

### What did you see instead?
Performous crashed immediately with the following output
```
performous
logger/notice: Logging all notices, warnings and errors. Log file: /home/jorrit/.cache/performous/infolog.txt
core/notice: Performous 1.1+ starting...
  Internationalization: Enabled
  MIDI Hardware I/O:    Enabled
  Webcam support:       Enabled
core/notice: Starting the audio subsystem (errors printed on console may be ignored).
audio/error: Audio device 'dev="USBMIC" mics="blue,red"': No such device.
audio/error: Audio device 'dev="Microphone" mics="*"': No such device.
songs/notice: Done loading the cache. You now have 0 songs in your list.
songs/notice: Starting to load all songs from disk, to update the cache.
songs/notice: Done Loading. Loaded 0 Songs.
songs/notice: Done Caching.
/usr/include/c++/9/bits/stl_vector.h:1042: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = unsigned char; _Alloc = std::allocator<unsigned char>; std::vector<_Tp, _Alloc>::reference = unsigned char&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__builtin_expect(__n < this->size(), true)' failed.
Aborted (core dumped)
```

### Output of `performous --version`: (_What version of Performous are you using?_)
```
Performous 1.1+
```

### What is your environment & configuration (arguments, platform, ...)?
OS: Fedora 30 x86_64
System: HP EliteBook 840 G3

As Performous didn't work to my expectations on Ubuntu 19.04 I tried Fedora, hence this was a newly installed system with rpmfusion enabled.

skmqpbusfz:
I noticed the same problem after my upgrade to Fedora 30. I decided to compile the current revision myself (was e9568253). It did crash just the same.
After debigging I knew it crashed when parsing the controllers.xml specifically at line 277:
```<device regex="Harmonix Drum [Kk]it for PlayStation" />```

After deleting this controller section and the next section (same square bracket regex) the program still crashed on startup but in a different place but also calling std::regex.

[...]

PR #496, opened 2020-03-15
Workaround failing regex on FC30

OznOg:
Allows to select c++11 regex or boost on cmake invocation.

to enable at build do:
cmake -DUSE_BOOST_REGEX=true ../performous/dir/

default behaviour (without setting USE_BOOST_REGEX) is to use std::regex

This is mainly a workaround for bug #476

### DIFF
--- a/game/regex.hh
+++ b/game/regex.hh
@@ -1,0 +1,16 @@
+
+#pragma once
+
+// Note: One could expect to have boost regex removed, but for some obscure
+// reason, std::regex code crash on Fedora 30 (see bug #476). Thus, until FC is
+// fixed or deprecated, may be eeded to keep this for a little while
+
+#ifdef USE_BOOST_REGEX
+#include <boost/regex.hpp>
+using regex = boost::regex;
+namespace regex_constants = boost::regex_constants;
+#else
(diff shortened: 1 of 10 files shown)


game/regex.hh
```diff
@@ -0,0 +1,16 @@
...
+
+#ifdef USE_BOOST_REGEX
+#include <boost/regex.hpp>
+using regex = boost::regex;
+namespace regex_constants = boost::regex_constants;
+#else
+#include <regex>
+using regex = std::regex;
+namespace regex_constants = std::regex_constants;
+#endif
```
Tronic: Shouldn't this import `regex_replace` too?

OznOg: yes and regex_search as well... let me figure out how this can compile...

OznOg: well actually this is fine, thanks to the argument-dependent name lookup. Here should be ok, as there is no abiguity between boost and std (when using boost regex, all is boost::regex, when using std all is std::regex)

Baklap4: Seems this is 'resolved'? since tronic thumbed-up the response?

Issue #39, opened 2013-05-20
Using C++11 facilities to clean up code and Boost dependencies


Tronic:
We can reduce but not entirely remove dependency on Boost by replacing parts of it with stdlib equivalents. This bug is a tracker for features to replace, and for discussion about compiler compatibility.

- [ ] boost/algorithm/string.hpp: No stdlib equivalents
- [ ] boost/bind.hpp: std::bind in functional
- [ ] boost/circular_buffer.hpp: No stdlib equivalent
- [ ] boost/cstdint.hpp: Now in stdlib (cstdint)
- [ ] boost/date_time.hpp: Partial support from stdlib (chrono)
- [ ] boost/filesystem.hpp: C++ still doesn't know of files :/
- [ ] boost/foreach.hpp: C++11 built-in range-based for!
- [ ] boost/format.hpp: No direct stdlib equivalent but see std::to_string(num)
- [ ] boost/function.hpp: std::function (functional)
- [ ] boost/iostreams/stream_buffer.hpp: ???
- [ ] boost/lexical_cast.hpp: See std::to_string and std::stoi etc. (string)
- [ ] boost/noncopyable.hpp: C++11 deleted functions (but noncopyable is cleaner in code)
- [ ] boost/program_options.hpp: Never gonna get into stdlib
- [ ] boost/ptr_container/: Not in C++11 and probably not in C++14 either.
- [ ] boost/regex.hpp: The last time I tried (quite recently), the libstdc++ implementation was unusably incomplete.
- [ ] boost/smart_ptr/scoped_ptr.hpp: std::unique_ptr in memory (also replaces std::auto_ptr)
- [ ] boost/smart_ptr/shared_ptr.hpp: std::unique_ptr (if shared_ptr was only used for calling a free function of a C lib), std::shared_ptr if the resource is actually shared, C++11 move semantics instead if a resourse is only being moved.
- [ ] boost/thread/: std::thread and stdlib mutexes (no idea if stdlib implementation is good yet or if all necessary lock types are included or not)
- [ ] boost/variant.hpp: Didn't make it into C++11

We don't currently use Boost.ASIO directly but likely will, soon. Sockets didn't make it into C++11.

zosrothko:
If you need a lightweight c++ library to replace part or whole Boost, look at  [Poco](https://pocoproject.org/)

Baklap4:
slowly we're getting there: https://github.com/performous/performous/pull/339

Baklap4:
[...]

PR #499, opened 2020-03-20
Use std filesystem

OznOg:


Baklap4:
Looks promising already :) Keep up the work! :)

OznOg:
Compilation fails because filesystem support with gcc 7 and clang 5 is still experimental.
That said, both are >2 years old now, is there a possiblility to move to >gcc8 and >clang6 ?
The other option is adding more compilation flags to add 'experimental' in header and namespace, but I'm affraid this makes code really painful to read.
Could be also that I add extra compiler check and do not activate the feature for older compiler, but ther the CMakeList.txt will get some more combinatory on compilers.

Let me know what do you prefer...

Tronic:
Drop the old versions.

Baklap4:
Drop the old version, let's get some new stuff in.

Lord-Kamina:
I'd like to keep support for now but definitely think we should drop them eventually.

Baklap4:
but when is 'eventually'? in most projects i worked on 'eventually' equals never... While in this code why not clean it up?

Tronic:
The big question is, does anyone still use those old versions? Even Debian stable has GCC 8 and Clang 7.

Lord-Kamina:
If somebody is still on Ubuntu 16, they would.

OznOg:
> If somebody is still on Ubuntu 16, they would.

Are people on ubuntu 16.04 compiling performous and/or expecting new packages considering that even the compiler/stdlib support for c++17 is partial? IMHO, people with ubuntu 16 are welcome to use older packages of performous, I don't think they expect backport of the bleeding edge.

Baklap4:
> Are people on ubuntu 16.04 compiling performous 

Currently the semaphore ci is still building it, aside from that i think most developers did indeed upgrade to 17.04 or 18.04 already, that's like 2 years ago so...

Baklap4:
@OznOg can we make this up to date with latest master?
It'd be nice to support this

OznOg:
so what are the takeaways here, do we drop support for ubuntu16.04 (now 20.04 is released) ? do we move to completely to c++17?

Lord-Kamina:
Note: macOS does support `<filesystem>` from Catalina onwards.

OznOg:
Do you know how to check properly the different versions of MacOS stl?

My journey is not over, seems that ubuntu also complains...

OznOg:
legacy compilers keeps failing, do I really need to try a fix? clang 5 and gcc7 are quite old....

Lord-Kamina:
Both of them should be fully c++17 compliant, though. So, it's likely the issue is with the flags being passed to them.

Lord-Kamina:
> Both of them should be fully c++17 compliant, though. So, it's likely the issue is with the flags being passed to them.

I see I'm evidently wrong; despite they supporting the standard it seems they don't support the `<filesystem>` header. Anyway, this does not still necessary mean the end for xenial; using gcc-8 is a drop-in replacement for gcc-7 on our CI, as the PPAs we're currently using already include it and I'm sure something similar can be done about Clang.

OznOg:
or we can just say we use boost for compilers that do not have filesystem's support

Lord-Kamina:
> or we can just say we use boost for compilers that do not have filesystem's support

I don't think performous as it currently is builds on vanilla 16.04 anyway. However, with the PPAs the CI is currently using, both clang-6 and gcc-8 are available so it's not really an issue.

OznOg:
ok, great :), I need to do some cleanup and we are ready for reviews

Lord-Kamina:
For the record, this command in terminal will give you the macOS version number:
`/usr/bin/sw_vers | grep ProductVersion | cut -f2 -d: | cut -f2 -d"    "`

OznOg:
do you have any update on this?

OznOg:
FYI, I tested this using strace and logged all files accessed with boost version and with c++17. I was able to fix a few things with this mean; now outputs are matching exactly.

OznOg:
any update?

OznOg:
please give me some update on "merge or not". If you don't feel like merging it, let's just drop it, but I don't like stalled branches for ever.

Baklap4:
eey Oznog, sorry for not getting back to you earlier.
I had some issues regarding my pc and getting the parts of it delivered.
Anytime this week i'll install linux on it and test this merge request as i'd love to have this within performous.

OznOg:
@Baklap4 any update? is it worth keeping boost compat now we are on 2021 and FS support is 4 years old?

OznOg:
hum... ping?

OznOg:
ping

CMakeLists.txt
```diff
@@ -30,25 +34,33 @@ if (USE_BOOST_REGEX)
...
-	message("Detected GCC version ${GCC_VERSION}")
-	if (GCC_VERSION VERSION_GREATER 7 OR GCC_VERSION VERSION_EQUAL 7)
-		set(CMAKE_CXX_FLAGS "-std=c++14 -Wall -Wextra -fcx-limited-range ${CMAKE_CXX_FLAGS}")
-	elseif (GCC_VERSION VERSION_GREATER 4.8 OR GCC_VERSION VERSION_EQUAL 4.8)
-		set(CMAKE_CXX_FLAGS "-std=c++1y -Wall -Wextra -Wno-attributes -Wno-deprecated-declarations -fcx-limited-range ${CMAKE_CXX_FLAGS}")
-		message(WARNING "Your GCC is rather old, disabling some useful compiler warnings (GCC 7+ or Clang 5+ recommended for proper C++17 support).")
+if (APPLE)
+	set (MACOS_SUPPORT_FILESYSTEM FALSE)
+	if (${CMAKE_OSX_DEPLOYMENT_TARGET})
+		message (STATUS "Chcking macOS deployment target... ${CMAKE_OSX_DEPLOYMENT_TARGET}")
```
Baklap4, line 42: Typo: "Chcking" -> "Checking"

Baklap4 (dismissed):
Would be nice to fix this typo, but aside from that i couldn't find things and it's working alright.
Did anyone @performous/core-owners  else end up having problems?

In regard of boost compat.. there's another issue to clean up boost dependencies, let's fix that there? (https://github.com/performous/performous/issues/39)

CMakeLists.txt
OznOg: done

OznOg: done

Baklap4 (approved):
LGTM

OznOg:
:champagne:

PR #504, opened 2020-03-21
Rework CMake to call waf directly

OznOg:
### What does this PR do?
Rework cmake build of libaubio so that it may work everywhere (hopefully)

### Closes Issue(s)

nop

### Motivation
I've been having problems to compile master

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,8 @@ if (APPLE)
 	set(ACCELERATE_COMMAND "--enable-accelerate")
 	set(AUBIO_CFLAGS "${CMAKE_C_FLAGS} -iframeworkwithsysroot /System/Library/Frameworks/ -mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET} -isysroot '${CMAKE_OSX_SYSROOT}/'")
 else()
-	set(ACCELERATE_COMMAND "--enable-blas")
+	set(ACCELERATE_COMMAND "--enable-atlas")
+        set_property(GLOBAL PROPERTY BLA_VENDOR OpenBLAS)
 	set(AUBIO_CFLAGS "${CMAKE_C_FLAGS}")
 endif()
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -91,20 +92,23 @@ if(NOT FFTW3_FOUND OR NOT BLAS_FOUND)
...
-    UPDATE_COMMAND ""
-	TMP_DIR tmp
-	CONFIGURE_COMMAND "${CMAKE_SOURCE_DIR}/3rdparty/build_aubio.sh" ${ACCELERATE_COMMAND} <INSTALL_DIR> ${AUBIO_CFLAGS}
-	BUILD_COMMAND ""
+        BINARY_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/
+        INSTALL_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-install/
+        UPDATE_COMMAND ""
+	TMP_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/tmp/
+	CONFIGURE_COMMAND ""
+	BUILD_COMMAND cmake -E env CFLAGS=${AUBIO_CFLAGS} ${CMAKE_SOURCE_DIR}/3rdparty/aubio/waf
```
Baklap4: I'd change `cmake` to `${CMAKE_COMMAND}`
As per cmake docs;
> This is the full path to the CMake executable cmake(1) which is useful from custom commands that want to use the cmake -E option for portable system commands. (e.g. /usr/local/bin/cmake)

OznOg: yes you're right

CMakeLists.txt
```diff
@@ -91,26 +92,29 @@ if(NOT FFTW3_FOUND OR NOT BLAS_FOUND)
...
+        BINARY_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/
+        INSTALL_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-install/
+        UPDATE_COMMAND ""
+	TMP_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/tmp/
+	CONFIGURE_COMMAND ""
+	BUILD_COMMAND ${CMAKE_COMMAND} -E env CFLAGS=${AUBIO_CFLAGS} ${CMAKE_SOURCE_DIR}/3rdparty/aubio/waf
+                  ${ACCELERATE_COMMAND} --disable-tests --notests --verbose
+                  --enable-fftw3f --disable-sndfile --disable-avcodec --disable-double
+                  --disable-samplerate --disable-docs --disable-wavread --disable-wavwrite
+                  --disable-tests --notests --disable-examples --disable-apple-audio
```
Lord-Kamina: You shouldn't hard disable audio engines here.

CMakeLists.txt
```diff
@@ -91,26 +92,29 @@ if(NOT FFTW3_FOUND OR NOT BLAS_FOUND)
...
+                  --disable-fat --disable-jack --libdir=lib
```
Lord-Kamina: Same as above

CMakeLists.txt
OznOg: those are options from the original (upstream) shell script; can you be more specific on what are the flags you want to see here?

game/CMakeLists.txt
```diff
@@ -170,8 +170,9 @@ if(WIN32)
   target_link_libraries(performous wsock32 ws2_32)
 endif()
 
-target_link_libraries(performous ced)
-list(APPEND LIBS ${AUBIO_INSTALL_DIR}/lib/libaubio.a)
+target_link_libraries(performous ced aubio)
+target_link_directories(performous BEFORE PRIVATE ${AUBIO_INSTALL_DIR}/lib/)
```
Baklap4: target_link_directories is only available only available from cmake 3.13 or higher.

OznOg: that is bad news, we'll have to do it in a ugly way then

OznOg: So, do we eventually need aubio to be built on windows?

Baklap4: Eventually yeah, however we're not using the pre-build crosscompiled libraries

OznOg:
so where do we stand with this? Is it worth keeping it?

Baklap4:
Thanks for the PR @OznOg !

PR #506, opened 2020-03-22
Rework screen path browsing

OznOg:
### What does this PR do?

It fixes a crash when browsing  .. when in the / directory;
And as I saw there was a fixme on sorting menu entries, I made a patch for this as well

### Closes Issue(s)
none


Baklap4:
Did something change?

PR #510, opened 2020-03-25
Add Python implementation of xml_gettext script.

Tronic:
Still not actually tested but should work now.


Baklap4:
Might be worth to change lang/readme.md accordingly or is it still the same regarding [setting up the extractor](https://github.com/performous/performous/blob/master/lang/README)
> Configuring the Performous XML parser
> -------------------------------------
> These steps enable translation of the XML locate bits.
>  1. Click Edit -> Preferences
>  2. Click the Parsers tab
>  3. Click the New button and fill out the form like this:
       Language: Performous XML
       List of extensions: *.xml
       Parser command: $SOURCE_ROOT/tools/scripts/xml_gettext.sh %o %F
       An item in keywords list:
       An item in input files list: %f
       Source code charset:
     Where $SOURCE_ROOT is the path to the performous sources root.
>  4. Click Okay

Tronic:
On Windows you need `c:/.../python.exe c:/.../xml_gettext.py %o %F` which is not quite optimal, so it would probably be better to compile this into an EXE file.

```
pip install pyinstaller
pyinstaller -y -F xml_gettext.py
```

This builds `dist/xml_gettext.exe` which can be then distributed with Performous and used without Python. On Mac and Linux the script itself should be runnable, provided that there is Python installed on the system.

Did you test if this works yet? I don't have poedit installed, so I didn't, and I am not quite sure what is expected to be in po file header. Now it does specify the charset at least, and msgcat didn't complain.

Baklap4:
If i try to update `lang/nl.po` and synchronize with the script i'm currently getting this output
```
26/03/2020 11:04:18: Traceback (most recent call last):
	File "C:\Users\aspeiard\Desktop\Persoonlijk\performous-master\gettext.py", line 36, in <module>
	assert pofile.endswith(".po"), pofile
26/03/2020 11:04:18: AssertionError: C:\Users\aspeiard\AppData\Local\Temp\poeD8A4.tmp\legacy_Performous_XML_extracted.pot
26/03/2020 11:04:18: Entries in the catalogue are probably incorrect.
26/03/2020 11:04:18: Updating the catalogue failed. Click on 'Details >>' for details.
```
Using poedit 2.3 build 5890

With the executeable i'm getting this error 
image

Baklap4:
Seems to work! Can you remove the tools/script/gettext_xml.**sh**
Also as for the exe, i think it can be deleted since it's just a wrapper and nuisiance around gettext_xml.py which has to be kept up to date aswell, which probably won't happen.

To make this pr complete change the lang/readme file aswell to have .sh changed into .py

Tronic:
Also removed the shell script and the EXE. I didn't include docs for Windows. Be sure to rebase squash when merging to avoid including the EXE in history.

Issue #454, opened 2019-01-10
Ctrl-Fx should provide visual feedback


jose1711:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Currently keys for tuning latency during song playback provide no visual output making it more difficult to recognize whether the shortcut is recognized by the app or what the current setting is. An OSD information showing the current latency set would be greatly appreciated.

### Actual result

The keys are bound to other keys instead of the keys described on the wiki.

### To Do
- [ ] Update the wiki with new key combinations

nieknooijens:
there is a function dispInFlash() for that. shouldn't be too hard to add to the key handler functions.
I'm surprised this isn't called on any setting yet.

Baklap4:
Hmm latency is being flashed tho?: 
![image](https://user-images.githubusercontent.com/3918517/51184770-9184a200-18d5-11e9-8f9e-f9c6d8cfbda2.png)

jose1711:
ha, you are right! so this is in fact an out-of-date documentation (+ in game hints) issue

Baklap4:
Closed since #511 was merged to current master and the hint is back up to date :)! Thanks for sharing @jose1711

PR #511, opened 2020-03-26
Feature/fix shortcuts

Baklap4:
### What does this PR do?

Update of in-game hints related to latency tuning during playback to match the changes in code

### Closes Issue(s)

#454 

### Motivation

Hints should be kept up-to-date.

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -305,19 +305,19 @@ to save the current settings to XML.
 		<ui unit=" ms" multiplier="1000" />
 		<limits min="-0.5" max="0.5" step="0.01" />
 		<short>Audio/video latency</short>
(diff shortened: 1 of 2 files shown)


Baklap4:
> Why did we change these shortcuts

Not sure. Just copied over the pr from @jose1711 in trying to fix #454 

> Using [] as shortcuts is a disaster for some keyboard layouts 

Correct i suggest to pick this up in another issue. As of the current state i'd like to get this merged so it reflects the code as is.

> if they will work with command, Command+. in mac is the interrupt shortcut in Mac. I haven't tested it but these could cause trouble.

No mac here so would be good to get this tested yes.

Baklap4:
@Lord-Kamina added an issue (#522) to get more logical/better shortcut bindings. As for now this is our current code so these changes only reflect that code. I say we should get this merged as the current UI shows wrong shortcuts.

How the shortcut is working is a nonissue for this aswell imho.

Issue #33, opened 2013-05-07
sound problems when rewinding songs twice


katsuowota:
when rewinding a song twice really fast, the sound stops and the song continues, when playing a  bit around with the song again (forwarding, rewinding) the sound starts playing again.

Baklap4:
How's this with the latest master after 4 years?

Baklap4:
@Lord-Kamina you did some stuff with audio lately did you still encounter this bug?

Lord-Kamina:
I've encountered this occasionally, yes.

Tronic:
I would suggest changing ffmpeg.cc so that it seeks to correct location rather than to beginning of song if the requested time is (more than a fraction of a second) behind current timestamp or much ahead of current timestamp. Then test if audio still plays in sync with lyrics/notes/video after such rewinding (this was a problem a decade ago when I wrote the code). Feel free to hack (I have no time).

OznOg:
I'm working on this problem, I think I may eventually endup with something; That said, I think could the PR 512 could be merged separatly as the rework for this is quite different, IMO

Issue #444, opened 2018-12-31
An invalid (?) mp3 file can cause memory leak in preview


jose1711:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I've stumbled upon an mp3 file that causes performous (preview/selection mode) to start eating gigabytes of memory heavily bogging system. Ffplay gives plenty of warnings and skips through the song within a few seconds, mplayer and vlc play it without issues. The actual problem probably lies with ffmpeg but I think performous can handle this in better way (maybe skip with error logged to console).

My system equipped with 8 GB memory eventually recovered but it took no less than 10 minutes.

Baklap4:
Hmmm so that's what causing it.. i encountered this aswell i think last weekend i went to a party, and performous was hogging my memory badly at 60-70%.. way too much

Baklap4:
@jose1711 can you upload this mp3 file here on github so we can try this out ourselves?

jose1711:
@Baklap4 I can share it privately

Baklap4:
Alright hit me up at discord :)

Baklap4:
When playing the file i'm getting a `FfmpegError: code=-1094995529, error=Invalid data found when processing input` from ffmpeg..

Baklap4:
@jose1711 can you provide us again with the mp3? I've lost the one you sent me ages ago..

Lord-Kamina:
BTW I just realized... There is a chance I fixed it already with #507, since I did fix a leak in ffmpeg code.

Baklap4:
Well this certainly isn't fixed, just crashed my system with aubio fixes on master. I'm gonna check with #512 now

Baklap4:
Also not fixed by #512

Baklap4:
Playing this song within vlc player or such does work. @OznOg if i give you the file will you be able to track it down?

OznOg:
```
Thread 73 "performous" hit Breakpoint 6, AudioBuffer::push (this=0x7fff700b4280, data=std::vector of length 2474, capacity 2474 = {...}, timestamp=39867.219689000005)
    at git/performous/game/ffmpeg.cc:104
104                     m_pos = timestamp * m_sps;
(gdb) n
105                     m_data.resize(m_pos, 0);
(gdb) p m_pos
$13 = 3827253090
```
This is way too larg :)

Baklap4:
So it seems once we disable the if statement 
```
[...]

PR #512, opened 2020-03-26
RFC: Fix ffmpeg leaks

OznOg:
### What does this PR do?
Try to fix leaks in ffmpeg
please let me know it I missed some of them

### Closes Issue(s)
probably #444

### Motivation
Prevent resources leaks


OznOg:
Do we want to keeep support on old libav or not?

game/ffmpeg.cc
```diff
@@ -302,10 +315,11 @@ class FfmpegError: public std::runtime_error {
 
 void FFmpeg::decodePacket() {
 #if LIBAVCODEC_VERSION_INT >= (AV_VERSION_INT(57, 37, 0))
-	AVPacket pkt;
```
Lord-Kamina: Is this intended? Why'd you put the declaration inside the loop?

Lord-Kamina:
> Do we want to keeep support on old libav or not?



> Do we want to keeep support on old libav or not?

I don't really know how old that libav is, can you elaborate?

Lord-Kamina:
Actually, I checked... we've got compatibility code there going all the way back to FFmpeg 1.0; that's fucking insane. I propose we keep compatbility up to FFmpeg 3 (Major version number 57 for most components)

Tronic:
Is libav still used somewhere? I never followed how that played out after Debian began shipping libav in its ffmpeg packages and all that. If all major distributions are back to ffmpeg already, I don't see point in supporting both. Same goes for old versions because ffmpeg tends to be rapidly updated in distributions and its APIs are frequently deprecated.

game/ffmpeg.cc
OznOg: Yes this is intendended for now, I need more tests anyway. Don't bother too much with reviews for now, the purpose of the PR was mainly to know your opinion on the legacy AV lib we want to support. I'll let you know when the code is ready for reviews.

OznOg:
IMHO, the less #ifdef we have the less problems we have, nevertheless, what is the oldest distro begin supported? is ubuntu 16.04 considered outdated for us or not?

game/ffmpeg.cc
Baklap4: Setting this as 'resolved'

Baklap4:
I suppose ubuntu 16.04 is our oldest target for now.
Semaphore has been updated to also include 18.04 builds

on ubuntu 16.04 libavcodec-dev version 7:3.4.2-2build2ppa-xenial1 is used
which depends on libavutil-ffmpeg54

Lord-Kamina:
> Is libav still used somewhere? I never followed how that played out after Debian began shipping libav in its ffmpeg packages and all that. If all major distributions are back to ffmpeg already, I don't see point in supporting both. Same goes for old versions because ffmpeg tends to be rapidly updated in distributions and its APIs are frequently deprecated.

Sorry, I tend to use "libav" and "ffmpeg" interchangeably when I write.

OznOg:
> Is libav still used somewhere? I never followed how that played out after Debian began shipping libav in its ffmpeg packages and all that. If all major distributions are back to ffmpeg already, I don't see point in supporting both. Same goes for old versions because ffmpeg tends to be rapidly updated in distributions and its APIs are frequently deprecated.

yes, my bad, I meant ffmpeg, but flags actually are LIBAV_CODEC and so on...

Lord-Kamina:
> LGTM @Lord-Kamina can you also take a look?

I'm not sure I understand the implications of all the changes to be honest but looks ok.

OznOg:
agree that it is quite a lot of changes, but can be summed u like this:
* Remove legacy ffmpeg compat
* add unique_ptr to implement RAII
* remove dead code (in while loops)

I used https://www.ffmpeg.org/doxygen/3.4/filtering_video_8c-example.html#a90 as "reference implmentation"

game/ffmpeg.cc
```diff
@@ -170,85 +159,129 @@ FFmpeg::FFmpeg(fs::path const& _filename, unsigned int rate):
...
+				throw FfmpegError(ret);
+			}
+
+			if (pkt.stream_index == m_streamId) {
+
+				if (audioQueue.wantSeek()) m_seekTarget = 0.0;
+				if (m_seekTarget == m_seekTarget) seek_internal();
+				decodePacket(pkt);
+				errors = 0;
+			}
```
Tronic: Does this (still) support seeking back from near to the end of a song, once the decoder has already reached EOF?

Baklap4: @OznOg can you test/check/verify this :)?

OznOg:
Well as said in previous comment, I tested the use case I understood without notifying any difference with master, but I would need any power user to confirm...

game/ffmpeg.cc
Baklap4: Running a song almost till the end and then rewind still works!
There is still a bug from a few years ago https://github.com/performous/performous/issues/33
which sadly still applies.

Baklap4:
@OznOg can you check if you might be able to fix #33 aswell on this branch?

PR #521, opened 2020-04-01
Rework path list to keep select at top level

OznOg:
When I added the sort on menu entries, I did not pay attention to the
fact that the "add this directory" was now at the bottom of the page,
making it difficult to find when there are a lot of directories.

This patch reworks that, and remove the whole menu enties sort and replace
it by only path sort, so that the first entries that are pushed remains
at the top.


OznOg:
The bad one
good

Baklap4:
Aight looking good! gonna update it with latest master, than rebase it :)!
Thanks again for the PR!

PR #523, opened 2020-04-03
Added the ease of a uninstall target

Baklap4:
### What does this PR do?

Added a cmake uninstall target.
Aside from deleting all files, it also remove the directory and all empty directories within
```
$CMAKE_INSTALL_PREFIX/$SHARE_INSTALL
```
This folder should be the location where performous keeps all his data. On unix systems it's:
`/usr/local/share/games/performous/*`

### Closes Issue(s)

None

### Motivation

I always look for how to remove an opensource project completely from my pc. Sadly many projects which i install with cmake don't have a specific target for uninstalling. That's why i'd love performous to have one; For the ease of uninstalling.

Yes i'm aware you can also just remove all the files listed in `install_manifest.txt` using 
```
cat install_manifest.txt | sudo xargs rm
```
But still having a target makes it just plain simple to execute.
Also does the `cat` command **not** clean up the left over directories, **which this target does.**

### Notes

Followed this tutorial: https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,15 +128,17 @@ if(WIN32)
 	install(FILES win32/ConfigureSongDirectory.bat DESTINATION .)
 endif()
 
-# uninstall target
-if(NOT TARGET uninstall)
-  configure_file(
-    "${CMAKE_CURRENT_SOURCE_DIR}/cmake/uninstall.cmake.in"
-    "${CMAKE_CURRENT_BINARY_DIR}/cmake/uninstall.cmake"
-    IMMEDIATE @ONLY)
+if(NOT APPLE)
+  # uninstall target
+  if(NOT TARGET uninstall)
+    configure_file(
+      "${CMAKE_CURRENT_SOURCE_DIR}/cmake/uninstall.cmake.in"
+      "${CMAKE_CURRENT_BINARY_DIR}/cmake/uninstall.cmake"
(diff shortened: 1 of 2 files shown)


Baklap4:
Why wrap it into `if not apple`? I think it'd just work out fine in mac aswell? correct me if i'm wrong :o

Baklap4:
Updated it, not sure why we need this check though?

Lord-Kamina:
Mostly to prevent people doing dumb stuff. In mac, you just grab the .app and drag it to the trash. :P

Baklap4:
well people running the cli and doing dumb stuff... that's about to happen if those people running this target they should know what it does 🤔 
Eitherway it just removes the dirs performous uses if drag the icon to the thrash does the same than  sure let's get this merged

Lord-Kamina:
Dragging to the trash won't remove them but those folders are three or four folders deep inside the User folder, will weigh a few kb, frankly nobody cares and if they do, there are a million apps that clean those folders from thing leftover by apps. (Plus, many people will prefer not having to reconfigure)

I'd agree it's necessary if we left some kind of launch script or something nasty like that, but since it's just a couple pngs and text files. *shrugs*

And anyway, I don't think you can properly run Performous from CLI without the launcher anyway.

PR #524, opened 2020-04-04
Added changing of languages runtime within the menu

### LINKED ISSUE #246: Add option to set in game language
As the title says, add an option (In game menu and xml) to set the in game language. Right now we have to do the following to launch performous in another language:
```
export LANG=nl_NL.UTF-8
export LANGUAGE=nl_NL
```
This is not user friendly :joy:

Referenced from #236 

# TODO:
- [x] Add menu item
- [x] Show only 100% translated languages??
    - [x] It defaults to English when there's no translation found. If objection please mention
- [x] Save it to config.xml
- [x] Load the language at runtime
- [x] ~~Include #242~~

Baklap4:
### What does this PR do?

Adds changing the language while in the game menu.
It also starts the game in the language you're playing in.
The languages are also translated into the language you're viewing performous.

### Closes Issue(s)

Closes #246 

### Motivation

As explained in #246 we could only change the language by setting some environment variables. This was not quite handy. Fixed it so we can do it from within the menu

### Testing Notes

If you want to test this you should install the app. Otherwise you have to create a folder structure (within your build folder) like this:
```
build
|--- lang
|----- nl.mo
|----- nl
|------- LC_MESSAGES
|---------- Performous.mo
```


game/configuration.cc
```diff
@@ -293,6 +311,16 @@ void writeConfig(bool system) {
 				Game::getSingletonPtr()->window().resetShaders();
 			}
 		}
+		if(name == "game/language") {
+			std::string oldValue = Game::getSingletonPtr()->getCurrentLanguage();
+			std::string newValue = item.getEnumName();
```
OznOg, line 316: same

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+		std::clog << "locale/debug: Checking if language is set in config." << std::endl;
+		if(config["game/language"].i() >= 0) {
+			lang = config["game/language"].getEnumStringValueAtIndex(config["game/language"].i());
+			std::clog << "locale/debug: Language in configuration found: " << lang << std::endl;
+		}
+
+		if (lang.empty()) lang = "en_US.UTF-8";
+		setLanguage(lang);
+	};
+	static bool enabled() {
```
OznOg, line 26: seems useless

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+		return true;
+	};
+
+	void setLanguage(const std::string language) {
```
OznOg, line 30: const& or useless const

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
 			std::clog << "locale/warning: Unable to detect locale, will try to fallback to en_US.UTF-8" << std::endl;
 			std::locale::global(gen("en_US.UTF-8"));
 		}
-	};
-	static bool enabled() {
-		return true;
-	};
+	}
+
+	const std::string getLanguage(const std::string language) {
```
OznOg, line 51: no need for const on return value; const & on parameter

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+
+		if (foundLanguage == allLanguages.end())
+			returnLang = "en_US.UTF-8";
+		else 
+			returnLang = *foundLanguage;
+
+		return returnLang;
+	}
 
+	const std::string getCurrentLanguage() { return m_currentLanguage; }
```
OznOg, line 70: same

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+	std::vector<std::string> GetAllLanguages() { return config["game/language"].getAllEnumStringValues(); }
+private:
+	std::string m_currentLanguage;
+	const char *m_package;
```
OznOg, line 74: can it be null? who has ownership of that data? maybe considere using unique_ptr, or reference if not owned

game/screen.hh
```diff
@@ -103,6 +105,9 @@ class Game: public Singleton <Game> {
 	void drawLogo();
 	///global playlist access
 	PlayList& getCurrentPlayList() { return currentPlaylist; }
+	void setLanguage(const std::string language) { m_translationEngine.setLanguage(language); };
+    const std::string getCurrentLanguage() { return m_translationEngine.getCurrentLanguage(); };
```
OznOg, line 109: non const on return value

game/configuration.cc
```diff
@@ -165,6 +165,24 @@ namespace {
 	}
 }
 
+std::string ConfigItem::getEnumStringValueAtIndex(unsigned int index) {
+	try {
+		auto item = m_enums.at(index);
+		return item;
```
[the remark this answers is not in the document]
Baklap4: At throws a index out of range exception better catch it and throw runtime error

game/configuration.cc
```diff
@@ -165,6 +165,24 @@ namespace {
...
+	} catch (const std::out_of_range& e) {
+        std::clog << "configuration/error: Out of Range error." << e.what() << std::endl;
+		throw std::runtime_error("Enum Index not found in enumeration.");
```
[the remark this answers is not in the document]
Baklap4: Alright will change :)

game/configuration.cc
```diff
@@ -293,6 +311,16 @@ void writeConfig(bool system) {
 				Game::getSingletonPtr()->window().resetShaders();
 			}
 		}
+		if(name == "game/language") {
```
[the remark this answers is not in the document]
Baklap4: It was late gonna stick with `if (`

game/configuration.cc
```diff
@@ -293,6 +311,16 @@ void writeConfig(bool system) {
...
+			std::string oldValue = Game::getSingletonPtr()->getCurrentLanguage();
```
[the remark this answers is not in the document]
Baklap4: Will change :+1:

game/configuration.cc
Baklap4: Will change

game/i18n.hh
Baklap4: Not sure but will check it to make it sure :+1:

game/i18n.hh
Baklap4: Yup gonna get changed

game/i18n.hh
Baklap4: Alright thanks :)

game/i18n.hh
Baklap4: Yuush will change

game/i18n.hh
Baklap4: It is set from the main where it reads the projectname shouldnt be null

game/screen.hh
Baklap4: Alright will change

game/i18n.hh
Baklap4: Hmm for now i'll keep it in, can't really find a way to toggle between GETTEXT_FOUND or not which is what it should toggle

Baklap4:
@Lord-Kamina could you gives this a try on macos?
Note; you'll have to proper install performous before you're able to check this.. since languages only get properly installed on install and not on build.. Maybe we need to fix that..

Lord-Kamina:
I'll check it later, but keep in mind that in macOS, the app is always "installed", the installation folder is the app bundle. Please take a look at the launcher and the directory tree I'm attaching now so you can get an idea, then either you modify or tell how I need to modify it to use this.

[tree.txt](https://github.com/performous/performous/files/4447245/tree.txt)

Lord-Kamina:
Nevermind, it works with the macOS bundle using the current structure.

However, switching to a language and then back to English doesn't work. It saves the setting but doesn't update the UI until you restart performous.

Lord-Kamina:
That said, there's two changes I'd lie to see in this:

1) An "Automatic" or similar mode, which essentially keeps current behavior; in mac (and I believe perhaps in Windows also) it would automatically detect OS language and use that if available (User-set language from the UI should still win but I think reverting the English should be the last option)

And 2) This is purely cosmetic, but possibly you can use "virtOptName" I added to MenuOptions way back when I made the code to change audio backend, that way maybe we could use proper language names for displaying in the user UI?

Tronic:
Btw, force-pushes make following the development harder. I would suggest not rewriting history / rebasing while on a PR, but only when merging to master.

Baklap4:
> An "Automatic" or similar mode, which essentially keeps current behavior; in mac (and I believe perhaps in Windows also) it would automatically detect OS language and use that if available (User-set language from the UI should still win but I think reverting the English should be the last option)

If i'm correct it should get the system locale when nothing is within the `config.xml`
As for the English, that's a bit of a 'problem' since we actually don't have an English language. We have a Performous.pot file, which is basicly a po template file. Yes this is currently written in english but as far as the application is concerned it could be russian aswell. It's just a mapping between keys within the code and the languages files we have (which don't list english).

So i think we have to add a definition for the english language to get this fixed.
Agreed?

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+			std::clog << "locale/notice: No given language.. using default system language: '" << lan << "''" << std::endl;
+			return lan;
+		}
+
+		auto allLanguages = GetAllLanguages();
+		auto foundLanguage = std::find(allLanguages.begin(), allLanguages.end(), language);
+		auto returnLang = std::string();
+
+		if (foundLanguage == allLanguages.end())
+			returnLang = "en_US.UTF-8";
```
Lord-Kamina, line 63: I meant here

game/i18n.hh
```diff
@@ -2,26 +2,74 @@
...
+	TranslationEngine(std::string package) : m_package(package) {
+		auto lang = boost::locale::util::get_system_locale(true);
+
+		std::clog << "locale/debug: Checking if language is set in config." << std::endl;
+		if(config["game/language"].i() >= 0) {
+			lang = config["game/language"].getEnumStringValueAtIndex(config["game/language"].i());
+			std::clog << "locale/debug: Language in configuration found: " << lang << std::endl;
+		}
+
+		if (lang.empty()) lang = "en_US.UTF-8";
```
Lord-Kamina, line 23: And here

game/i18n.hh
Baklap4: I've added those since the system language doesn't have to be one of the languages we have translations for.

Baklap4:
Correct as a first time there will be a `-1` i have no idea how to get rid of this.. :/

Regarding the userfriendly part of the names; i agree but i'm not quite sure how to do this with the current configuration system in place

Lord-Kamina:
> Correct as a first time there will be a `-1` i have no idea how to get rid of this.. :/
> 
> Regarding the userfriendly part of the names; i agree but i'm not quite sure how to do this with the current configuration system in place

You could maybe use faux, dynamically generated values? Take a look at what I did in `populateBackends()`

Baklap4:
> You could maybe use faux, dynamically generated values

@Lord-Kamina i've fixed that.

Also the it now auto selects the language as it used to do.

If your system uses anything we don't support it falls back to English.

It also enables translating the languages list into the translated language name.

Some things can maybe be cleaned up a bit but it's 3 am atm here so i just had to push it.

game/configuration.cc
```diff
@@ -439,3 +494,17 @@ void populateBackends (const std::list<std::string>& backendList) {
...
+
+void populateLanguages(const std::map<std::string, std::string>& languages, bool refreshOutDated) {
+	ConfigItem& languageConfig = config["game/language"];
+	if (refreshOutDated) {
+		languageConfig.removeAllEnums();
+	}
+	for (auto const& language : languages) {
+		languageConfig.addEnum(language.second);
+	}
+	static std::string selectedLanguage = std::string();
```
OznOg, line 506: why is that static? BTW the initialization is useless.

game/fs.cc
```diff
@@ -12,6 +12,7 @@
 #include <mutex>
 #include <set>
 #include <sstream>
+#include <regex>
```
OznOg, line 15: is that realy needed in here?

game/game.hh
```diff
@@ -75,6 +76,8 @@ class Game: public Singleton <Game> {
 	void drawLogo();
 	///global playlist access
 	PlayList& getCurrentPlayList() { return currentPlaylist; }
+	void setLanguage(const std::string language) { m_translationEngine.setLanguage(language, true); };
```
OznOg, line 79:
```suggestion
	void setLanguage(const std::string &language) { m_translationEngine.setLanguage(language, true); };
```

game/game.hh
```diff
@@ -75,6 +76,8 @@ class Game: public Singleton <Game> {
...
+	const std::string getCurrentLanguage() { return m_translationEngine.getCurrentLanguage().second; };
```
OznOg, line 80:
```suggestion
	std::string getCurrentLanguage() { return m_translationEngine.getCurrentLanguage().second; };
```

game/game.hh
```diff
@@ -107,6 +110,7 @@ private:
 	AnimValue m_dialogTimeOut;
 	// Dialog members
 	std::unique_ptr<Dialog> m_dialog;
+	TranslationEngine& m_translationEngine;
```
OznOg, line 113: shoudl/could it be const?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
-			std::locale::global(gen("en_US.UTF-8"));
-		}
+	TranslationEngine(const char *package) : m_package(package) {
+
+		initializeAllLanguages();
+		populateLanguages(GetAllLanguages());
+
+		std::clog << "locale/debug: Checking if language is set in config." << std::endl;
+
+		auto prefValue = config["game/language"].getEnumName();
```
OznOg, line 23: why not storing english tong name instead of enums?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+		auto lang = getLanguageByHumanReadableName(prefValue);
+
+		setLanguage(lang);
 	};
 	static bool enabled() {
 		return true;
 	};
 
+	static ConfigItem& translationConfig() {
```
OznOg, line 32: you are creating a singleton; is that really *needed* why not just returning the configIten by value?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+		static ConfigItem& backend = config["game/language"];
+		return backend;
+	}
+
+	void initializeAllLanguages() {
+		auto path = getLocaleDir().string();
```
OznOg, line 38: shouldn't you keep path as is instead of converting?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+		else {
+			m_currentLanguage = { language , getLanguageByKey(language) };
+		}
+
+		try {
+			std::locale::global(m_gen(m_currentLanguage.first));
+			std::cout << "locale/notice: Current language is: '" << m_currentLanguage.second << "'" << std::endl;
+			populateLanguages(GetAllLanguages(true), true);
+		}
+		catch (...) {
```
OznOg, line 64: is that really throwing *anything*?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+	}
+
+	std::string getLanguageByHumanReadableName(const std::string& language) {
+		if (language == _("Auto")) {
+			return boost::locale::util::get_system_locale(true);
+		}
+
+		auto allLanguages = GetAllLanguages();	
+		auto languageKey = std::string();
+		for (auto it = allLanguages.begin(); it != allLanguages.end(); ++it) {
```
OznOg, line 77: use range for loop?

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+		}
+
+		auto allLanguages = GetAllLanguages();
+		if (allLanguages.find(languageKey) == allLanguages.end()) {
+			return allLanguages["en_US.UTF-8"];
+		}
+		return allLanguages.at(languageKey);
+	}
+
+	const std::pair<std::string, std::string> getCurrentLanguage() { return m_currentLanguage; }
```
OznOg, line 99: if you really want by value
```suggestion
	 std::pair<std::string, std::string> getCurrentLanguage() { return m_currentLanguage; }
```
or by reference:
```suggestion
	const std::pair<std::string, std::string> &getCurrentLanguage() { return m_currentLanguage; }
```

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+
+	std::map<std::string, std::string> GetAllLanguages(bool refresh = false) {
+		if (refresh) {
+			m_languages.clear();
+		}
+		if (m_languages.size() != 0) return m_languages;
+		m_languages.emplace(std::pair("None", _("Auto")));
```
OznOg, line 106: I think 
```suggestion
		m_languages = { {"None", _("Auto") }, ...}; 
```
should work

game/i18n.hh
```diff
@@ -2,26 +2,130 @@
...
+		m_languages.emplace(std::pair("pl_PL.UTF-8", _("Polish")));
+		m_languages.emplace(std::pair("pt_PT.UTF-8", _("Portuguese")));
+		m_languages.emplace(std::pair("sv_SE.UTF-8", _("Swedish")));
+		m_languages.emplace(std::pair("zh_CN.UTF-8", _("Chinese")));
+		return m_languages;
+	}
+
+private:
+	std::pair<std::string, std::string> m_currentLanguage;
+	std::string m_package;
```
OznOg, line 128: not really clear what all this stands for...

game/fs.cc
Baklap4: Yup. Within `fs_paths.win.inc` we use a regex and we can't add the include in that file

game/game.hh
Baklap4: When making it const i'm getting:
```
Error (active)	E1086	the object has type qualifiers that are not compatible with the member function "TranslationEngine::setLanguage" C:\Users\Arjan\source\repos\performous\performous2\game\game.hh#79	
```

game/i18n.hh
Baklap4: Boost locale `generator.add_messages_path()` only supports std::string

game/i18n.hh
Baklap4: image
I guess we can catch runtime exception agreed?

game/i18n.hh
Baklap4: Using that notation now 👍

game/i18n.hh
Baklap4: it's the performous name. It's used for reference performous.mo files for every language and it comes from the main method

game/configuration.cc
Baklap4: Removed the initialization and static

game/fs.cc
Baklap4: Yup, fs_paths.win.inc has a regex match and the include cannot be in there

game/game.hh
Baklap4: Changed to reference

game/game.hh
Baklap4: Removed the const

game/game.hh
Baklap4: Cannot be const due to `std::string getCurrentLanguage() { return m_translationEngine.getCurrentLanguage().second; }`

game/i18n.hh
Baklap4: Hmmm @Lord-Kamina suggested it this way like he did with Audio backends

game/i18n.hh
Baklap4: This method gets called from the config WriteConfig method. We don't have any reference to TranslationEngine there. Currently looking into it if it's necessairy

game/i18n.hh
Baklap4: Catching runtime exception now

game/i18n.hh
Baklap4: Fixed

game/i18n.hh
Baklap4: Did it by reference

game/i18n.hh
Baklap4: Yup works, nice cleanup

game/i18n.hh
Baklap4: It's the name of the application "performous" it's used as the *.mo files are all called Performous.mo once installed in different folders. It's given from the main method

game/i18n.hh
Baklap4: Removed it as it wasn't necessairy

game/game.hh
OznOg: Returns a copy THUS can be const

game/fs.cc
OznOg: ok

game/game.hh
```diff
@@ -75,6 +76,8 @@ class Game: public Singleton <Game> {
 	void drawLogo();
 	///global playlist access
 	PlayList& getCurrentPlayList() { return currentPlaylist; }
+	void setLanguage(const std::string& language) { m_translationEngine.setLanguage(language, true); };
+	std::string getCurrentLanguage() { return m_translationEngine.getCurrentLanguage().second; };
```
OznOg, line 80:
```suggestion
	std::string getCurrentLanguage() const { return m_translationEngine.getCurrentLanguage().second; };
```

game/i18n.hh
```diff
@@ -2,26 +2,126 @@
...
+		}
+
+		auto allLanguages = GetAllLanguages();
+		if (allLanguages.find(languageKey) == allLanguages.end()) {
+			return allLanguages["en_US.UTF-8"];
+		}
+		return allLanguages.at(languageKey);
+	}
+
+	const std::pair<std::string, std::string>& getCurrentLanguage() { return m_currentLanguage; }
```
OznOg, line 91:
```suggestion
	const std::pair<std::string, std::string>& getCurrentLanguage() const { return m_currentLanguage; }
```

ooshlablu:
I can confirm that this works. I can change the language and anything that has correct translation strings is converted appropriately after the settings are saved.

game/configuration.cc
```diff
@@ -130,6 +132,13 @@ std::string const ConfigItem::getValue() const {
 		else std::clog << "audio/warning: Currently selected audio backend is unavailable on this system, will default to Auto." << std::endl;
 		return "Auto";
 	}
+	if (this->getName() == "game/language") {
+		int autoLanguageType = 1337;
+		static int val = std::get<int>(m_value);
```
OznOg, line 137: why static? seem awkward here....

game/configuration.cc
```diff
@@ -415,6 +449,27 @@ int PaHostApiNameToHostApiTypeId (const std::string& name) {
 	throw std::runtime_error("Invalid PortAudio HostApiTypeId Specified.");
 }
 
+unsigned int LanguageToLanguageId(const std::string& name) {
```
OznOg, line 452: now we are in c++17, time to use std::optional<>

game/configuration.cc
```diff
@@ -415,6 +449,27 @@ int PaHostApiNameToHostApiTypeId (const std::string& name) {
...
+	if (name == _("Asturian")) return 1;
+	if (name == _("Danish")) return 2;
+	if (name == _("German")) return 3;
+	if (name == _("English")) return 4;
+	if (name == _("Spanish")) return 5;
```
OznOg, line 457: what are all those magic values refering to (1, 2, 3, ...)? is that some order somewhere?

game/configuration.hh
```diff
@@ -43,6 +43,7 @@ class ConfigItem {
 	std::string const& getShortDesc() const { return m_shortDesc; } ///< get the short description for this ConfigItem
 	std::string const& getLongDesc() const { return m_longDesc; } ///< get the long description for this ConfigItem
 	void addEnum(std::string name); ///< Dynamically adds an enum to all values
+	void removeAllEnums(); /// Removes all the enum values from this entry.
```
OznOg, line 46: I'm affraid the naming "enume values" is quite missleading, there is no enum in there, isn't there?

Baklap4: So to get back to this. Within our `schema.xml` we define "enums" these "enums" are basicly a label with an integer backing it up. All of the enums defined in there aren't actual Enum class values throughout our C++ codebase.

The name enum for this is poorly choosen or the usage is poorly applied. As for now i think we'll have to live with this. But i can feel a configuration.cc/hh overhaul coming.

game/configuration.cc
Baklap4: As stated below the "enum" parent needs an integer value to work correctly. The numbers here are arbitrary and doesn't refer to anything except the selected language in the parent scope of the schema.xml

game/configuration.cc
Baklap4: Not quite sure what you want me to do here. All the cases are handled in this method

game/configuration.cc
Baklap4: Removed the static and the check on the next line as it didn't provide any value.

game/configuration.cc
OznOg: don't mind, I think this can be done in a future rework

game/configuration.cc
OznOg: ok

game/configuration.hh
OznOg: ok

game/configuration.cc
OznOg: ok

OznOg (approved):
LGTM

Issue #395, opened 2018-09-07
Modernize CMake


Lord-Kamina:
Save fixing a few things here and there, I believe we have not touched our CMakeLists in ages. It would probably be in everybody's interest if at some point we were to write them from the ground up using modern CMake design patterns, as described in 

https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1

https://www.youtube.com/watch?v=eC9-iRN2b04

and https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/

Baklap4:
Linking #767 for reference

OznOg:
I also recommend https://www.youtube.com/watch?v=bsXLMQ6WgIk and https://www.youtube.com/watch?v=m0DwB4OvDXk

PR #529, opened 2020-04-07
Issue/refactor cmake

Lord-Kamina:
### What does this PR do?

Split all this out from the cache PR; chiefly it refactors and streamlines the CMakeLists, as well as giving the script for building in mac a much-needed slimming-down.

### Motivation

Originally, because I wanted to do away with the dirtynes of specifying the C++ standard in the mac command-line, that broke compiling and down the rabbit hole I went.

### More


game/platform.hh
```diff
@@ -25,12 +21,10 @@
 #endif
 
 struct Platform {
-enum platforms { windows, linux, macos, bsd, solaris, unix };
+enum class Host_OS { Performous_OS_Win32, Performous_OS_Linux, Performous_OS_macOS, Performous_OS_BSD, Performous_OS_Solaris, Performous_OS_Unix };
```
Tronic, line 24: I would suggest restoring the original OS names, possibly capitalized, but get rid of all the extra prefixes and underscores. This whole thing is already namespaced within `Platform` so it is unclear if even an `enum class` brings any extra value here. `Platform::Host_OS::Performous_OS_macOS` WTF?!

Baklap4: I'd suggest making it enum class OS, and indeed capitalized OS names.

I'd make it an enum class to prevent warnings to happen and possible mistakes to be made. See https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum for an explanation

Lord-Kamina: @Tronic there were some macro clashes because "linux" was being expanded to 1 in other places.

OznOg, line 24: You could remove the Platform struct (was just here to create a namespace) and do it
enum class Platform { OS_Win32, OS_Linux, OS_macOS, OS_BSD, OS_Solaris, OS_Unix };

Lord-Kamina: I think OS_Linux and OS_Win32 might also be defined by some compilers or in some circumstances, which is why I added Performous to the name also.

>

Lord-Kamina: Also, there are a few other things inside Platform. I don't think finding elsewhere to put them really warrants slightly shorter member names.

Baklap4: Maybe Platform::OS::WIN32, PLATFORM:OS::LINUX etc?

game/CMakeLists.txt
```diff
@@ -60,17 +60,52 @@ endif()
...
+	elseif (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 3.4 )
+		message(STATUS "Setting C++ standard to: C${Esc}[36m++14${Esc}[m")
+		set_property(TARGET performous PROPERTY CXX_STANDARD 14)
+	else ()
+	# FIXME: This case is probably not correct but seriously, I don't think it will come up very frequently anyway.
+		set_property(TARGET performous PROPERTY CXX_STANDARD 11)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-attributes -Wno-deprecated-declarations")
+		message(WARNING "Your GCC is rather old, disabling some useful compiler warnings (GCC 5+ or Clang 3.4+ recommended for proper C++14 support).")
+	endif()
+endif()
```
Baklap4, line 101: Can we also add a msvc section?

Lord-Kamina: Can you make it?

Baklap4: Hmmm currently not on windows @allcreator any experience with this?

Baklap4: Hmmm currently not on windows @allcreater any experience with this?

game/platform.hh
Tronic: All caps is quite likely to collide with macros too. Win32, Linux, MacOS etc. might be safest if no prefix is used because traditional C coders don't like naming things like that.

game/CMakeLists.txt
Lord-Kamina: > Hmmm currently not on windows @allcreater any experience with this?

https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019

This should let you do it?

Baklap4:
@Lord-Kamina could you also take https://github.com/performous/performous/issues/395 in account?

Lord-Kamina:
> @Lord-Kamina could you also take #395 in account?

Plenty of work to be done then still but I did implement some of those things here.

game/CMakeLists.txt
allcreater: > 
> 
> Hmmm currently not on windows @allcreater any experience with this?

I could think about it :)
But in theory I think it's better to minimize platrofm-specific declarations instead of multiplying them :) As far as I remember there is cross-platforms hints for language standard etc.

Lord-Kamina: Well, I guess the ideal thing would be to declare which language features we actually are using and let CMake figure the standard out.

Lord-Kamina: This is what we need to check but frankly I'm not entirely sure about all the ones we're using:

https://cmake.org/cmake/help/v3.1/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES

Lord-Kamina: And here is a quickly done, preliminary list: Feel free to add/remove stuff,

```
cxx_attributes
cxx_auto_type
cxx_constexpr
cxx_defaulted_functions
cxx_deleted_functions
cxx_enum_forward_declarations
cxx_explicit_conversions
cxx_extended_friend_declarations
cxx_lambdas
cxx_lambda_init_captures
cxx_noexcept
cxx_nonstatic_member_init
cxx_nullptr
cxx_range_for
cxx_right_angle_brackets
```

Lord-Kamina: Can somebody else weigh in re: this? @performous/core-owners

Baklap4:
Labeled: "wont fix" -> Read: Wont merge.
We still can use some stuff out of this pr and propose a new pr.

Lord-Kamina:
> Labeled: "wont fix" -> Read: Wont merge.
> We still can use some stuff out of this pr and propose a new pr.

Ok, so...

Several of the changes have already been merged in one form or another; to summarize...

- 9d96cef5c7ed0c196ef6e495b39279a184716974 was improved and merged in #612
- eeb8291e09234523fb8310f0ef44fa7904f8bd70 was cherry-picked in #614 
- e07e4fcf0a0cf77ae27b0020bb8449e9600d397e is still under discussion in #625 
- 6a746c6b70b0d2126f8f7e9dad759930c2ed6af7 was partially duplicated and merged in #596 
- 16452c9296f6f77040ebec14061273750fa7d773, 7ffa818197163f820ddb9457dc77c7e6cfe9ebad, a2be36fa5a23a446db596059ed8d9e25756b87fc and 48e504dca7ec08b2e945a8ffccca357b63da18b7 are part of #626 
- e9a65320d6fb5c3db20920a814acc058d052d803's idea was taken in with #499. I'm not sure, though, whether we are using any other c++17 exclusive features? if not, maybe we should allow c++14 when using Boost.Filesystem?
- a2b2bf1bf80886abfdd9a2a4168ca6a58d170362 is now outdated, but we should take the idea of it into our current CMake code, IMO.

Baklap4:
@Lord-Kamina I think all of the issues have been addressed in separate pr's which have already been merged to master. If not please reopen it again.

PR #532, opened 2020-04-09
Added project build artifacts

Baklap4:
### What does this PR do?

Creates a `.deb`, `.rpm` and `.tar.gz` for linux builds.
Creates a `.tar.gz` build for osx.
Creates a `nsis installer` for windows builds. - Can't be added to CI yet as we don't have a crosscompile or native platform in the CI yet. It's been extensively tested in combi with #650 

DMG and ~~Nsis installer~~ will be done in a separate PR as that's a bit more work.

### Closes Issue(s)

Partly #389 

### Motivation

Generating releases makes it more usable for the public.
Because the people who don't know how to build / don't want to build are stuck with no releases yet since 2016...

### Additional Notes

If anyone can tell me how to make a deb, rpm (ubuntu) (@OznOg) and dmg file (osx) that'd be even better. (@Lord-Kamina )


Lord-Kamina:
The macOS script outputs a dmg if configured correctly but I'm not sure the one the CI makes will be  usable by everyone.

Baklap4:
Hmmm what should be the flags or feel free to add them yourself on top of this commit :+1: 

> Build not be usable by everyone

True but still better than no builds at all.
We can improve the CI to create 'better' release builds upon tags.

Lord-Kamina:
> Hmmm what should be the flags or feel free to add them yourself on top of this commit 👍
> 
> > Build not be usable by everyone
> 
> True but still better than no builds at all.
> We can improve the CI to create 'better' release builds upon tags.

I mean because in my mac I especially add flags like `headerpad_max_install_names`, so there's a chance the CI could fail when making the complete bundle.

We'd need to install macdylibbundler, npm and an npm module I can't remember that makes the bundle (you cam get it if you take a look at the  script)

Once all requirements are met, the bundle should be created without the -D (debug) switch. That will increase build time quite a bit though probably so I wouldn't add it to regular CI builds.

Baklap4:
Answered my questions.. i can use cpack for rpm and debian packages

CMakeLists.txt
```diff
@@ -169,8 +169,13 @@ SET(CPACK_DEBIAN_PACKAGE_SECTION "Games")
 
 # RPM specific settings
 SET(CPACK_RPM_PACKAGE_GROUP "Amusements/Games")
-SET(CPACK_RPM_PACKAGE_AUTOREQ ON)
+
+SET(CPACK_RPM_PACKAGE_REQUIRES "cairo-devel, librsvg2-devel, glibmm24-devel, libxml++-devel, boost-devel, SDL2-devel, libepoxy-devel, ffmpeg-devel, portaudio-devel, opencv-devel, portmidi-devel, libjpeg-turbo-devel, pango-devel, libpng-devel, glm-devel, openblas-devel, fftw-devel, cpprest-devel")
 SET(CPACK_RPM_PACKAGE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/licence.txt")
+message("${LC_MESSAGES_DIRS}")
+message("HOOOOOI")
```
Tronic, line 176: Dutch for "hello?"

Baklap4: Yup had to sync between laptops this is gonna get squashed/deleted :joy:

ooshlablu:
I have built and tested a package on ubuntu 20.04. Everything works as expected, the game functions normally.

Baklap4:
Aight in the end we have finally some artifacts which can be downloaded from semaphore.
**Linux**:
* Tar.gz
* deb
* rpm

**osx**:
* tar.gz

**Windows**:
* Nada

I played a bit around with nsis but couldn't get it to work yet. So i removed it from the PR. Same goes for DMG for osx.

# Note
The builds are made on ubuntu 1804. This means that if you download the `deb` it'll only work for ubuntu 1804 as cpack has the dependencies at specific version baked into it.

# Build times
image

# Artifacts
image

**Artifacts will only be available for 1 day after building.**

ooshlablu:
Built packages and ran on ubuntu18.04 and 20.04.

:+1:

deuteragenie:
I cannot see any project artifacts on semaphore.

Baklap4:
> I cannot see any project artifacts on semaphore.

Correct you'll need to log in for now.
I've asked semaphore if it's possible to allow anonymous access to these artifacts here: https://github.com/semaphoreci/semaphore/issues/69

ooshlablu:
I've tested the debs for both clang and gcc on ubuntu 18.04 and they run without any issues.

I could not get the MacOS build to run. I extracted it, put it into a folder named `Performous.app`, which shows the icon correctly, but double-clicking to launch does nothing. I ran it from the command line, and got the following error:
```
dyld: Library not loaded: /opt/local/lib/libicuuc.65.dylib
  Referenced from: /Applications/Performous_semaphor.app/Contents/MacOS/performous
  Reason: image not found
 ```
I'm pretty sure this is to be expected, given there's a ton of active work around the MacOS builds right now :smiley:

**Comment by Baklap4** [comment]
@ooshlablu it's an error in setup regarding build machine vs actual machine. If you local don't have for instance icuuc65 installed it won't run :)

**Comment by Baklap4** [comment]
Staticly linked builds will be picked up in #629

**Review by Lord-Kamina — DISMISSED** [review]
Please use a separate pipeline file to define nightly or whatever builds.
Check here for an idea: https://docs.semaphoreci.com/essentials/schedule-a-workflow-run/

This way, not every build of every PR will generate artifacts. Also, we could for example configure these macOS builds to create proper portable bundles (It would require some minor work via SSH to install a few extra apps and libraries in the CI image we use for the cache)

**Comment by deuteragenie** [comment]
Also, project artifacts should only be generated when master/head is updated.
A daily (or maybe weekly) build would need to start comparing the last mast/head with the current one, and stop if it did not change.  This is because there can be many days without activity on the master branch, and it would be wasteful and potentially confusing to keep many artifacts around.
If possible, keep artifacts for 30 days or more, to allow for bisecting bugs easily if necessary.
Not sure if artifacts should have the date of generation in their name - this would depend on how they are presented in semaphore (the most recent should be available first).
The web pages should be updated to point to the "daily" builds so that people interested can find where to download them from easily.

**Comment by Baklap4** [comment]
> Please use a separate pipeline file to define nightly or whatever builds.

@Lord-Kamina That's already configured within semaphore itself and uses + runs the current semaphore.yml

> This way, not every build of every PR will generate artifacts.

@Lord-Kamina / @deuteragenie Why wouldn't we want artifacts for every PR? I think that's a great bonus, as this allows us to just download the deb or rpm and install it (if the machine is the same configured as build machine or even better when we have a static linked build (#629)). If space is an issue once this artifacts thing is out of beta we'll deal with it then. For now artifacts are kept for just one day at max.

> we could for example configure these macOS builds to create proper portable bundles

@Lord-Kamina Will be done in a different PR just like windows build / installers. As this has been open for way too long already + I can't quite test macos builds or windows builds as of now correctly

> Also, project artifacts should only be generated when master/head is updated.

@deuteragenie  i disagree. See point mentioned above about testing that 'release' and specific pr. It's not in our way to not generate it and it helps non-programmers to test pr's too.

> This is because there can be many days without activity on the master branch, and it would be wasteful and potentially confusing to keep many artifacts around.

@deuteragenie I disagree a nightly build as the name states is a nightly build. Even if there's no activity on master. I wouldn't call it wasteful as the underlying OS _can_ change from day to day. And if something breaks we want to see this ASAP. Regarding many artifacts this is just not true as the artifacts generated are kept just for 1 day max.

>  date of generation in their name

@deuteragenie They have their respective commit sha in their name. As stated above an artifact is at max 1 day old so it's true to the 'nightly build'.

> The web pages should be updated to point to the "daily" builds so that people interested can find where to download them from easily.

@deuteragenie like i stated above i've currently listed an issue over at semaphore to allow anonymous access to artifacts. Once that is achieved on their side we can do somtehing like this. Currently the artifacts are only available for people who are logged into their semaphore account. I wouldn't call this a blocking feature. If semaphore doesn't wanna support this we should look into something else. But for the time being this can be used.


Can't see why this can't be merged as of the current state. Please re-evaluate your review @Lord-Kamina

**Comment by Lord-Kamina** [comment]
> > Please use a separate pipeline file to define nightly or whatever builds.
> 
> @Lord-Kamina That's already configured within semaphore itself (by using the web ui -> https://docs.semaphoreci.com/essentials/schedule-a-workflow-run/#web-ui ) and uses + runs the current semaphore.yml
> 
> > This way, not every build of every PR will generate artifacts.
> 

I see that re: the nightlies already being scheduled.

> Can't see why this can't be merged as of the current state. Please re-evaluate your review @Lord-Kamina

Still, I stand by my comment that we should have a separate YAML file specifically for such builds and artifact generation, rather than making the main workflow more complex.

**Comment by Baklap4** [comment]
> Still, I stand by my comment that we should have a separate YAML file specifically for such builds and artifact generation, rather than making the main workflow more complex.

I can create a workflow for the nightly build so that it is managed from the repo, but i'd also still like the artifacts on each PR. I can't use the `epilogue` section for this as this hits after each job and will thus fail for either gcc or clang depending on who's earlier finished. I could take a look into checking the `CC` env variable and depending on the result create the artifacts for the specific build. Would this be alright?

**Comment by Lord-Kamina** [comment]
Sounds ok.

**Comment by Baklap4** [comment]
@Lord-Kamina Added the epilogue (it works!) and added a project definition. For now the nightly builds are the same as our normal builds. We can change that later for example if we wanted 'release' type builds.

**Comment by ooshlablu** [comment]
## Docker Build
I've built packages using my docker build images:
- `ooshlablu/performous-build:ubuntu18.04`
- `ooshlablu/performous-build:ubuntu20.04`
- `ooshlablu/performous-build:fedora32`
- `ooshlablu/performous-build:fedora33`
- `ooshlablu/performous-build:fedora34`

using the provided build script within the containers: `./build_performous.sh -c -g -p 532`

## Fedora
I do not have a running Fedora Desktop machine, but I copied the packages built on my images to fresh Official Fedora Docker Images: `fedora:32`, `fedora:33`, `fedora34`
I had to enable RPM Fusion to install dependencies:  
```
dnf install -y https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
```  
Once RPM Fusion was installed, the rpm install worked, and I was able to run Performous enough to get a version at least:  
```
[root@56be0ab82278 /]# performous -v
Performous 1.1-1045-gf1db3bbc
```
This worked for all versions of Fedora.

_Running a straight `performous` obviously resulted in some fun errors about not being able to find various systems such as graphics, sound, and input, but at least it was looking for everything it was supposed to :-)_

## Ubuntu
I copied the packages produced in the containers to real computers running their respective Ubuntu Desktop version. Both Ubuntu 18.04 & 20.04 run fine with singing, guitar, bass and any combination of the three.

File: .semaphore/semaphore.yml [hunk 2]
```diff
@@ -35,13 +37,26 @@ blocks:
...
           env_vars:
+            - name: COMPILER
+              value: clang
             - name: CC
               value: clang-9
             - name: CXX
               value: clang++-9
           commands:
             - mkdir build.clang && cd build.clang && cmake -DENABLE_WEBSERVER=ON -DCMAKE_VERBOSE_MAKEFILE=1 -DENABLE_WEBCAM=ON ..
             - make VERBOSE=1
```
ooshlablu, line 48: Any chance that a `make -j$(nproc --all) VERBOSE=1` would work here? Or would Semaphore malfunction?

ooshlablu (approved):
Can't say much for the game code itself, but it works in my testing.

.semaphore/semaphore.yml
Baklap4: i guess that should work. Let's try it in a separate pr

PR #535, opened 2020-04-12
Seek backward

### LINKED ISSUE #33: sound problems when rewinding songs twice

### LINKED ISSUE #444: An invalid (?) mp3 file can cause memory leak in preview

OznOg:
Be able to seek backward (or anywhere actually) when playing

# Closes issue

* Closes #444 
* Closes #33

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -39,8 +39,8 @@ namespace {
 }
 
 AudioBuffer::uFvec AudioBuffer::makePreviewBuffer() {
-        uFvec fvec(new_fvec(m_data.size() / 2));
-        float previewVol = float(config["audio/preview_volume"].i()) / 100.0;
+	uFvec fvec(new_fvec(m_data.size() / 2));
+	float previewVol = float(config["audio/preview_volume"].i()) / 100.0;
 	{
 		std::lock_guard<std::mutex> l(m_mutex);
 		for (size_t rpos = 0, bpos = 0; rpos < m_data.size(); rpos += 2, bpos ++) {
@@ -51,7 +51,7 @@ AudioBuffer::uFvec AudioBuffer::makePreviewBuffer() {
 };
 
 bool AudioBuffer::wantMore() {
-	return m_write_pos < m_read_pos + m_data.size() / 2;
+	return m_write_pos < m_read_pos + static_cast<std::int64_t>(m_data.size() / 2);
 }
 
 /// Should the input stop waiting?
@@ -96,23 +96,21 @@ bool AudioBuffer::prepare(std::int64_t pos) {
 	m_read_pos = pos;
 	m_cond.notify_all();
 	// Has enough been prebuffered already and is the requested position still within buffer
-	return m_write_pos > m_read_pos + m_data.size() / 16 && m_write_pos <= m_read_pos + m_data.size();
+	auto ring_size = static_cast<std::int64_t>(m_data.size());
+	return m_write_pos > m_read_pos + ring_size / 16 && m_write_pos <= m_read_pos + ring_size;
 }
 
 // pos may be negative because upper layer may request 'extra time' before
 // starting the play back. In this case, the buffer is filled of zero.
 //
 bool AudioBuffer::read(float* begin, size_t samples, std::int64_t pos, float volume) {
-	if (eof(pos))
-		return false;
-
 	if (pos < 0) {
 		size_t negative_samples;
 		if (static_cast<std::int64_t>(samples) + pos > 0) negative_samples = samples - (samples + pos);
 		else negative_samples = samples;
 
 		// put zeros to negative positions
-		std::fill(begin, begin + negative_samples, 0); 
+		std::fill(begin, begin + negative_samples, 0);
 
 		if (negative_samples == samples) return true;
 
@@ -122,17 +120,19 @@ bool AudioBuffer::read(float* begin, size_t samples, std::int64_t pos, float vol
 	}
 
 	std::unique_lock<std::mutex> l(m_mutex);
-	if (static_cast<std::uint64_t>(pos) >= m_read_pos + m_data.size()
-			|| static_cast<std::uint64_t>(pos) < m_read_pos) {
+	if (eof(pos) || m_quit)
+		return false;
+
+	if (pos >= m_read_pos + static_cast<std::int64_t>(m_data.size()) || pos < m_read_pos) {
 		// in case request position is not in the current possible range, we trigger a seek
 		// Note: m_write_pos is not checked on purpose: if pos is after
 		// m_write_pos, zeros present in buffer will be returned
-		std::fill(begin, begin + samples, 0); 
+		std::fill(begin, begin + samples, 0);
 		m_read_pos = pos + samples;
 		m_seek_asked = true;
-		std::fill(m_data.begin(), m_data.end(), 0); 
+		std::fill(m_data.begin(), m_data.end(), 0);
 		m_cond.notify_all();
-		return true; 
+		return true;
 	}
 
 	for (size_t s = 0; s < samples; ++s) {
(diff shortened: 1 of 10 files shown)


game/ffmpeg.cc
```diff
@@ -22,6 +22,7 @@ extern "C" {
 #include AVUTIL_OPT_INCLUDE
 #include AVUTIL_MATH_INCLUDE
 #include AVUTIL_ERROR_INCLUDE
+#include </usr/include/ffmpeg/libavutil/timestamp.h>
```
Baklap4, line 25:
```suggestion

```

Baklap4:
To make the comments not get lost;

looks like it's working pretty accurate; seeking backwards works again; restarting songs works flawless again, can't go into negative and at the end it stops the song (seeking forward)
The only thing i thought of when seeking backward the sound gets silent. While seeking forward it speeds up the music a lot, maybe make this silent aswell?

As for the queen song it now skips it within the song browser as it can't be played (It's not hogging memory anymore!) but i can still 'select' the song and play it.. it will result in a song which only shows the background/video and never ends this also might occur by using the webserver frontend and choosing the song from there

@OznOg remarked that once a part of the song was sung the scores can't be overwritten while rewinding and re-singing parts.
Not sure if this is intentional, but i can imagine it is since i don't have the slightest idea how we can re-calculate the score given upon a time in history.. that's probably the sole reason it can't be 're-scored' @performous/core-owners anyone who can verify this?

Tronic:
No rewinding scoring is intentional (no cheating), although it can be changed if so preferred. The scoring implementation uses (IIRC) 10 ms timeframes, where new ones are only appended to the vector whenever time has passed but never removed, this is possibly within player.cc or related classes, not within audio code. Just resizing the vector smaller doesn't do, you'd only have to rescore, so it isn't entirely trivial to rewind this. Related issues: recording of performances (even if still not implemented, this is a must-have feature) also needs to handle rewinding in one way or another.

Baklap4:
Then for now i'd say the rewinding is not an option and thus it shouldn't be allowed as it is right now.

As for the forwarding a song you've mentioned this at discord

> @Baklap4  actually, for the move forward "playing silence" it may be difficult as it is not the seek itself that creates the speed up, but the keyboard repeating forward. So there is no way to know if the seek is over or not. I could add a tempo, but may be a little bit ugly

Also this for now would be nice to have but does not block this PR.

Tronic:
Rewinding can be useful for going back to look at pitch waves, if something interesting happened.

OznOg:
> There are several indenting issues.
> Most files have indents of/equal to 8 spaces

Tank you for the feed back but don't spend too much time reviewing this for now, as I told you, this is early prototyping, I aim to clean up and have everything match coding style.
I let you know when ready for reviews (That said, any suggestion is welcome, I just don't want you to waste your time)

Baklap4:
Ghehe not wasting the time 👍 you're already contributing good fixes so keep the work up 👍 As for me i don't have other comments for it yet :)

game/screen_songs.hh
```diff
@@ -9,7 +9,7 @@
 #include "video.hh"
 #include "playlist.hh"
 #include "menu.hh"
-#include <aubio/fvec.h>
+#include "aubio/aubio.h"
```
Lord-Kamina, line 12: Why change this?

OznOg: does not compile actually mainline relies on ffmpeg.hh to pull it, But my patches isolate ffmpeg.h thus here does not compile. I don't remember the exact error, is that really a problem?

OznOg: (FYI still work in progress expect many changes in near future...)

Lord-Kamina: I know it might change and it's not really a problem; I'm just curious as to why you're doing some things. I noticed also you got rid of the circular buffer; won't that translate to using more memory?

OznOg: Actually, there is one ring buffer, but that matches better what I need to do. the boost one doe not allow random access which is what I need it. I saw there is a ringbuffer class already in the code, I need to check if I can use it (for now it is a custom implementation). For the memory, actually, I could really drop down the amount needed (I tried 40k during tests) but I kept the value 4320256 as you seem to have chose it carefully.
The main idea in these bunch of patches is to have the decoder thread separated from ffmpeg so that ffmpeg class is just reading and resampling/scaling on demand. I also split ffmpeg in two child classes (for video and audio) which allows to separate clearly things needed on each sides.

Lord-Kamina: I see.
The buffer size is set to fit roughly 30 seconds (and a bit extra) which is essentially  all the samples in the preview on the song selection screen. The idea is that the aubio code runs only once per song.

Lord-Kamina: However, according to the boost documentation, circular_buffer **does** allow random access iterators, or are you talking about something else?

OznOg: Actually it allows random access int the current range. But here I use completelly arbitrary addresses: with boost, I needed to "fill" all the beginning of ringbuf when "seeking" to the end, moreover, accessing the internal "indexes" is made difficult by the API. I must say I spent quite a lot of time trying to keep it, but I eventually decided this was really overkill and not really adapted to what I was doing (for example, iterators and so on). I end up with something that is really close to what you do when using alsa api, with the read pointer and the write pointer.

Lord-Kamina: I guess we'll have to compare performance and memory usage when you're done. 🤷🏼‍♂️. Re: the buffer size, the idea is aubio code runs on song preview load and only once at the beginning (because by nature the tempo algorithm has a latency of samples it needs to process before producing a useful result and also because IMO it's better for the song to take a couple or even several milliseconds more to load than have it stutter as new tempos are calculated). 

I don't mind you shrinking the size if it still works as expected so do feel free to experiment.

OznOg: Yes, for sure. And if you have some specific benches and/or samples with specific "problems" (like the corrupt mp3 of bug #444) I would be happy to test them.

Tronic: I doubt that there is any noticeable performance difference between `std::deque` or a circular buffer. The only real reason to use circular buffers in Performous is to allow lock-free access from PortAudio callbacks, where you are not allowed to wait for anything (things like allocating memory or acquiring a lock will occasionally cause delays that break audio recording and playback).

OznOg: I think this is now ready for reviews.

game/ffmpeg.hh
```diff
@@ -110,51 +33,110 @@ extern "C" {
...
 	// libav-specific variables
 	int m_streamId = -1;
-	int m_mediaType;  // enum AVMediaType
 	std::unique_ptr<AVFormatContext, decltype(&avformat_close_input)> m_formatContext{nullptr, avformat_close_input};
 	std::unique_ptr<AVCodecContext, decltype(&avcodec_free_context)> m_codecContext{nullptr, avcodec_free_context};
+};
+
+class AudioFFmpeg : public FFmpeg {
+  public:
+	using AudioCb = std::function<void(const std::int16_t *data, size_t count, int64_t sample_position)>;
```
Lord-Kamina, line 77: Why `int64_t`?
I think that is probably the reason why the macOS build is failing due to conflicting parameter types in std::min.

OznOg: For the 'signed' reason: position can be negative as start position in ffmpeg may not be 0 and thus the first frame decoded have negative timestamps.
For the size, well, 64 is 'large enough', may have been 32bits

Lord-Kamina: Then I guess you should static_cast count to int64_t before comparing them?

OznOg: I forced the type of `write_pos_in_ring` to `size_t` (see the patch added) so that the std::min is done on size_t and the compiler does not complain anymore.

game/video.cc
```diff
@@ -1,27 +1,99 @@
...
+					push(Bitmap());
+					std::clog << "ffmpeg/debug: done loading " << file << std::endl;
+					l.lock();
+                                        m_cond.wait(l, [this]{ return m_quit || m_seek_asked; });
+				} catch (std::exception& e) {
+					std::clog << "ffmpeg/error: " << file << ": " << e.what() << std::endl;
+					if (++errors > 2) { std::clog << "ffmpeg/error: FFMPEG terminating due to multiple errors" << std::endl; break; }
+					l.lock();
+				}
+			}
```
Tronic, line 80: The lock handling here is becoming quite convoluted. I'd like to see another PR to refactor this, avoiding manual unlocking and locking. Maybe with separate unique_lock objects for each critical section. But that requires further consideration to avoid unintentionally breaking anything. For instance, *possibly* `m_quit` could be an atomic needing no lock protection but *only* if other parts of code do not rely on this loop possibly executing one extra round after the flag has been set (I did not look into whether this would actually be a problem, this is just an example).

OznOg: I agree, for the ugliness of those locks here. The main reason is the handling of exceptions mixed up with locking. I'll try to find better idea on this... For the m_quit, actually it must be protected by the same lock as the condvar predicat depends on its value. Removing the need on lock on it would possibly deadlock on stop. That said, the lock here is not a problem per se, locking things do not mean 'hang'. There should not be 'hangs'  as lock is never held during blocking calls. If you experience a hang because of this, please let me know how to reproduce it, as it would be a bug (obviously).

OznOg: have a look to updated code, used an "unlock_guard". This reverses the paradigm but pointing in code where we **do not** want lock to be held. IMO, more natural when reading, and less chances to miss a unlock/lock with exceptions.
What do you think?

game/ffmpeg.hh
Lord-Kamina: Fair enough.

OznOg:
Any new comment?

game/screen_audiodevices.hh
```diff
@@ -32,7 +32,7 @@ class ScreenAudioDevices: public Screen {
 
 	Audio& m_audio;
 	std::unique_ptr<ThemeAudioDevices> m_theme;
-	unsigned int m_selected_column;
+	unsigned int m_selected_column { 0 };
```
Tronic, line 35: Might prefer `= 0` or `{}` but doesn't really matter.

game/video.cc
```diff
@@ -1,27 +1,101 @@
...
+	m_cond.notify_all();
+	m_grabber.get();
+}
+
+Video::Video(fs::path const& _videoFile, double videoGap): m_videoGap(videoGap), m_textureTime(), m_alpha(-0.5, 1.5) {
+	// make ffmpeg here to get any exception in the current thread
+	auto ffmpeg = std::make_unique<VideoFFmpeg>(_videoFile, [this] (auto f) { push(std::move(f)); });
+	m_grabber = std::async(std::launch::async, [this, file = _videoFile, ffmpeg = std::move(ffmpeg)] {
+			int errors = 0;
+			std::unique_lock<std::mutex> l(m_mutex);
```
Tronic, line 54: Two levels of indent? I don't think this is covered in the style guide but isn't the normal practice to indent only once even if multiple braces are opened on the same line?

game/video.hh
```diff
@@ -2,28 +2,45 @@
...
 
   private:
-	FFmpeg m_mpeg;
-	double m_videoGap;
-	Bitmap m_videoFrame;
+	const double m_videoGap;
 	Texture m_texture;
 	double m_textureTime;
-	double m_lastTime;
+	double m_readPosition = 0;
```
Tronic, line 24: `= 0.0` maybe (yes, another pedanticism that doesn't matter).

game/video.cc
```diff
@@ -50,38 +50,38 @@ Video::Video(fs::path const& _videoFile, double videoGap): m_videoGap(videoGap),
...
+				UnlockGuard<decltype(l)> unlocked(l); // release lock during possibly blocking ffmpeg stuff
+				ffmpeg->handleOneFrame();
+				errors = 0;
+			} catch (FFmpeg::Eof&) {
+				{
 					UnlockGuard<decltype(l)> unlocked(l); // release lock for possibly blocking calls
-					std::clog << "ffmpeg/error: " << file << ": " << e.what() << std::endl;
-					if (++errors > 2) { std::clog << "ffmpeg/error: FFMPEG terminating due to multiple errors" << std::endl; break; }
+					push(Bitmap()); // EOF marker
+					std::clog << "ffmpeg/debug: done loading " << file << std::endl;
```
Tronic, line 76: Due to concurrency bugs in logging implementation, things like that should be written as
`std::clog << "ffmpeg/debug: done loading " + file + "\n" << std::flush;`

... but it is noted that there are plenty of logging statements in Performous that still violate this, and sometimes messages do get garbled.

game/screen_audiodevices.hh
OznOg: done

game/video.cc
OznOg: done

game/video.hh
OznOg: done

game/video.cc
Tronic: I suggest a separate PR for fixing these, or to replace the logging system entirely, if anyone is up for that.

OznOg: ok, let's do that

PR #536, opened 2020-04-13
Fix some leaks

OznOg:
Too many leaks make performous quite unusable after 30minutes...

for instance, just starting a song and exiting the software reports:
before patches:
SUMMARY: AddressSanitizer: 717529669 byte(s) leaked in 333158 allocation(s).
after:
now SUMMARY: AddressSanitizer: 143451 byte(s) leaked in 4998 allocation(s).

So there are actually remaining leaks, mainly reported in libfontconfig, but I cannot figure out how to fix them (documentation is poor, and lib seems to leak anyway...)

### DIFF
--- a/game/glshader.cc
+++ b/game/glshader.cc
@@ -162,6 +162,11 @@ Shader& Shader::link() {
 
 	// Link and check status
 	glLinkProgram(program);
+
+	// always detach shaders, linked or not, they need to be detached
+	for (auto id : shader_ids)
+		glDetachShader(program, id);
+
 	glGetProgramiv(program, GL_LINK_STATUS, &gl_response);
 	dumpInfoLog(program);
 	if (gl_response != GL_TRUE) {
--- a/game/glshader.cc
+++ b/game/glshader.cc
@@ -156,16 +156,14 @@ Shader& Shader::link() {
 		throw std::runtime_error("Couldn't create shader program.");
 	}
 	// Attach all compiled shaders to it
-	for (ShaderObjects::const_iterator it = shader_ids.begin(); it != shader_ids.end(); ++it)
-		glAttachShader(program, *it);
+	for (auto id : shader_ids) glAttachShader(program, id);
 	ec.check("glAttachShader");
 
 	// Link and check status
 	glLinkProgram(program);
 
 	// always detach shaders, linked or not, they need to be detached
-	for (auto id : shader_ids)
-		glDetachShader(program, id);
+	for (auto id : shader_ids) glDetachShader(program, id);
 
 	glGetProgramiv(program, GL_LINK_STATUS, &gl_response);
 	dumpInfoLog(program);
--- a/game/glutil.cc
+++ b/game/glutil.cc
@@ -13,19 +13,8 @@ namespace glutil {
 		return result;
 	}
 		
-	VertexArray::VertexArray() {
-		glGenVertexArrays(1, &m_vao);
-		glGenBuffers(1, &m_vbo);
-	}
-
-	VertexArray::~VertexArray() {
-		clear();
-	}
-
 	void VertexArray::clear() {
 		m_vertices.clear();
-		glDeleteVertexArrays(1, &m_vao);
-		glDeleteBuffers(1, &m_vbo);
 	}
 
 	void VertexArray::draw(GLint mode) {
--- a/game/glutil.cc
+++ b/game/glutil.cc
@@ -24,6 +24,8 @@ namespace glutil {
 
 	void VertexArray::clear() {
 		m_vertices.clear();
+		glDeleteVertexArrays(1, &m_vao);
+		glDeleteBuffers(1, &m_vbo);
 	}
 
 	void VertexArray::draw(GLint mode) {
--- a/game/glutil.hh
+++ b/game/glutil.hh
@@ -80,11 +80,7 @@ namespace glutil {
 	private:
 		std::vector<VertexInfo> m_vertices;
 		VertexInfo m_vert;
-		GLuint m_vbo = 0, m_vao = 0;
 	public:
-		VertexArray();
-		~VertexArray();
-
 		VertexArray& vertex(float x, float y, float z = 0.0f) {
 			return vertex(glmath::vec3(x, y, z));
 		}
--- a/game/opengl_text.cc
+++ b/game/opengl_text.cc
@@ -9,14 +9,16 @@
 #include "fs.hh"
 
 void loadFonts() {
-	FcConfig *config = FcInitLoadConfig();
+	auto config = std::unique_ptr<FcConfig, decltype(&FcConfigDestroy)>(FcInitLoadConfig(), &FcConfigDestroy);
 	for (fs::path const& font: listFiles("fonts")) {
-		FcBool err = FcConfigAppFontAddFile(config, reinterpret_cast<const FcChar8*>(font.string().c_str()));
+		FcBool err = FcConfigAppFontAddFile(config.get(), reinterpret_cast<const FcChar8*>(font.string().c_str()));
 		std::clog << "font/info: Loading font " << font << ": " << ((err == FcTrue)?"ok":"error") << std::endl;
 	}
-	if (!FcConfigBuildFonts(config))
+	if (!FcConfigBuildFonts(config.get()))
 		throw std::logic_error("Could not build font database.");
-	FcConfigSetCurrent(config);
+
+        // FcConfigSetCurrent increments the refcount of config, thus the local handle on config can be deleted safely.
+	FcConfigSetCurrent(config.get());
 
 	// This would all be very useless if pango+cairo didn't use the fontconfig+freetype backend:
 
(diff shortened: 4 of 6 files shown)


game/glutil.cc
```diff
@@ -24,6 +24,8 @@ namespace glutil {
 
 	void VertexArray::clear() {
 		m_vertices.clear();
+                glDeleteVertexArrays(1, &m_vao);
+                glDeleteBuffers(1, &m_vbo);
```
Tronic, line 28: Incorrect indentation (spaces, should be tab).

game/glshader.cc
```diff
@@ -168,6 +168,8 @@ Shader& Shader::link() {
 		throw std::runtime_error("Something went wrong linking the shader program.");
 	}
 	ec.check("glLinkProgram");
+	for (ShaderObjects::const_iterator it = shader_ids.begin(); it != shader_ids.end(); ++it)
+		glDetachShader(program, *it);
```
Tronic, line 172: Maybe this could be a for-each loop instead of iterator based? Or ideally, use RAII to detach them instead of manual deletion.

game/glutil.cc
Lord-Kamina: This is wrong, VertexArray is a client buffer, VAO and VBO should only be created and destroyed once.

game/glshader.cc
OznOg: This is not deletion, this needs to be done prior to deletion, apparently, deletion is done in destruction of the class. That said, the doc seem to indicate that deletion could happen here as well, as shaders are said to be useless when progam (shaded) is linked.
Do I perform deletion there as well?

game/glutil.cc
OznOg: 'k

game/glshader.cc
Tronic: Sounds like both should be handled in the destructor together, and this loop could be replaced with something like `shader_ids.clear()`.

OznOg: wel that's not that simple, shaders here are just int8, so they don't have destructors. Not sure creating a wrapper class here is really useful.

game/glutil.cc
```diff
@@ -24,6 +24,8 @@ namespace glutil {
 
 	void VertexArray::clear() {
 		m_vertices.clear();
+		glDeleteVertexArrays(1, &m_vao);
+		glDeleteBuffers(1, &m_vbo);
```
Lord-Kamina, line 28: Here's a link to support why I don't delete them here:

https://community.khronos.org/t/understanding-vaos-vbos-and-drawing-two-objects/72778/2

OznOg:
> As I said, this is wrong; VAOs and VBOs should be created once during initialization and be disposed of when we don't need to draw any longer.
> 
> Here's a link to support this:
> 
> https://community.khronos.org/t/understanding-vaos-vbos-and-drawing-two-objects/72778/2

not sure I understand, things are done in destructor of the class, where should they be called?

Lord-Kamina:
> > As I said, this is wrong; VAOs and VBOs should be created once during initialization and be disposed of when we don't need to draw any longer.
> > Here's a link to support this:
> > https://community.khronos.org/t/understanding-vaos-vbos-and-drawing-two-objects/72778/2
> 
> not sure I understand, things are done in destructor of the class, where should they be called?

Destructor of the GL Context; currently I was calling them in the Window destructor.

Lord-Kamina:
Yup; the VertexArray constructor should just be a default constructor doing nothing. Try that (and remove the destroy calls from the destructor)

OznOg:
> Destructor of the GL Context; currently I was calling them in the Window destructor.

actually, the GLContext deletion is also added by this patch, not done before check the changes; is that correct int he new way?

Lord-Kamina:
> > Destructor of the GL Context; currently I was calling them in the Window destructor.
> 
> actually, the GLContext deletion is also added by this patch, not done before check the changes; is that correct int he new way?

Looks like overkill to me, because the GLContext most likely will only be destroyed when performous is quitting.

Take a look at this:

https://stackoverflow.com/questions/27340558/sdl-delete-gl-context-memory-safety

OznOg:
> > > Destructor of the GL Context; currently I was calling them in the Window destructor.
> > 
> > 
> > actually, the GLContext deletion is also added by this patch, not done before check the changes; is that correct int he new way?
> 
> Looks like overkill to me, because the GLContext most likely will only be destroyed when performous is quitting.
> 
> Take a look at this:
> 
> https://stackoverflow.com/questions/27340558/sdl-delete-gl-context-memory-safety

This post is really arguable and is mostly the point of view. The object doesn't know when and where it will be used in the future, and correct memory management is making software easier to maintain, IMHO.
Moreover, chasing leaks is so painful that leaking "on purpose" is just adding noise to analysis, and making it harder.

OznOg:
> Yup; the VertexArray constructor should just be a default constructor doing nothing. Try that (and remove the destroy calls from the destructor)

done

game/glshader.cc
Tronic: In fact, I believe that the code still leaks memory when linking the program fails, because then an exception is thrown and this cleanup loop is not executed. C++ doesn't have try-finally because you are supposed to use RAII for everything.

Attach/Detach is a clear pair that warrants for a simple RAII wrapper and a separate std::vector inside this function to hold the RAII wrappers.

Ideally, CreateShader/DeleteShader (current code leaks on shader compile error) and LinkProgram/DeleteProgram would be similarly wrapped (possibly using move semantics), but perhaps all that is good to be left for another PR.

OznOg: what about moving the detach just after the `glLinkProgram(program);`? I agree that RAII is the 'normal' way to handle resources, but here, the API is C and calls are within 3 lines of code, not sure adding 20 lines of RAII wrappers would really make the code easier to read... Let me know, if you prefer the RAII, let's open a task for it.

Tronic: That should do.

Lord-Kamina:
> > > > Destructor of the GL Context; currently I was calling them in the Window destructor.
> > > 
> > > 
> > > actually, the GLContext deletion is also added by this patch, not done before check the changes; is that correct int he new way?
> > 
> > 
> > Looks like overkill to me, because the GLContext most likely will only be destroyed when performous is quitting.
> > Take a look at this:
> > https://stackoverflow.com/questions/27340558/sdl-delete-gl-context-memory-safety
> 
> This post is really arguable and is mostly an opinion. The object doesn't know when and where it will be used in the future, and correct memory management is making software easier to maintain, IMHO.
> Moreover, chasing leaks is so painful that leaking "on purpose" is just adding noise to analysis, and making it harder.



> > > > Destructor of the GL Context; currently I was calling them in the Window destructor.
> > > 
> > > 
> > > actually, the GLContext deletion is also added by this patch, not done before check the changes; is that correct int he new way?
> > 
> > 
> > Looks like overkill to me, because the GLContext most likely will only be destroyed when performous is quitting.
> > Take a look at this:
> > https://stackoverflow.com/questions/27340558/sdl-delete-gl-context-memory-safety
> 
> This post is really arguable and is mostly an opinion. The object doesn't know when and where it will be used in the future, and correct memory management is making software easier to maintain, IMHO.
> Moreover, chasing leaks is so painful that leaking "on purpose" is just adding noise to analysis, and making it harder.

It is but given that, it's probably not worth the effort and making it more difficult to read for something that most likely will last the application's entire lifecycle.

game/glshader.cc
```diff
@@ -162,6 +162,11 @@ Shader& Shader::link() {
 
 	// Link and check status
 	glLinkProgram(program);
+
+	// always detach shaders, linked or not, they need to be detached
+	for (auto id : shader_ids)
```
Lord-Kamina, line 167: This is more a stylistic concern buy I say either use brackets or don't use a line-break. Not sure what the others think?

game/glutil.cc
```diff
@@ -13,15 +13,6 @@ namespace glutil {
 		return result;
 	}
 		
-	VertexArray::VertexArray() {
-		glGenVertexArrays(1, &m_vao);
-		glGenBuffers(1, &m_vbo);
-	}
-
-	VertexArray::~VertexArray() {
```
Lord-Kamina, line 21: Don't delete this destructor, just the constructor.

OznOg: well it is empty now (just calling vector clear), thus useless IMHO

game/glshader.cc
OznOg: done

game/glutil.cc
Lord-Kamina: It's clearning a vector of containing an array of objects consisting of several arrays.

OznOg: I don't see the point here: clear deletes all elements of vector, so does the destructor.

Lord-Kamina: *shrugs* I'm not gonna hold the PR back because of this but it seems odd to me you'd remove the `clear()` call but go to plenty of trouble to clean stuff up when the application exits.

PR #509, opened 2020-03-25
Rely on pkg-config to find Pango

orivej:
Fixes build with Pango that needs `-I/usr/include/harfbuzz` (as specified in its pkg config).

PkgConfig results can be overridded by setting e.g. Pango_PKGCONF_INCLUDEDIR, Pango_PKGCONF_LDFLAGS (see [FindPkgConfig documentation](https://cmake.org/cmake/help/v3.6/module/FindPkgConfig.html)).

IMPORTED_TARGET was added in CMake 3.6.

Fixes #490
Closes #493

PR #537, opened 2020-04-14
RE: support of native Windows complation under Visual Studio

Baklap4:
### Notes

Added a new PR for the visual studio build. The other one was outdated and i had trouble rebasing it up to current master.

### What does this PR do?

Introduce some fixes allowing to build (not-standard compiler features) and run (compiler-specific asserts at STL) program with MSVC.

### Closes Issue(s)

game/animvalue.hh : added include required by std::min/std::max
game/audio.cc : removed dereferencing of end iterator (UB)
game/controllers-midi.cc : added include required by std::stringstream
game/glshader.cc : replaced variable-sized arrays to std::string to remove compiler extension requirement
game/libxml++-impl.hh
game/notes.hh : comparsion functor should not lead to result when a > b and b > a at same time (runtime-checked by compiler at debug build)
game/screen_audiodevices.cc : fixed access beyond array bounds (UB)

### Motivation

Desire to develop and debug this great game under some developer's favourite IDE

### More

### Additional Notes

game/screen_audiodevices.cc also fixes strange bug (used device selectons losts after audio device menu reopening) but  don't sure it is it correct. To be honest I don't understand what  is idx and index values at all - so please say what could I do with it.

Sorry about so complex request :'(


Baklap4:
@allcreater I moved the pr to here

game/libxml++-impl.hh
```diff
@@ -5,7 +5,7 @@
 #include <libxml++/libxml++.h>
 
 namespace xmlpp {
-#if LIBXMLPP_VERSION_2_6
+#if defined(LIBXMLPP_VERSION_2_6)
```
Lord-Kamina, line 8: Your issue is here.

game/libxml++-impl.hh
```diff
@@ -19,7 +19,7 @@ namespace xmlpp {
 	static inline void set_first_child_text(Element* element, const Glib::ustring& content) {
 		return element->set_child_text(content);
 	}
-#elif LIBXMLPP_VERSION_3_0
+#elif defined(LIBXMLPP_VERSION_3_0)
```
Lord-Kamina, line 22: And here.

The problem with changing `#if`to `#if defined()``

Check this example to understand why: https://onlinegdb.com/BJj90o7OU

game/libxml++-impl.hh
Baklap4: Ahh thanks!

allcreater:
could you give me push access to this branch, please?
I still don't have final solution how to make it work but one small fix is already awaiting :)

game/libxml++-impl.hh
allcreater: but wait just a minute, if LIBXMLPP_VERSION_3_0 is just a symbol without literals (defined by `#define LIBXMLPP_VERSION_3_0` instead of something like `#define LIBXMLPP_VERSION_3_0 1`, the correct form is with *defined* operator!

probably the source of troubles is different define forms on different platforms. MSVC don't understand the code without *defined* and it do it right.

Lord-Kamina: But the module defines it to be 0.

Lord-Kamina: Here: https://github.com/performous/performous/blob/39dad10cc289d2eb6e73cba300ae0602fd718048/cmake/Modules/FindLibXML%2B%2B.cmake#L19-L20

A possible solution would be to explicitly use `#if LIBXMLPP_VERSION_2_6 == 1`

cmake/Modules/FindPango.cmake
```diff
@@ -6,6 +6,38 @@
 
 include(LibFindMacros)
 
-libfind_pkg_check_modules(Pango_PKGCONF IMPORTED_TARGET pango)
-set(Pango_LIBRARY PkgConfig::Pango_PKGCONF)
-libfind_process(Pango)
+if(NOT WIN32)
```
Baklap4, line 9: @orivej Does this still work on your system?
Due to #509 we kinda broke the windows build we had.. 
Windows mostlikely doesn't have pkg-config installed.. i'm looking for a way to combine both worlds

Baklap4:
@allcreater I've given this a retry and succeeded! I'll open up a new pr for that but first 2 other PR's need to get merged:
https://github.com/performous/performous/pull/637
https://github.com/performous/performous/pull/636

I hope to merge these pretty soon. When they are can i ping you for testing the new PR for the windows build?

I'm closing this as this won't get merged anyway.

allcreater:
@Baklap4 great news! 
Of course I'll be glad to test it and even help with development. I still hope to make useful contribution for Performous project :)

PR #540, opened 2020-04-16
Improved and simplified frequency detection.

Tronic:
I derived a new formula for phase correction after finding out that the earlier one was slightly off. I couldn't sing test it yet, and would appreciate it if someone could try how this does against the old one in proper karaoke.


Lord-Kamina:
I don't have a means of testing it ATM but it'd be great if it works better.

Tronic:
Later parts of the pipeline should probably be adjusted as well because this finds sharper peaks than the old phase correction did, and that'll affect processing.

Tronic:
Removed a nasty bug that would cause complete failure (integer division instead of floating-point).

Tronic:
Could we get this merged? This should be good to go but PitchDetector2 might need more work and I don't know when I'll be able to look at it again.

Baklap4:
@Tronic can you bring it up to date with current master?
As for the code-changes that looks alright.

I haven't tested it yet though.
Do you have anything you wanted tested specifically?

Baklap4:
@ooshlablu tested this his response:

> ooshlablu — 08/07/2021
I'm kind of not sure how to report on 540
It built, runs, fine and everything
I think I kind of notice like more ""spikes"" when singing when the range is  thin on the dynamic view, which is what has been around forever
I also did this with the behringer/58 setup which may be really sensitive
I can also try with the usb mic

@ooshlablu did you also try it with usb mic?

ooshlablu:
I tested again with both a USB mic and and the SM58, and I got similar results with both:

This is a sing through with the SM58:
Screenshot from 2021-08-16 15-24-22

This is a sing through with the USB mic:
Screenshot from 2021-08-16 15-26-54

They both exhibit a choppiness that shows up as sharp highs and lows at some points.

For comparison, here's how it looks running in a build for PR #655:
Screenshot from 2021-08-16 15-29-32

Tronic:
The new algorithm is correct, the old one is not. But the problems you are seeing might be caused because correct frequencies pile up to higher total db level, causing the harmonics being misdetected instead of the fundamental frequency. This would be because various filtering constants in code are tuned for the old algorithm and not the correct one.

Ideally the pitchdetector2 PR would take these changes into consideration but I still don't know whether I can finish that in reasonable time.

Tronic:
@Baklap4 I added minor levels tuning. You may need to slightly increase your mic volume but it might also just work better out of the box now. Here at least in Practice I get almost clean output (most of the time only one note) and in singing it seems to behave well too.

One possible problem is if now cuts off too quickly on consonant sounds that are not toned, but for me that was not a problem in brief testing. Tested with a good mic and with Macbook built-in speakers/mic. dB level shown in practice is about -20 dB when singing, cuts off around -30 dB.

Tronic:
@ooshlablu Would you mind testing again now?

ooshlablu:
I will test this again sometime within the next 24 hours and let you know how it looks compared to the earlier tests.

ooshlablu:
@Tronic @Baklap4  I added the `Won't Fix` label to this so we don't get confused, especially since it will cause merge conflicts.

ooshlablu:
@Tronic @Baklap4 So after testing, this one looks really smooth and works really well. If I had a choice between this and #541, I would pick this one at the moment. I will post results there.

Tronic:
Thanks for all the testing. Let's see if the remaining issues of the other one are easily fixed (e.g. mic volume level, possibly implementing auto-adjust for that), but if not, this should be merged for 1.2, and the other one left for a later release as planned.

Baklap4:
Removed the `Wont Fix` label as discussed here: https://github.com/performous/performous/pull/541#issuecomment-913545642

So it gets merged for 1.2.0 release

Baklap4 (approved):
LGTM

The code in question was indeed simplified!

Maybe it's worth it documenting this process in laymen terms on a separate wiki page?

PR #541, opened 2020-04-19
Pitchdetector2 - further tuning

Tronic:
This is an experimental branch that does more than #540 which itself can better be compared with the master branch. The changes in this one seem to greatly reduce the number of extra tones detected and shown on practice screen. More testing is still needed.


game/CMakeLists.txt
```diff
@@ -93,6 +93,12 @@ foreach(lib ${OUR_LIBS} SDL2 PangoCairo LibRSVG LibXML++ AVFormat SWResample SWS
 	add_definitions(${${lib}_DEFINITIONS})
 endforeach(lib)
 
+if(LibXML++_VERSION_2_6)
+  set(LIBXMLPP_VERSION 2)
+else()
+  set(LIBXMLPP_VERSION 3)
```
Lord-Kamina, line 99: This isn't really necessary but I guess it makes it more readable?

Lord-Kamina:
Also... this is just a question/suggestion, since you're going elbow-deep into pitch processing, how do you think about adding a "Difficulty" setting like Ultrastar does? Is it easily doable?

game/CMakeLists.txt
Tronic: This review should be in #492, not here. I don't recall if there was a reason for this change other than cleanup.

Tronic:
Scoring and thus difficulty settings would be handled in an entirely different part than the pitch detector (I gather that might be in `player.cc` but you'll need to look it up). The idea has been discussed before, and there is not much use for less strict scoring because Performous already gives gradually less score with less accuracy (it is not on/off like in singstar). Not having difficulty levels keeps scores comparable, too.

game/CMakeLists.txt
Lord-Kamina: Fair enough. Anyway, it's just a comment.

Lord-Kamina:
> Scoring and thus difficulty settings would be handled in an entirely different part than the pitch detector (I gather that might be in `player.cc` but you'll need to look it up). The idea has been discussed before, and there is not much use for less strict scoring because Performous already gives gradually less score with less accuracy (it is not on/off like in singstar). Not having difficulty levels keeps scores comparable, too.

I thought maybe it could go in here so long as it is using a less strict comparison of notes expected and sung (then again, perhaps that logic IS elsewhere)

Tronic:
Analyzer only gives out detected frequencies in Hz, it doesn't know about notes.

Lord-Kamina:
> Analyzer only gives out detected frequencies in Hz, it doesn't know about notes.

Ah ok. So they get converted to notes/tones elsewhere?

Tronic:
Now it works much better than the old one. Clear and uberprecise detection in practice, and seems to be working in singing too. During testing I got one segfault but couldn't determine if it was caused by this patch or not.

Tronic:
@Lord-Kamina See `Notes::powerFactor` that's the essence of vocal scoring. `MusicalScale` for conversions between Hz and note numbers. `Player::update` for the main vocal grading loop.

Lord-Kamina:
> Now it works much better than the old one. Clear and uberprecise detection in practice, and seems to be working in singing too. During testing I got one segfault but couldn't determine if it was caused by this patch or not.

Cool. Maybe post a trace or dump if you have it to see of we can figure it out?

Tronic:
I didn't get a dump and couldn't get it to crash again.

Lord-Kamina:
Oh, well. Was it during singing or in practice mode or what?

Tronic:
This is still a work in progress. I'm developing the algorithm in Python, and what you see here is a port to C++ (curiously, the Python version runs faster). There are further developments done and more planned, that I plan to add here once complete. Unfortunately I'm still stuck to testing mostly without real microphone input, so any feedback from your experiments is more than welcome.

Baklap4:
Converted to draft according to latest comment: "This is still a work in progress"

Tronic:
It might take a very long time until I can pick this up again. If you don't mind keeping around a draft, feel free to keep it here. Possibly some feedback of the other Pitchdetector PR inspires me to complete this work still this Summer.

Tronic:
This is also available for testing. IMO this has clearly better performance than either master or pitchdetector1 but YMMV because the algorithm is largely redesigned.

If there is agreement that this algorithm works well, I suggest merging this instead of pitchdetector1. If further development on this is needed, then pitchdetector1 should probably be merged meanwhile.

Baklap4:
@Tronic this includes/is built on top of the changes from #540 right?

If that's the case @ooshlablu should test this instead and #540 can be closed

Tronic:
@Baklap4 Yes, this is intended to supersede the other PR. The new phase equation is included and levels are tuned in this as well.

The problem is that this changes a lot more, while the other one was intended to be merged without much testing due to it being a very minimal change with little expected adverse effects. I suggest several developers do thorough testing in noisy and quiet environments and with different singers before merging this.

In particular, high-pitched (female) voice needs to be tested. And since whistling is supported as a new feature, one might try that as well (be sure not to blow on your mic but beside it).

Baklap4:
> And since whistling is supported as a new feature

Always look on the bright side of life is a great song for whisteling ;)

ooshlablu:
I'll test #540 anyway since I already built the package and it seems like it's easier to test by myself.

I will do a test of this myself at some point today and see if i can get a proper group to fully test this PR over the weekend. Finding someone that can actually whistle to a song might be difficult tho :-)

Tronic:
> someone that can actually whistle to a song

It was surprisingly easy to hit the notes on Hard difficulty. The pitch wave looks different too.

ooshlablu:
For me, this is really choppy and appears to be not picking up the pitch compared to how #540 and current master is.

I tried with 2 different mics on Behringer USB interface, trying different volume levels, and a straight USB mic, and had results that look like this:
Screenshot from 2021-09-01 20-51-45
Screenshot from 2021-09-01 20-51-50


I couldn't find where to tune in the settings, and the results are similar on all the difficulty levels, just thinner lines :)

If I had to pick between #540 and this, I would pick that to merge for now. Let me know what I can do to get you any info you need to help.

Tronic:
The pitch waves show high frequency, suggesting that something other than your singing (base frequency) is being detected. Could you check in practice screen visually which notes appear? Also check your volume level, should be around -20 dB while singing. If much louder, it probably cannot work, and more silent should produce no notes.

ooshlablu:
@Tronic I see what you are saying now. If I play with the output volume from Performous very low/off it seems more stable and less jumpy overall when singing along with a song. This does have major interference with the way I play tho, which is with a group rock-band style with 1 mic and guitar controllers, rowdy,and hence high volumes through high-powered, and very responsive speakers :-).

At the moment, I can't get this to work with any sort of outside noise. I've tried adjusting the USB mic volume in pulse audio, which I don't think has any effect since this uses SDL to interface devices directly (I could be wrong, but that's how it seems to me when messing with volumes), and I also have tried turning down the input gain from the mic through the Behringer interface. At very low gain through the Behringer, things do seem to smooth out, but there are "blanks" and other chops that are definitely the result of low sensitivity at such low volumes.

@Baklap4 I think for now #540 is the better candidate that does make a difference in terms of scoring and actual game play if this is something that should make it into a new release. I think we should mark the PRs appropriately.

Baklap4:
Alright so #540 within Release 1.2.0 and #541 in an upcoming release?

ooshlablu:
Yeah, I believe that should be the path for now

Tronic:
@ooshlablu I think you need to adjust mic volume in alsamixer, switch to capture volumes on the physical device of the mic(s). If your device supports that.

Could you confirm which dB value you see in practice screen while singing? Does it ever reach the top (0 dB)?

Baklap4:
1.2.0 is released a month ago. This was expected to be merged after this. Can we get this PR up-to-date and test it? @ooshlablu

ooshlablu:
@Tronic if you could look at the conflicts, the packages will build and get attached to this PR now. I can install it and test it again and see how it looks now.

Tronic:
Hmmh, that's some pretty nasty conflicts. I'll see if I can get this compiled and running with actual mics, to be able to merge with meaningful dB limits.

twollgam:
If there may be problems with the new pitch detection why not make it configurable?

Tronic:
It is too much trouble to compile, run and use two different algorithms in parallel.

Lord-Kamina:
> It is too much trouble to compile, run and use two different algorithms in parallel.

Is it? Not saying you do it, I assume if you haven't finished it it's because you're busy with other things; but it might not be a terrible idea to put them both behind a common interface the rest of the code can interact with, and that might allow for gradual tuning and transitioning to the new algorithm.

Tronic:
If you feel like doing it then by all means. I suspect that there isn't much difference and that it really boils down to tuning the dB levels correctly (if interested, make those configurable as well and play around with it). Other than that, I believe the code is pretty much finished as is in this PR - unfortunately I haven't had time to even compile this on MacOS and give it that test run.

Tronic:
Tried building, cannot find correct libicu (complains about missing components which are in /opt/homebrew/opt/ which I have set in `CMAKE_LIBRARY_PATH` - I guess Meson worked before but you removed it).

Lord-Kamina:
> If you feel like doing it then by all means. I suspect that there isn't much difference and that it really boils down to tuning the dB levels correctly (if interested, make those configurable as well and play around with it). Other than that, I believe the code is pretty much finished as is in this PR - unfortunately I haven't had time to even compile this on MacOS and give it that test run.

I'll play around with that at some point, I think.



> Tried building, cannot find correct libicu (complains about missing components which are in /opt/homebrew/opt/ which I have set in `CMAKE_PREFIX_PATH` - I guess Meson worked before but you removed it).
> 
> EDIT: Solved by force-linking icu so that CMake finds the correct one (not the MacOS one), hopefully that doesn't break my system. Also needed a dozen extra dependencies not mentioned in the wiki (you really ought to be _reducing_ the number of deps), let's see if I ever get to compiling this...
> 
> EDIT2: Missing
> 
> ```cmake
> target_include_directories(fs SYSTEM PRIVATE ${Boost_INCLUDE_DIRS})
> ```
> 
> Trying to merge master now, there are some conflicts that affect levels.

Hmm, can you try again deleting `performous/cmake/Modules/FindICU.cmake`? CMake ships its own FindICU module, so I'm under the impression we're most likely using an ancient version, slightly modded.

~What deps are missing from the wiki? Off the top of my head, I think libpng, but not sure what else?~ nevermind, the linux list is pretty much up-to-date but not the macOS list.

I'd never had issues with the Boost.fs bit. I'll have to give that a look-see also.
That said, if you upgrade to at least macOS 10.15, performous will use `std::filesystem` instead.

Tronic:
Plays OK with Macbook internal mic and great on a table mic now. The internal mic picks up too much of the song playing. Didn't compare with pitchdetector1 other than that it also seemed to work OK, while this prior to tuning didn't.

Tronic:
As an idea for future development, Tones could keep track of whether they were selected on the previous iteration, and offer some hysteresis (notes appearing on screen should not easily disappear and notes not already on screen need to be a bit stronger before the appear - that causes annoying flicker in practice and short-lived garbage in singing with bg noise).

Tronic:
> Hmm, can you try again deleting `performous/cmake/Modules/FindICU.cmake`? CMake ships its own FindICU module, so I'm under the impression we're most likely using an ancient version, slightly modded.

Didn't help.

> ~What deps are missing from the wiki? Off the top of my head, I think libpng, but not sure what else?~ nevermind, the linux list is pretty much up-to-date but not the macOS list.

Some might be pulled by other deps or supposed to be included in src tree but I had to install
```
fmt aubio nlohmann-json cairo glibmm
```

> I'd never had issues with the Boost.fs bit. I'll have to give that a look-see also. That said, if you upgrade to at least macOS 10.15, performous will use `std::filesystem` instead.

I'm on MacOS 12.4.

Lord-Kamina:
> > Hmm, can you try again deleting `performous/cmake/Modules/FindICU.cmake`? CMake ships its own FindICU module, so I'm under the impression we're most likely using an ancient version, slightly modded.
> 
> 
> 
> Didn't help.
> 
> 

Damn. Will have to look into it.


> 
> > ~What deps are missing from the wiki? Off the top of my head, I think libpng, but not sure what else?~ nevermind, the linux list is pretty much up-to-date but not the macOS list.
> 
> 
> 
> Some might be pulled by other deps or supposed to be included in src tree but I had to install
> 
> ```
> 
> fmt aubio nlohmann-json cairo glibmm
> 
> ```
> 

Sounds about right. Aubio and K think also Nlohmann can be fetched and built by performous using command-line options. Still, gotta update that wiki at some point.

> 
> 
> > I'd never had issues with the Boost.fs bit. I'll have to give that a look-see also. That said, if you upgrade to at least macOS 10.15, performous will use `std::filesystem` instead.
> 
> 
> 
> I'm on MacOS 12.4.
> 
> 

There's clearly a bug in the logic there, then.

Lord-Kamina:
BTW. Are you running from the terminal and/or codesigning the binaries?

Tronic:
> BTW. Are you running from the terminal and/or codesigning the binaries?

Running from terminal, AFAIK not codesigning, just normal make / make install.

Lord-Kamina:
> > BTW. Are you running from the terminal and/or codesigning the binaries?
> 
> 
> 
> Running from terminal, AFAIK not codesigning, just normal make / make install.
> 
> 

I see. We're currently seeing an issue where performous only works properly when launched from Terminal (or xcode). The app bundle doesn't detect any mic input. This is due to the security restrictions starting in Big Sur. It appears the only possible fix is to bite the bullet and get a paid developer account, which I'm honestly considering (we were also talking with the Ultrastar people about the possibility of sharing one or some other akin solution)

Tronic:
Any progress with getting this tested, reviewed and merged?

ooshlablu:
@Tronic I'm still having major issues when I use the artifacts produced in the PR. I'm still seeing the very choppy peaks as before, but also now it seems to not pick up my fundamental note, it's down by almost a third. I tried with several different mics, even an SM 58 which is a cardioid mic and should avoid most interference, but I'm seeing the same thing.

Tronic:
> @Tronic I'm still having major issues when I use the artifacts produced in the PR. I'm still seeing the very choppy peaks as before, but also now it seems to not pick up my fundamental note, it's down by almost a third. I tried with several different mics, even an SM 58 which is a cardioid mic and should avoid most interference, but I'm seeing the same thing.

Can you try with `performous --log pitch` and with uncommented pitch logging line in code (you'll need to compile it yourself to uncomment that line in analyzer.cc). We are interested on whether the correct frequency appears on the pitch line and is missing on the tones line that directly follows it, or whether the tones can in other way be explained by the peaks that it sees. Pasteing a piece of that log at the moment where it fails would be helpful.

berarma:
@Tronic Are you still interested? I could help testing it.

Tronic:
> @Tronic Are you still interested? I could help testing it.

Let me know if you have results. I won't be able to dig much into it at this time though.

Issue #486, opened 2019-09-30
off-by-one in microphone colours


mirabilos:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Try to assign the microphone a colour.

### What did you expect to see?

The chosen colour is used.

### What did you see instead?

The one to the left is used (or, for the red one, an invisible-in-the-config blue one is used).

![Screenshot_20190930_170254](https://user-images.githubusercontent.com/861078/65891363-a8581f80-e3a4-11e9-84af-08c603011dab.png)

This ↑ is a screenshot of the configuration dialogue. I have an orange microphone connected, but I need to select one to the right. If I select orange, it gets pink instead ☹

The default left-most one is shown as red but shows up as blue during gameplay.

### Output of `performous --version`: (_What version of Performous are you using?_)

```
Performous 1.1+
```

git commit 9928c27

### What is your environment & configuration (arguments, platform, ...)?

Debian GNU/Linux sid/amd64

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```

Generic options:
  -h [ --help ]          you are viewing it
  -l [ --log ] arg       subsystem name or minimum level to log
  -v [ --version ]       display version number
  --songlist arg         save a list of songs in the specified folder

Configuration options:
  --audio arg            specify an audio device to use
  --audiohelp            print audio related information
  --jstest               utility to get joystick button mappings

ERROR: unrecognised option '--logLevel=DEBUG'
```

Lord-Kamina:
This is really weird? I've never seen it happen. Will have to investigate. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Lord-Kamina:
@Baklap4 have you been able to reproduce this?

Baklap4:
Never seen this happen, currently closing as cannot-reproduce. @mirabilos if this is still an issue on current master please let us know :)

PR #542, opened 2020-04-19
Fix offbyone in devices

OznOg:
### What does this PR do?

Fix an off-by-one access in devices array in devices screen

### Closes Issue(s)

don't think so

### Motivation
Was crashing on my machine

### Additional Notes
I did not use range-based for loop because i is use inside loop body.


Baklap4:
Perhaps links to this issue; https://github.com/performous/performous/issues/486 ?

Baklap4:
@mirabilos can you try and see if this fixes #486 ?

OznOg:
Update?

mirabilos:
OznOg dixit:

>Update?

Sorry, didn’t get around to doing so yet, with the current global situation.
Will try to.

Lord-Kamina:
> @mirabilos can you try and see if this fixes #486 ?

I don't know... I've never been able to reproduce #486

mirabilos:
> @mirabilos can you try and see if this fixes #486 ?

@Baklap4 I cannot reproduce #486 on my laptop; judging from the screenshot it was taken on my desktop machine at $dayjob, which is not easily reached due to the current global situation.

I’ll put it into my TODO, but don’t hold this off on my account.

Issue #452, opened 2019-01-08
Feature: Drop cpprestsdk, use restinio + nlohmann-json


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
Currently we aren't able to crosscompile for windows :( due to cpprestsdk not being portable through mxe: https://github.com/Microsoft/cpprestsdk/issues/798 since they ignore us there was a thinking about dropping it and replace it with [Restinio](https://github.com/Stiffstream/restinio) as networking library and [Nlohmann-json](https://github.com/nlohmann/json) as json library.

Lord-Kamina:
Once I finish the BPM/cover flashing thing I'll continue working on this.

Baklap4:
restinio looks pretty simple and already pretty equal to the current code.. i don't think we need to change that much

nieknooijens:
why  Nlohmann-json ? json-cpp worked fine!

Lord-Kamina:
Nlohmann has a modern interface, no extra dependencies, is more comformant to standard, uses significantly less memory and is significantly faster than json-cpp.

Lord-Kamina:
So... it seems somebody at https://github.com/microsoft/vcpkg/issues/3865 is working on a librsvg port for vcpkg; which might mean it would soon be possible to compile performous directly from windows (wouldn't that be awesome?).

If this does come to pass we won't NEED to drop cpprest (although we might still _choose_ to do so in the future for better performance but that's another issue)

Baklap4:
Linking #494 
We're almost there but having currently issues with the latest changes on master regarding pkgconfig. Hopefully soon fixed

PR #545, opened 2020-05-01
[WIP] Port webserver from cpprestsdk

Lord-Kamina:
### What does this PR do?
Migrates our webserver to Nlohmann JSON and either Restinio or Restbed (not fully decided yet, appreciate input from @performous/core-owners); both have pros and cons re: dependencies.

- Restinio depends on Boost.Asio, http_parser and libfmt (other stuff too but those are what we'd need): On one hand we already depend on Boost.Asio so it's not a big deal. http_parser we could either use it externally or have it build and link statically from the restinio tree. libfmt more-or-less the same.

- Restbed depends on standalone Asio (and an old version at that): That would IMO a) be a potential security risk (although probably not too important considering our context) and I do expect they'll update it at some point and/or support Boost.Asio. Chief advantage is I think Restbed's developers might be saner and using Restbed would probably require less patching of the source-code and build-scripts on our part.

### Closes Issue(s)
#452

### Motivation

CppRestSDK is full of Microsoft ew.

### More


Tronic:
RESTinio definitely looks like the more sane option here, although the webserver will still come with heavy dependencies so it should still be possible to disable the feature at build time. Restbed's API is ridden with shared_ptrs, which is not a very good sign, and seems more cumbersome from their examples, compared to RESTinio which actually has a modern feel to it and uses concepts similar to NodeJS Express.

Lord-Kamina:
> RESTinio definitely looks like the more sane option here, although the webserver will still come with heavy dependencies so it should still be possible to disable the feature at build time. Restbed's API is ridden with shared_ptrs, which is not a very good sign, and seems more cumbersome from their examples, compared to RESTinio which actually has a modern feel to it and uses concepts similar to NodeJS Express.

Generally, my impression towards RESTinio has always been that it's a very good library but its developers make weird decisions hahaha

Lord-Kamina:
Now is when the real work begins. This has a ways to go still but in the meantime... @performous/core-owners (especially, @Tronic); is what I did to the Restinio CMakeLists acceptable? Essentially the idea is if you have the restinio dependencies installed in your system then cool you can use them but else they come bundled and don't require any extra interaction.

Tronic:
I'm too busy to look into that but generally I dislike bundling things, especially anything that requires a build (i.e. is not header-only). I believe that the packagers of distributions such as Debian are obligated to revert these and package each dependency separately in any case.

Lord-Kamina:
> I'm too busy to look into that but generally I dislike bundling things, especially anything that requires a build (i.e. is not header-only). I believe that the packagers of distributions such as Debian are obligated to revert these and package each dependency separately in any case.

Fmt is header only and http-parser is a single header and a single c file

Baklap4:
Just getting into this PR but as for now i'd suggest to split nlohmann-json and restinio in seperate PR's. As you've already done in the code. This will help for an easier merge 👍

data/config/schema.xml
```diff
@@ -397,5 +380,104 @@ to save the current settings to XML.
...
+		<enum>3</enum>
+		<enum>4</enum>
+		<enum>5</enum>
+		<enum>6</enum>
+		<enum>7</enum>
+		<enum>8</enum>
+		<enum>9</enum>
+		<enum>10</enum>
+		<enum>11</enum>
+		<enum>12</enum>
```
Tronic, line 409: If you are going to list enum options, maybe take another step and list the actual mask? Otherwise, couldn't you just use min and max value? Most of the values also don't make sense, e.g. 1 bit mask doesn't do anything useful.

Lord-Kamina: 1) Using int doesn't work.
2) These aren't bitmasks but slash notation; 1 is /1 or also 128.0.0.0. Arguably you're right that it's not very useful but somebody might still want to use that. Either way, I'm thinking I'll probably leave only 16 to 31 or so in the UI and leave a note in the schema saying if they want more they can edit it directly.

Tronic: Actually I haven't even seen netmask definition with any other webserver, and I've worked with a lot of them. Normally you specify bind addresses and/or hostnames (a name lookup may return multiple addresses to bind to like `::1` and `127.0.0.1` for `localhost`). System services may need to always bind `0.0.0.0` and `::` and then do custom filtering to work with interfaces that may go up and down or change IPs while the service is running, but that is hardly required with a game like Performous. Please try to keep things simple.

As to int type and limits: you should add limits to int types if not currently supported in `configuration.cc`.

Lord-Kamina: Ints do support limits but not converting this to a simple user-friendly representation in the UI, which is what I'm doing so far. Anyway, the average user will never have to mess with this (the idea is to have Disabled, localhost, and all connections) just like we do now, but you can also select the custom netmask so you can do things like "only accept connections from WLAN"

Tronic: For that user experience, you probably want a single config option that allows `Internet`, `LAN` and `localhost` only, and looks up the interface address and iface's netmask for the middle option. Another feasible option is to *not* limit by address, always bind `0.0.0.0` and `::` but add an auth token (e.g. included in URL shown as QR code in-game).

Lord-Kamina: > For that user experience, you probably want a single config option that allows

The problem is, we have no way to determine from (C++) what constitutes "LAN", so the best bet to support it is to just let the user decide. Again, this is an option that I think is nice to have (and chiefly, because it's pretty easy to implement) but I doubt many people will actually use.

Lord-Kamina: Also, both "Custom netmask" and "All Connections" (which is essentially another name for "Internet") will bind to 0.0.0.0; the difference is the former will also make use of RESTinio's IPBlocker feature.

game/requesthandler.hh
```diff
@@ -18,12 +19,37 @@ using Performous_Router_t = restinio::router::express_router_t<restinio::router:
...
-using Performous_Server_Traits = restinio::traits_t<
-         restinio::asio_timer_manager_t,
-         restinio::shared_performous_logger_t,
-         Performous_Router_t,
-         boost::asio::strand<boost::asio::executor>>;
-       
+class Performous_IP_Blocker {
+	public:
+		restinio::ip_blocker::inspection_result_t inspect(const restinio::ip_blocker::incoming_info_t& info) noexcept {
+			if (config["webserver/access"].i() != 3) return restinio::ip_blocker::allow(); // Just in case.
```
Baklap4, line 26: Can this `3` be changed to an enumeration? i don't like magic numbers :)

game/requesthandler.cc
```diff
@@ -3,24 +3,16 @@
...
-//         	[&](auto& settings) {
-//         		settings
-//                 .port( port )
-//                 .address( url )
-//                 .separate_accept_and_create_connect(true)
-//                 .request_handler(std::make_unique<Performous_Router_t>());
-//                 } {}
-
 RequestHandler::RequestHandler(std::string url, unsigned short port, Songs& songs):m_listener("http://" + url), m_songs(songs), m_restinio_server(restinio::own_io_context(), make_server_settings(url, port)) {
+	m_local_ip = RequestHandler::getLocalIP(std::string("1.1.1.1"));
```
Baklap4, line 7: Hmmm what is this ip address?

Lord-Kamina: CloudFlare's DNS. It's what we were already using.

game/requesthandler.hh
Lord-Kamina: It is an enumeration but I believe all configuration enums are implemented as int underneath?

game/requesthandler.cc
```diff
@@ -64,40 +64,129 @@ boost::asio::ip::address_v4 RequestHandler::getLocalIP(const std::string& servic
...
+        std::string content_type = std::string();
+        if(path.find(".js") != std::string::npos) {
+            content_type = "text/javascript; charset=utf-8";
+        } else if (path.find(".css") != std::string::npos) {
+            content_type = "text/css";
+        } else if (path.find(".png") != std::string::npos) {
+            content_type = "image/png";
+        } else if (path.find(".gif") != std::string::npos) {
+            content_type = "image/gif";
+        } else if (path.find(".ico") != std::string::npos) {
```
Tronic, line 154: I would suggest
```c++
std::map<std::string, std::string> mimetypes{
    {"png", "image/png"},
    ...
};
```

game/requesthandler.cc
Baklap4: Alright 👍 now i remember we needed a 3rd party so we can get our ip address

game/requesthandler.cc
Lord-Kamina: The problem with this approach is we're using dumb matching so most likely it'd end up being the same.

game/requesthandler.hh
Baklap4: Not sure, however an typed enum will just make it more readable, maybe not a part of this pr

game/requesthandler.cc
Tronic: Seriously I don't know what the fuck to do with you. I understand that you may not be an experienced programmer and that you could come up with less than optimal solutions. However, after seeing you work on Performous for a decade and *still* most of your commits being shit code, and when I give feedback on how to do it better, you **always** just dismiss it and do your funky style anyway. If you don't wish to listen to my feedback, I'm gonna give up. Please feel free to keep fucking up the code.

Lord-Kamina: I honestly don't know what the fuck your problem is. I do value and take your input into account but on the contrary, I'd say it's you who always does what he wants.

Further, what I'm saying is we're doing a dumb match for a ".extension", using a map would still mean iterating through every element in the map and comparing it because we don't have an extension string, properly speaking.

I'm sure we can improve the code to be more optimal but right now you're riding my ass and being utterly unsufferable over WIP code and in particular a piece of code that wasn't even written by me! (That is the code in master; I just changed the order and some of the cases).

If you want to improve something in a particular way, you can commi fixes just as much as I can.

Baklap4: Let's not attack eachother please. The code was re-ordered that's all. Yes it can/should be improved but please let's be nice to eachother.

Lord-Kamina:
Indentation is still a mess but I'd rather fix that in another PR so as to not make this huge diff even harder to read.

Baklap4:
Looking good so far, however sorting the library (on the frontend) doesn't work correctly. The list doesn't get sorted as if it doesn't take query params atm

Lord-Kamina:
> Looking good so far, however sorting the library (on the frontend) doesn't work correctly. The list doesn't get sorted as if it doesn't take query params atm

Ah, I missed that. I'm pretty sure I implemented the code for sorting but I must have forgotten to catch the actual press or something.

Lord-Kamina:
> Looking good so far, however sorting the library (on the frontend) doesn't work correctly. The list doesn't get sorted as if it doesn't take query params atm

Should be fixed now.

The idea is that now both sorting and filtering are completely independent between UI and webserver; else, it was very easy for someone to troll you from the webserver.

Baklap4:
Hmmm while loading my song library i notice a lot of '404' upon adding a song which is already found in the frontend. A little bit later this is fixed i suspect some threading issue not synching the state of m_songs correctly

Lord-Kamina:
> Hmmm while loading my song library i notice a lot of '404' upon adding a song which is already found in the frontend. A little bit later this is fixed i suspect some threading issue not synching the state of m_songs correctly

You mean when adding to playlist? I haven't seen this; on the other hand, I'm testing with a small library.

Baklap4:
Yup adding to the playlist while loading around 5-10k songs

Lord-Kamina:
> Yup adding to the playlist while loading around 5-10k songs

I tried again with my full library (which is still only a fraction of yours); 2650 songs and saw no 404s.

Baklap4:
Hmm i'll think i'll have to retry this to see if it's just my system

deuteragenie:
- It would probably make sense to split this PR in several smaller PRs:
1. Various improvements and refactorings
2. Introducing code and configurations that will be required by the new webserver implementation, without breaking the existing stuff
3. The new webserver implementation

In principle, point 3. should be limited to changes to webserver.cc and requesthandler.cc

- Restino states that they are a C++ 14 lib.  Any issue expected with c++17 ?
- Restino appears to offer a choice of asio implementation.  Why choose the boost one ?

Lord-Kamina:
> * It would probably make sense to split this PR in several smaller PRs:
> 
> 1. Various improvements and refactorings
> 2. Introducing code and configurations that will be required by the new webserver implementation, without breaking the existing stuff
> 3. The new webserver implementation. In principle, point 3. should be limited to changes to webserver.cc and requesthandler.cc
>

Many of the smaller things already existed as unmerged PRs; I'm aiming to either get them merged or scrapped and adjust accordingly. I'll also probably try to tidy the commit history in this PR up a bit before actually merging it.
 
> * Restino states that they are a C++ 14 lib.  Any issue expected with c++17 ?

I don't think so, and per https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects this shouldn't really be a concern in most cases.

> * Restino appears to offer a choice of asio implementation.  Why choose the boost one ?

I think the main reason was chiefly because standalone asio was not as well maintained as boost in package managers.

game/CMakeLists.txt
```diff
@@ -67,22 +67,19 @@ set(SOURCES ${SOURCE_FILES} ${HEADER_FILES} ${RESOURCE_FILES})
...
 if (USE_BOOST_FS)
-    find_package(Boost 1.55 REQUIRED COMPONENTS filesystem)
-
+    find_package(Boost ${NEEDED_BOOST_VERSION} REQUIRED COMPONENTS filesystem)
     target_link_libraries(fs PUBLIC Boost::filesystem)
     target_compile_definitions(fs PUBLIC -DUSE_BOOST_FS)
 else()
     find_package(Filesystem REQUIRED)
     target_link_libraries(fs PRIVATE std::filesystem)
-endif()
```
Baklap4, line 85: This shouldn't have been deleted (`cmake ..` error)

Baklap4 (changes requested):
The current state of this pr is pretty much broken:
1. Game/CMakeLists.txt has an error.
2. The linked ced commit doesn't work (master of ced works)
3. Getting some restinio build error

Please fix these first

game/CMakeLists.txt
Lord-Kamina: Must have been a botched merge from the rebase on master.

game/songs.cc
```diff
@@ -82,39 +79,27 @@ void Songs::reload_internal() {
...
-    } else {
-    	std::clog << "songs/info: Could not open songs meta cache file " << songsMetaFile.string() << std::endl;
-    	return;
-    }
+	nlohmann::json jsonRoot = nlohmann::json::array();
+	if (!fs::exists(songsMetaFile)) {
+		std::clog << "songs/notice: song cache not found..." << std::endl;
+		return;
+	}
+	jsonRoot = readJSON(songsMetaFile);
```
Baklap4, line 89: catch runtime_exception here else it crashes the game i think

game/songs.cc
```diff
@@ -129,134 +114,126 @@ void Songs::LoadCache() {
...
-
+	std::ofstream outFile(cacheDir.string());
 	try {
-		std::stringstream stream;
-		jsonRoot.serialize(stream);
-		std::ofstream outFile(cacheDir.string());
-    	outFile << stream.rdbuf();
-    	outFile.close();
+    	outFile << jsonRoot.dump();
+    	outFile.close(); // std::ofstream should clean-up after itself but being careful never hurt anybody.
```
Baklap4, line 228: use try-catch-finally mechanism so you can close the file in the finally block.

game/songs.cc
```diff
@@ -504,44 +482,45 @@ void Songs::sortChange(int diff) {
...
 	if(sortOrder < 0) {
-		m_order = 0;
+		sortNum(webServer) = 0;
 	} else if(sortOrder <= 6) {
-		m_order = sortOrder;
+		sortNum(webServer) = sortOrder;
 	} else {
-		m_order = 0;
+		sortNum(webServer) = 0;
 	}
```
Baklap4, line 496:
```
if(sortOrder < 0 || sortOrder > 6) sortNum = 0
else{sortNum = sortorder}

File: game/songs.cc [hunk 9]
```diff
@@ -504,44 +482,45 @@ void Songs::sortChange(int diff) {
...
-		  case 5: std::sort(m_filtered.begin(), m_filtered.end(), customComparator(&Song::path)); break;
-		  case 6: std::sort(m_filtered.begin(), m_filtered.end(), customComparator(&Song::language)); break;
+		switch (sortNum(webServer)) {
+		  case 0: std::stable_sort(vector.begin(), vector.end(), customComparator(&Song::randomIdx)); break;
+		  case 1: std::sort(vector.begin(), vector.end(), customComparator(&Song::collateByTitle)); break;
+		  case 2: std::sort(vector.begin(), vector.end(), customComparator(&Song::collateByArtist)); break;
+		  case 3: std::sort(vector.begin(), vector.end(), customComparator(&Song::edition)); break;
+		  case 4: std::sort(vector.begin(), vector.end(), customComparator(&Song::genre)); break;
+		  case 5: std::sort(vector.begin(), vector.end(), customComparator(&Song::path)); break;
+		  case 6: std::sort(vector.begin(), vector.end(), customComparator(&Song::language)); break;
```
Baklap4, line 523: It's the same in both bases except the list is reversed. Isn't there something smart to fix this and just have one switch statement instead of 2?

we could make a list copy which is already in the correct format. and then use it in the switch

game/webserver.cc
```diff
@@ -1,84 +1,101 @@
...
 		}
-
-		std::clog << "webserver/error: Couldn't start webserver tried 3 times using normal port and 3 times using fallback port. Stopping webserver." << std::endl;
-		Game::getSingletonPtr()->notificationFromWebserver("Couldn't start webserver.");							
-		if(m_server) {
-			m_server->close().wait();
-		}
+		message += std::string("using the main and fallback ports; won't try again.");
+		std::clog << message << std::endl;
+		Game::getSingletonPtr()->notificationFromWebserver(message);
```
Baklap4, line 25: indents

Baklap4 (changes requested):
Alright tested it for a bit:
0. Build failures on Ubuntu 21.04 with boost 1.74.0. To fix it add: `add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)` somehwere in cmakelist file. Hopefully you'll come up with a nicer version of that :)
1. The webserver config menu is out of sight because it became a new menu item. Maybe rescale these menu items?
2. On exit i'm getting a segfault
3. When running on port 80 and starting performous i'm getting Permission denied. Here should the retry-mechanism kick in. It retries but is not increasing the 'tried' times so it gets in a loop of continuous retrying instead of failing after 3 times on normal port than falling back on fallback port.
4. sometimes i'm getting performous in 'weird colors': 
image
5. ~~I can't search for language~~ I thought this was a bug, but then after digging i have all my songs in a folder per language. The game detects filters also on 'path' so a language can always be found. Would be nice to search for language though -> song.cc#206 - `strFull()`
6. I have only tested this with 4 songs as my library. Once the items from here are picked up i'll test with my big collection

Lord-Kamina:
> Alright tested it for a bit:
> 
> 0. Build failures on Ubuntu 21.04 with boost 1.74.0. To fix it add: `add_definitions(-DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT)` somehwere in cmakelist file. Hopefully you'll come up with a nicer version of that :)
> 
> 1. The webserver config menu is out of sight because it became a new menu item. Maybe rescale these menu items?
> 

Meaning you have to scroll to see it? I'm sure that can be fixed. That will change eventually in my PR to fix text rendering but that one's not nearly close to finished so maybe I can do something else in the meantime.

> 2. On exit i'm getting a segfault
> 

Can you give me a trace? I specifically tested this quite a bit because this used to be a bug in CppRest also. Maybe something got fucked up in a rebase?

> 3. When running on port 80 and starting performous i'm getting Permission denied. Here should the retry-mechanism kick in. It retries but is not increasing the 'tried' times so it gets in a loop of continuous retrying instead of failing after 3 times on normal port than falling back on fallback port.
> 

I'll try this out. I've never used port 80 so maybe I missed it.

> 4. sometimes i'm getting performous in 'weird colors': 
> 
> image
> 

That's really, really weird. And... I don't see how that could be caused by the PR? 🥴

> 5. ~~I can't search for language~~ I thought this was a bug, but then after digging i have all my songs in a folder per language. The game detects filters also on 'path' so a language can always be found. Would be nice to search for language though -> song.cc#206 - `strFull()`
> 

Sure.

> 6. I have only tested this with 4 songs as my library. Once the items from here are picked up i'll test with my big collection

Baklap4:
> Meaning you have to scroll to see it? I'm sure that can be fixed. That will change eventually in my PR to fix text rendering but that one's not nearly close to finished so maybe I can do something else in the meantime.

Yup, i think for now switching it with for example path would suffice. The actual fix comes in the text rendering thingy

> Can you give me a trace? I specifically tested this quite a bit because this used to be a bug in CppRest also. Maybe something got fucked up in a rebase?

After a one time (without gdb) it didn't happen to me anymore. Maybe other ppl should test this too

> I'll try this out. I've never used port 80 so maybe I missed it.

The thing with port 80 is you'll get an exception for a permission error (which is correct) but was the only thing i could think of to 'simulate' a port being used (= also permission error)

> That's really, really weird. And... I don't see how that could be caused by the PR? 

For now let it be, if i keep getting it i guess i'll file a bug. Maybe my gpu is toasted 🤷

Lord-Kamina:
> The thing with port 80 is you'll get an exception for a permission error (which is correct) but was the only thing i could think of to 'simulate' a port being used (= also permission error)
> 
> 
I see. I'll try it when I get home.

Tronic:
Looks like this PR is dragging badly, but to get back to the initial motivation:

> CppRestSDK is full of Microsoft ew.

Do you mean it has some actual flaws (like WINAPI crap), or just that it is made by Microsoft? If it is open source, that shouldn't matter.

Lord-Kamina:
> Looks like this PR is dragging badly, but to get back to the initial motivation:
> 
> > CppRestSDK is full of Microsoft ew.
> 
> Do you mean it has some actual flaws (like WINAPI crap), or just that it is made by Microsoft? If it is open source, that shouldn't matter.

I didn't like that it's bloated, not very performant (although @yoda-jm

PR #553, opened 2020-06-07
ConfigItem: Less hacky interface for oldValue.

Lord-Kamina:
### What does this PR do?

Provides a less-hacky interface for keeping track of a ConfigItem's previous value.

### More


Baklap4:
Looks good in general :) 
If i may add another suggestion (which might collide with logging) how about enumifying the components? I personally hate to see semi-random strings in code.

Baklap4:
I personally still stand with my previous comment, making the names available through an enumeration for example: `config.type` or something alike. 
With this we can convert the strings to typed enumerations making it more foolproof to use.

As for now this PR only changes setting (and getting) a value by method instead of direct property access which is slightly more expensive in terms of performance. But i think the method way is easier to read.

Lord-Kamina:
> I personally still stand with my previous comment, making the names available through an enumeration for example: `config.type` or something alike.
> With this we can convert the strings to typed enumerations making it more foolproof to use.
> 
> As for now this PR only changes setting (and getting) a value by method instead of direct property access which is slightly more expensive in terms of performance. But i think the method way is easier to read.

Do you mean for the values themselves? The names of the config items? Most things in our config system are derived from reading the XML schemas. Conceptually, I agree with you but considering our current systems and that we're working with c++, I'm not sure it's worth the effort. If we were using Swift, we wouldn't even be having this discussion OTOH. That said, if you want to put the time into it, go ahead.

Baklap4 (approved):
LGTM

ooshlablu:
Tested it for an hour with vocals and guitar, works fine as far as I can tell.

PR #554, opened 2020-06-08
Fix os_image for mac build in semaphore.yml

Lord-Kamina:
### What does this PR do?

Updates the os_image definition in semaphore.yml.

### More

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -13,7 +13,7 @@ blocks:
           - echo "Installing Performous dependencies..."
           - sudo apt -y update
           - sudo add-apt-repository -y ppa:litenstein/performous-bionic
-          - sudo apt -y install cmake gettext help2man clang-9 gcc-8 g++-8 libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libpng-dev libjpeg-dev libxml++2.6-dev portaudio19-dev libopencv-dev libportmidi-dev libqrencode-dev libicu-dev libglm-dev libopenblas-dev libfftw3-dev
+          - sudo apt -y install cmake gettext help2man clang-9 gcc-8 g++-8 libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libpng-dev libjpeg-dev libxml++3.0-dev portaudio19-dev libopencv-dev libportmidi-dev libqrencode-dev libicu-dev libglm-dev libopenblas-dev libfftw3-dev
           - checkout
           - git submodule update --init --recursive
       jobs:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -42,7 +42,7 @@ blocks:
       agent:
         machine:
           type: a1-standard-4
-          os_image: macos-mojave-xcode11
+          os_image: macos-xcode11
       secrets:
         - name: macos-ci-pubkey
         - name: check-deps-script
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -12,28 +12,27 @@ blocks:
         commands:
           - echo "Installing Performous dependencies..."
           - sudo apt -y update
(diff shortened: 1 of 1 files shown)


Lord-Kamina:
Oh; fuck; they're deprecating the macOS 10.14 image; and here I've been doing everything in my power to avoid upgrading to Catalina.

Lord-Kamina:
And, it's fixed; I aalso took the liberty of updating the GCC and Clang versions we use for Linux builds.

.semaphore/semaphore.yml
```diff
@@ -42,31 +41,32 @@ blocks:
...
+          - export PATH=/opt/local/bin:/opt/local/sbin:$PATH
+          - export MANPATH=/opt/local/share/man:$MANPATH
+          - export DISPLAY=:0.0
+          - cache restore macports-10_15
+          - mkdir -p /Users/semaphore/local
+          - mkdir -p /Users/semaphore/Applications
+          - sudo ln -sf /Users/semaphore/local/Applications /Users/semaphore/Applications/MacPorts
           - sudo ln -sf /Users/semaphore/local/ /opt/local
-          - test -e /Users/semaphore/local/.deps_installed || /Users/semaphore/check-install-macports.sh | cat 2>&1
+          - test -e /Users/semaphore/local/bin/port || ( curl -LJ https://www.dropbox.com/s/4c5oav6za5l4gvv/macports-10_15-install.tar.gz?dl=1 | tar x )
```
OznOg, line 61: why is macport on a dropbox instead of some "official" repository?

Lord-Kamina: It's my Dropbox hahaha, it's because it's the already built ports; to save time essentially.

OznOg:
LGTM if it does the job

.semaphore/semaphore.yml
```diff
@@ -12,28 +12,27 @@ blocks:
         commands:
           - echo "Installing Performous dependencies..."
           - sudo apt -y update
-          - sudo apt -y install cmake gettext help2man clang gcc-7 g++-7 libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libpng-dev libjpeg-dev libxml++2.6-dev portaudio19-dev libopencv-dev libportmidi-dev libqrencode-dev libicu-dev libcpprest-dev libglm-dev libopenblas-dev libfftw3-dev
+          - sudo add-apt-repository -y ppa:litenstein/performous-bionic
```
Baklap4, line 15: For what would we need this ppa?

.semaphore/semaphore.yml
Baklap4: Hmm maybe create a repo + release for this containing that tarball? This way it's listed on github, saving you space on dropbox and for everyone to see?

.semaphore/semaphore.yml
Baklap4: As mentioned on discord we need it for Boost.ASIO which is in ubuntu 18.04 at 1.65 while we at least require boost 1.66

Ubuntu 19.04 is shipped with boost 1.67

Baklap4: @performous/core-owners What do we think of using a personal more updated ppa?
I'm not quite fond of it myself but if it's necessaryit's necessary..
I'm not fond of it for the reason as it might mess up the users system since we're not OS maintainers we can't really tell the impact. And aside from this we also need to keep our documentation up to date regarding what and how to install stuff properly so you can build performous

Lord-Kamina: As I said in Discord; my issue is that despite being arguably the most popular Linux flavor, Ubuntu keeps some really outdated packages here and there and sometimes the reasons are dubious at best... for example, take LibXML++; there is no sane reason for the current cutting-edge of ubuntu still having only LibXML++2.6; and yet it does (seeing how, even if some packages would break by using 3.0; they could ship both, since they're designed to coexist).

With regards to boost, Back when 18.04 first came out, 1.65 was already almost a full year old, 1.66 was out and 1.67 came out around the same time as Ubuntu 18.04. I just don't see it as justifiable that in TWO-YEARS of their "LTS", they didn't update the goddamn boost version that was already obsolete when it first came out.

I think it's ridiculous to hold updating our code when even Windows has easily-accessible libraries that are kept more updated.

I might have missed something of course but from a bit of search into package-dependencies, the only thing that might be broken by adding the updated icu and boost packages (in addition to that ppa, they are also available on separate ppas to minimize new packages) to bionic is an old version of OpenTTD (another game)

Do keep in mind that packages in launchpad have to build before they are published, and if they won't build if they break anything important in the first place.

nieknooijens (approved):

PR #562, opened 2020-07-08
Add fixed scaling of the note graph

twollgam:
Actually the note graph is scaled dynamically depending on the notes that are displayed
during the next 8 seconds. This commit adds a new configuration to switch between the
dynamic scaling and fixed scaling. Fixed scaling uses all notes of the song so that
the note graph does not change while playing.


twollgam:
I closed this pr last year cause there were files included not belonging to the pr. Now I have rebased the branch and pushed a commit only containing the right files.

twollgam:
Does not work as expected. Do I have to create new pr?

Issue #557, opened 2020-06-17
Song/Player database ID handling needs refactoring 


Lord-Kamina:
### Do you want to request a *feature* or report a *bug*?

Not a bug per-se

### What did you do?

Stumbled into this while elaborating a fix for #556; in Players, Songs and PlayerItem class there are some really fucked up members defined as int, like `int empty()`...

Then there's the way player IDs is handled; it seems it's only int so it's possible to pass a -1 for "not found" but even that makes little sense considering the current logic has it so the first valid ID is 1; so even keeping that logic, we could use size_t and 0 for invalid.

Also semi-related, we have a bunch of plain size_t here and there; I think we should homogeneize them all to std::size_t.

### Output of `performous --version`: (_What version of Performous are you using?_)

```
(paste your output here)
```

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```
(paste your output here)
```

Baklap4:
Can you tell me the difference between `size_t` and `int`? passing `-1` is a very common practise which this probably derives from

Tronic:
`size_t` is unsigned and 64 bit (in practice, not by standard of course), int is 32 bits and signed. This mismatch causes compiler warnings about unsigned vs. signed comparisons, and could lead to very real security issues if one could somehow overflow the comparison. There's also signed `ssize_t`, very rarely used anywhere.

It is a good idea to use `size_t` for anything that is a count, and throw an exception instead of returning special values like -1 (in C++, unlike in C).

I think @Lord-Kamina is right, bools should also be bools.

Baklap4:
Ahh thanks for clarifing :) 

With this explanation i think it should be done :)

Lord-Kamina:
-1 is used to initialize the value. That bit of logic could even be maintained (for initialization at least) frankly because 0 is not even being used either but the other things are ugly.

twollgam:
I would like to offer a pull request, but how can I test the Players/PlayerItem stuff? If I start performous it does not show a player screen.

twollgam:
Ok, now I found the player screen. I miss a screen to manage (create, remove, configure) players.

Baklap4:
I think this has been resolved lately (and merged) am i right @twollgam ?

Baklap4:
Maybe this is solved already by #538 ?

PR #573, opened 2020-10-15
Song/Player database ID handling needs refactoring #557

twollgam:
The player id got its own type definition based on size_t. The
constant UndefinedPlayerId is defined as size_t(-1) so that ids can
start with 0.

Additionally wrong comments were removed und contructors are
cleaned up.


Baklap4:
@twollgam I think this can be closed in favor of #575 ? It seems that pr also contains the changes from here?

twollgam:
Yes, you are right. I close this pr.

PR #575, opened 2020-10-23
Refactoring game hh

twollgam:
### What does this PR do?

The PR refactors the declaration of class Game into an own header file game.hh, that resides in header screen.hh until now.

### Motivation

Clean up project structure.


game/player.hh
```diff
@@ -62,13 +62,20 @@ struct Player {
...
+using PlayerId = size_t;
+
 /** Static Information of a player, not
   dependent from current song.
 
   Used for Players Management.
   */
 struct PlayerItem {
-	int id; ///< unique identifier for this PlayerItem, Link to hiscore
+    static PlayerId UndefinedPlayerId;
```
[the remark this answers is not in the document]
twollgam: - const yes
- The UndefinedPlayerId is also used outside this class in the Players class so the 2 constructors did not solve that problem.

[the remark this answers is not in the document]
OznOg: sorry for being unclear, actually, I was thinking at 2 versions of addplayer in players

game/player.cc
```diff
@@ -83,3 +84,8 @@ void Player::calcRowRank() {
 		m_lineScore = 0;
 	}
 }
+
+PlayerItem::PlayerItem(PlayerId id)
```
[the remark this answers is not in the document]
twollgam: What is the benefit of putting implementation into the header? The drawback is lack of readability IHMO.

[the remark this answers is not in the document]
OznOg: Well less code. This kind of constructor is sufficiently strait forward not to pollute the header (no extra header pulled). Moreover, allows the constructor to be inlined. That said, this is just an opinion you are free to keep it like this if you really prefer.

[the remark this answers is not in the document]
twollgam: Even if you are right with this constructor, I think putting implementation in the header may lead developers to implement too much in the headers. (broken windows theory)
The inlining-thing is premature optimization (IMHO) and the modern linkers can inline the code too, even if it is written in the definition file.

So I want not to convince any one, but I want to tell the reasons, why I would keep it. :-)

game/player.hh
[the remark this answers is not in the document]
twollgam: Ok thanks, I will have a look at it.

twollgam:
@OznOg I committed changes that removes the UndefinedPlayerId constant as you suggested. Works well. You could continue the review now.

OznOg:
> @OznOg I committed changes that removes the UndefinedPlayerId constant as you suggested. Works well. You could continue the review now.

ok, fine :)

game/player.hh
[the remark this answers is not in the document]
OznOg: :+1:

Baklap4:
with the current changes this pr sadly got into conflicts with current master. Can we resolve it, review it and ship it?

twollgam:
@Baklap4 I tried to resolve the conflicts and checks have passed. Could you please take a look if this pr is in a state that it may be reviewed again.

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+        sudo apt-get -y install libpango1.0-dev
+        sudo apt-get -y install librsvg2-dev
+        sudo apt-get -y install libxml++2.6-dev 
+        sudo apt-get -y install libavcodec-dev libavformat-dev libswscale-dev libswresample-dev libjpeg-dev
+        sudo apt-get -y install portaudio19-dev 
+        sudo apt-get -y install libboost-all-dev
+        sudo apt-get -y install libglm-dev 
+        #sudo apt-get -y install libopencv-dev 
+        #sudo apt-get -y install libportmidi-dev 
+        #sudo apt-get -y install libssl-dev
```
Baklap4, line 61: All `sudo apt-get -y install` can be combined into one big statement. I'd also recommend disabling the recommended installs with `--no-install-recommends`

Also please enable all packages.

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+    # cross-platform coverage.
+    # See: https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Info
+      run: echo GITHUB_WORKSPACE=$GITHUB_WORKSPACE
+      
+    - name: Apt update
+      run: sudo apt-get update
```
Baklap4, line 27: apt-get update needs a `-y` flag to automaticly update the tree instead of waiting for user interaction.

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    - name: Install tools
+      run: sudo apt-get -y install git cmake build-essential gettext help2man openssl wget #clang-3.8
```
Baklap4, line 30: add the `--no-install-recommends`

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+ 
+    - name: Prepare sources for act (run actions local)
+      run: echo #rm -r 3rdparty/aubio 3rdparty/ced
```
Baklap4, line 33: If i'm correct this doesn't need to be done since we build within a container. However if something gets cached it's better to run `git clean -ffdx` to clear out all non-repo files.

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+
+    - name: Checkout 
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Prepare Submodules
+      run : git submodule update --init --recursive
```
Baklap4, line 41: This statement shouldn't be needed anymore with the checkout action.

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+        #sudo apt-get -y install libportmidi-dev 
+        #sudo apt-get -y install libssl-dev 
+
+    - name: Create Build Environment
+      # Some projects don't allow in-source building, so create a separate build directory
+      # We'll use this as our working directory for all subsequent commands
+      #run: cmake -E make_directory ${{runner.workspace}}/build
+      run: |
+        mkdir -p build
+        #cd build
```
Baklap4, line 69: Please remove unused commands

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
...
+      shell: bash
+      # Execute the build.  You can specify a specific target with "--target <NAME>"
+      run: cmake --build . --config $BUILD_TYPE
+
+    - name: Test
+      working-directory: build
+      shell: bash
+      # Execute tests defined by the CMake configuration.  
+      # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
+      run: ctest -C $BUILD_TYPE
```
Baklap4, line 92: We don't have any tests as of yet sadly

.github/workflows/build_ubuntu.yml
```diff
@@ -0,0 +1,92 @@
+name: Build Ubuntu
```
Baklap4, line 1: In general i'm not sure if we would like another build system joining the club. Might be better to propose this within another PR.
Personally i'm in favor of github actions over semaphore since github actions has linux, mac AND windows support whilst semaphore only has linux and mac. Also once https://github.com/performous/performous/pull/532 is merged you can use that `.semaphore/semaphore.yml` as baseline for all of our build needs including distribution!

game/players.cc
```diff
@@ -66,23 +56,23 @@ int Players::lookup(std::string const& name) const {
...
 	PlayerItem pi;
 	pi.id = id;
 	pi.name = name;
 	pi.picture = picture;
-	pi.path = "";
-
 
-	if (pi.id == -1) pi.id = assign_id_internal();
+	if (pi.id == PlayerItem::UndefinedPlayerId) 
+        pi.id = assign_id_internal();
```
Baklap4, line 75: Please use braces or correct indent (4 spaces starting from if statement position)

Baklap4 (changes requested):
Overall looks quite clean! I still did find a something and i think the github actions deserve their own PR (but a nice addition!)

Going to ask the testers to get this tested but in the quick run i did i didnt find anything

.github/workflows/build_ubuntu.yml
twollgam: Oh, I am sorry. That merged accidentally. :-(

game/players.cc
twollgam: OK, I have checked the sources and the indent of line 74 and 75 is correct, but they are different. While 74 uses the original tab line 75 uses 8 spaces for indent. What do you suggest in this case?

Baklap4: Ahh i see, the one is a tab the other just spaces. Uhm indent the line 75 with 2 tabs instead of one :)

Sometime later we'll have an editorconfig available to force such things ;P

.github/workflows/build_ubuntu.yml
twollgam: Can I revert the 4 commits and push again?

Baklap4: It'd be better to do a rebase where you remove those 4 commits. That's alright :)

ooshlablu:
I tested this PR with a couple of songs and everything looks fine to me.

Baklap4 (approved):
LGTM

PR #576, opened 2020-10-24
Fix some coverity issues

OznOg:
Coverity reported some use of uninitialized and/or access errors, this PR fixes some of them


Baklap4 (approved):
Looks good to me, i think it is already tested thoroughly?

OznOg:
> Looks good to me, i think it is already tested thoroughly?
Well tested as usual, BTW, most of the fix are just init, would just prevent undefined behaviour

PR #580, opened 2020-12-01
Create c-cpp.yml

twollgam:
Test c++ workflow


Baklap4:
Would be nice to see this working @twollgam :)!

twollgam:
@Baklap4 This pull request was an accident, but I got a workflow for ubuntu up and running. It is not perfect yet, so there are no artifacts, packages etc. The next weeks I will try to set it up for windows, which would be a nice base for daily builds.

Issue #588, opened 2021-05-19
cmake error


Git-Forked:
[CMakeError.log](https://github.com/performous/performous/files/6505082/CMakeError.log)
[CMakeOutput.log](https://github.com/performous/performous/files/6505083/CMakeOutput.log)

Git-Forked:
Trying to build on Manjaro (based on Arch).

OznOg:
could you try the following patch and let me know if it can fix your issue?
```

diff --git a/game/CMakeLists.txt b/game/CMakeLists.txt
index 9f8159c8..68011e96 100644
--- a/game/CMakeLists.txt
+++ b/game/CMakeLists.txt
@@ -147,9 +147,8 @@ else()
 endif()
 
 if(UNIX AND NOT APPLE)
-       # Note: cannot use list APPEND here because it inserts semicolons instead of spaces
-       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
-       set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pthread")
+        find_package(Threads)
+        target_link_libraries(performous Threads::Threads)
 endif()
 
 if(WIN32)

```

OznOg:
@Git-Forked have you been able to test my patch?

Git-Forked:
Sorry, no idea how to do that.

Baklap4:
Save the mentioned code to a file let's call it `patch`, then use git to apply it: `git apply ./patch`

OznOg:
BTW I wonder if the "IF NOT APPLE" remains useful: I guess threads are required for mac also, and I expect cmake to give unified use of them.

OznOg:
@Git-Forked you may try branch OznOg:devel/fix_some_warnings pointed in mr #589 if it is ieasier for you.

Baklap4:
This should be fixed on current master because of . Can you try to give this a test @Git-Forked ?

For now i'm closing it, but if you still run into issues please reopen the issue.

PR #589, opened 2021-05-30
Devel/fix some warnings

OznOg:
### More

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -301,21 +301,11 @@ void FFmpeg::avcodec_free_context(AVCodecContext *avctx) {
 	::avcodec_free_context(&avctx);
 }
 
-struct FFmpeg::Packet: public AVPacket {
-	Packet() {
-		av_init_packet(this);
-		this->data = nullptr;
-		this->size = 0;
-	}
-
-	~Packet() { av_packet_unref(this); }
-};
-
 void FFmpeg::handleOneFrame() {
 	bool read_one = false;
 	do {
-		Packet pkt;
-		auto ret = av_read_frame(m_formatContext.get(), &pkt);
+                std::unique_ptr<AVPacket, std::function<void(AVPacket*)>> pkt(av_packet_alloc(), [] (auto *pkt) { av_packet_free(&pkt); });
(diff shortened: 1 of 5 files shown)


Baklap4:
Looking good so far! Build currently fails for mac though. Looking forward to the complete PR. As for closes Issues. You partly resolve the boost dependency #39 and _most likely_ #588 if you include that fix too and #395

Lord-Kamina:
@OznOg could you take a look  also at https://github.com/performous/performous/pull/529 and see whether there's something there that could be integrated here?

OznOg:
yes, no pb, that said this is really draft, the main puropse of the PR was to check Mac compilation (and it fails...)

game/ffmpeg.cc
```diff
@@ -301,31 +301,31 @@ void FFmpeg::avcodec_free_context(AVCodecContext *avctx) {
...
-		decodePacket(pkt);
+		if (pkt->stream_index != m_streamId) continue;
+
+                ret = avcodec_send_packet(m_codecContext.get(), pkt.get());
+                if(ret == AVERROR_EOF) {
+                        // End of file: no more data to read.
+                        throw Eof();
+                } else if(ret == AVERROR(EAGAIN)) {
+                        // no room for new data, need to get more frames out of the decoder by
+                        // calling avcodec_receive_frame()
```
Baklap4, line 324: Let's log something here or remove this else if clause

OznOg: this is done exactly not to log anything as it is "normal" behaviour to have no room left. The EAGAIN needs to be handled separately of errors

Baklap4: Alright but since we currently do nothing within this it's just an empty block which can be removed then right? As we're not throwing anything or doing anything. You could move it then to the last else if
`else if (ret < 0 && ret !=  AVERROR(EAGAIN))

OznOg: I did this to separate the flow and add specific comment to explain this behaviour, but if you think I should change it, I will.

Baklap4: hmmm i can see the reasoning for this, aight leave it like it is :)

Baklap4 (approved):
LGTM

PR #571, opened 2020-09-26
Don't hack ced include dir

OznOg:
Let the lib provide the path itself and let cmake find it itself.

Blocked By: https://github.com/performous/compact_enc_det/pull/6

PR #591, opened 2021-06-27
Fix cmake stuff

yoda-jm:
PTAL, this helps compiling Performous on Gentoo


Baklap4:
Hmmm let's see this is the 3rd cmake stuff PR we have open 😂 
What is duplicated/better

Linked PR: #571 #529

@Lord-Kamina @OznOg Please also take a look at this

OznOg:
@yoda-jm could you give some more context on what is going wrong on gentoo? I must say I don't think there is anything specific to gentoo and moreover some part seems to do the opposite of the former code (fft3 AND openblas required)

yoda-jm:
On gentoo fortran is not installed by default, so no lib blas installed.
As the error message state it, only one of the 2 library is required to make it compile (I can compile Performous without blas).
My patch is then to fix the test so it matches the error message and what is required.

Tronic:
BLAS is not just Fortran, it is also required by some C/C++ and by all Python data science stuff (because Numpy depends on it). There are several forks of it, and the one you really want is OpenBLAS because it is much faster than the others that distributions also ship.

Maybe Fortran is required for compiling the most efficient code if you don't use a binary distribution.

OznOg:
I don't really understand how you manage to build as the blas is needed by aubio build itself being needed by performous. The same applies for fft3w. How do you perform the aubio build? am I missing something?

deuteragenie:
FYI: I build on Ubuntu with the standard aubio pkg and all appear to be fine. Not sure why the sources in the tree are needed.

Lord-Kamina:
Yeah, that `OR` was intentional. Tronic is correct re: OpenBLAS also.

The reason we need the sources in the tree is because I had modified them a lot in order to build with the specific combination of options we use and also, because I don't want to use waf (although I can't remember whether the PR that builds aubio with CMake was merged)


All this said, aubio is playing an almost entirely cosmetic role. If it's causing trouble building on gentoo, I say we just make it an optional dependency.

yoda-jm:
Indeed Performous did not start (it is strange that the compilation worked without apparent warning or fatal error).
OpenBLAS can be compiled on gentoo without Fortran, I should have wrongly chosen another BLAS implementation at start.
If the the test is ok, then we should probably fix the message in order to be clear that both are required (or just do the 2 REQUIRED tests without any message at all).

Baklap4:
What's the current status of this PR? I see some new things, some conflicts and a lot of discussion lately about cmake. See also #720 

If we gotta merge this, please resolve the conflicts and let's get it to merge. Else close this PR :)

twollgam:
@Baklap4 @yoda-jm now the last comment is nearly 2 years old. I think we should close the pr. Else we should get up a gentoo inside the pipeline so that we can prove compiling on gentoo.

Baklap4:
Adding a pipeline for gentoo should help yes, for now closing the pr. If we still want things from this pr lets reopen it or make another pull request :)

PR #595, opened 2021-07-03
let cmake fetch external projects instead of using git submodules

OznOg:


OznOg:
this need cmake version >= 3.14 thus needs to update reference images to use newer versions.

deuteragenie (commented):
Instead of pulling all sources from the performous repo, wouldn't it be better to:
1. Pull the sources from the original ced repo
2. Apply a patch to make up for any differences (ie modifications done by the performous team, but not upstreamed). The "patch" could either be applied as a diff, or, more simply, by replacing the required file(s).

This would allow to only maintain the diff (or files that have been modified) in the performous repo, and get rid of everything else ced related.

OznOg:
Yes, this remains a draft;
FYI I tried doing it like that, but I have some issues:
* ced does not allow to build without UT and pulls gtest, thus the vanilla is painful as is
* includes are badly defined on target thus needs to be patched (already done in our mirror)
* upstream repo seems dead, thus what to expect from using it.

So, I'm still experimenting, but all this make me feel uncomfortable with this repo... The best would obviously to manage to do without it....

WIP...

deuteragenie:
> Yes, this remains a draft;
> FYI I tried doing it like that, but I have some issues:
> 
>     * ced does not allow to build without UT and pulls gtest, thus the vanilla is painful as is
> 
>     * includes are badly defined on target thus needs to be patched (already done in our mirror)
> 
>     * upstream repo seems dead, thus what to expect from using it.
> 
> 
> So, I'm still experimenting, but all this make me feel uncomfortable with this repo... The best would obviously to manage to do without it....
> 
> WIP...

If upstream is dead, then there is no reason to create a submodule referring to it.  Better just cut the tie and make it sime (tm).

OznOg:
> If upstream is dead, then there is no reason to create a submodule referring to it. Better just cut the tie and make it simple (tm).

that's why I proposed git subtree; maybe I should push an example to show what would be the result.

cmake/Modules/FindAubio.cmake
```diff
@@ -45,8 +45,10 @@ ExternalProject_Add(build-aubio-from-sources
         UPDATE_COMMAND ""
 	TMP_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/tmp/
 	CONFIGURE_COMMAND ""
-	BUILD_COMMAND ${CMAKE_COMMAND} -E env CFLAGS=${AUBIO_CFLAGS} ${WAF}
-                  ${ACCELERATE_COMMAND} --disable-tests --notests --verbose
+	BUILD_COMMAND ${CMAKE_COMMAND} -E env CFLAGS=${AUBIO_CFLAGS}
+                  CXX=${CMAKE_CXX_COMPILER}
+                  CC=${CMAKE_C_COMPILER}
```
Baklap4, line 50: Are CC and CXX needed? Normally they are already set to your local compiler setup 🤔

cmake/Modules/FindAubio.cmake
```diff
@@ -0,0 +1,73 @@
...
+
+if (NOT USE_SELF_BUILT_AUBIO)
+    find_package(PkgConfig REQUIRED)
+    # The version we look for is legacy, may need some fine tunning
+    pkg_check_modules(AUBIO IMPORTED_TARGET aubio>=0.4.9)
+endif()
+
+if (NOT AUBIO_FOUND)
+
+if (APPLE)
```
Baklap4, line 12: L12 to L71 needs an indent

CMakeLists.txt
```diff
@@ -85,55 +79,21 @@ else()
...
+  GIT_TAG        master
+  SOURCE_DIR ced-src
 )
 
-ExternalProject_Get_Property(build-aubio-from-sources INSTALL_DIR)
-set (AUBIO_INSTALL_DIR ${INSTALL_DIR})
-add_library(aubio STATIC IMPORTED)
-set_target_properties(aubio PROPERTIES IMPORTED_LOCATION ${AUBIO_INSTALL_DIR}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}aubio${CMAKE_STATIC_LIBRARY_SUFFIX})
-add_dependencies(aubio build-aubio-from-sources)
+FetchContent_MakeAvailable(ced-sources)
```
Baklap4, line 90: Instead of listing this here let's add it as a cmake module L82 - L90

cmake/Modules/FindAubio.cmake
OznOg: that is the point, env is not inherited in Add_extranalProject so need to pass it (same as CXXFLAGS, actually)

CMakeLists.txt
OznOg: Yes, this is still really draft, I agree this would be much cleaner (and prevent pollution of common part)

cmake/Modules/FindAubio.cmake
Baklap4: Ahhh fair enough! Kinda makes sense yea

game/3dobject.cc
```diff
@@ -31,7 +31,7 @@ struct Face {
 void Object3d::loadWavefrontObj(fs::path const& filepath, float scale) {
 	int linenumber = 0;
 	std::string row;
-	std::ifstream file(filepath.c_str(), std::ios::binary);
+	std::ifstream file(filepath.string(), std::ios::binary);
```
Tronic, line 34: Would it be possible to write this so that non-ASCII characters work on Windows? I would expect there to be a way to make fs::path and ifstream talk using wchars when running on Windows. Opening files on Windows just won't work with regular char filenames (in UTF-8 like all other OS).

OznOg: actually,that was one of my questions: is path::string() hadling the windows charset correctly? you seem to say no... so I will probably take the other version of my path where I was using path directly but need more "hack" stuff in the fs.h.
This was the minimalistic patch to compile, as I cannot test on windows, I chose KISS.
I'll push an update let me know if you think it is better.

Tronic: At best it could be using the ANSI codepage currently active on Windows (depends on locale settings) but still cannot use any characters not included in those less than 256 that the codepage can hold. The only way to open all files on Windows is to use wchar filenames, but C++ fstreams should on Windows have overloads taking wchar and I believe that the fs library should be able to talk directly with that (but I myself am not quite up to date as I do no development on Windows nowadays).

Tronic: Okay, looking at std::filesystem reference, you need `filepath.native()` instead of `filepath.string()`

OznOg: now I use path directly, is that ok?

Tronic: ... or just pass the filepath object directly and you'll get implicit conversion to wstring or string depending on platform.

Lord-Kamina:
If we're going with this approach, we'll need to make a ced package for mxe (pretty trivial to do but we gotta remember to do it)

Lord-Kamina:
By the way... I just noticed, submodules had been fixed in the unmerged webserver PR to specifically pull from the performous mirrors, here:

https://github.com/performous/performous/pull/545/commits/274a520141a9b6acaa38cf47b79931612867bcc0

Baklap4:
Build is currently failing due to a lower version of cmake (cmake_3.12.3-3~ubuntu18.04.1) being used which doesnt support `FetchContent_MakeAvailable` This is introduced in 3.14.7

```
CMake Error at CMakeLists.txt:90 (FetchContent_MakeAvailable):00:02
  Unknown CMake command "FetchContent_MakeAvailable".00:02
-- Configuring incomplete, errors occurred!00:02
See also "/home/semaphore/performous/build.gcc/CMakeFiles/CMakeOutput.log".
```

Not quite sure if we can 'update' the build machines though. Semaphore 2 is using ubuntu 1804 and Semaphore classic is using ubuntu 1604.

Ubuntu 20.04 ships with cmake 3.16.3

Baklap4:
> we'll need to make a ced package for mxe (pretty trivial to do but we gotta remember to do it)

What do you mean with this @Lord-Kamina ? With the current state we can build with mxe without making a mxe package for ced. There might be some differences in how to build though

OznOg:
> If we're going with this approach, we'll need to make a ced package for mxe (pretty trivial to do but we gotta remember to do it)

I've been able to test in the mxe Xcompil environment, and the download did no cause any problem.

Lord-Kamina:
> What do you mean with this @Lord-Kamina ? With the current state we can build with mxe without making a mxe package for ced. There might be some differences in how to build though

Because of the whole "don't download shit while building" approach.

Baklap4:
> Because of the whole "don't download shit while building" approach.

Like said before that's not being done at least not within the mxe environment. The mxe environment doesn't have ced that's correct. We build like this currently:

```
git clone https://github.com/mxe/mxe.git
cp <performous_root>/win32/mxe/settings.mk
cd mxe
make MXE_PLUGIN_DIRS=plugins/gcc10 gcc
#wait for a while
make MXE_PLUGIN_DIRS=plugins/gcc10
cd ..
mkdir build
cd build
PATH=$PWD/../mxe/usr/bin/:$PATH  i686-w64-mingw32.shared-cmake ../../performous/
PATH=$PWD/../mxe/usr/bin/:$PATH make -j8
```

The magic is within the secondlast line. From our mxe build dir we configure performous (and thus downloading and copying the files (aubio, ced) to the build dir) and then run the build itself. While building nothing gets downloaded.

It's basicly the same mxe does with all the libraries within its own `mxe/src` folder. It first downloads all the deps. Then Builds and install it.

Baklap4:
@OznOg The build seems to work at windows in regard of building aubio. The full build is still a no go though due to vcpkg being a bitch. Aubio is correctly build and or detected if already installed so that's good!

Lord-Kamina:
> > Because of the whole "don't download shit while building" approach.
> 
> Like said before that's not being done at least not within the mxe environment. The mxe environment doesn't have ced that's correct. We build like this currently:
> 
> ```
> git clone https://github.com/mxe/mxe.git
> cp <performous_root>/win32/mxe/settings.mk
> cd mxe
> make MXE_PLUGIN_DIRS=plugins/gcc10 gcc
> #wait for a while
> make MXE_PLUGIN_DIRS=plugins/gcc10
> cd ..
> mkdir build
> cd build
> PATH=$PWD/../mxe/usr/bin/:$PATH  i686-w64-mingw32.shared-cmake ../../performous/
> PATH=$PWD/../mxe/usr/bin/:$PATH make -j8
> ```
> 
> The magic is within the secondlast line. From our mxe build dir we configure performous (and thus downloading and copying the files (aubio, ced) to the build dir) and then run the build itself. While building nothing gets downloaded.
> 
> It's basicly the same mxe does with all the libraries within its own `mxe/src` folder. It first downloads all the deps. Then Builds and install it.

Sorry, I miscommunicated what I meant. MXE aims to not have anything download at any point except during the download stage of a package. OTOH, this might not be a terrible issue because at least in some environments this seems to be more a strong suggestion than an actual requirement.

Baklap4:
@OznOg If you get the pipelines up and running for this branch that'd be nice. As far as windows concerned it build correctly from this branch :)

Baklap4:
The meson build can be fixed with this wrap file called `ced.wrap`:
```
[wrap-git]
url = https://github.com/performous/compact_enc_det.git
revision = head

[provide]
dependency_names = ced
```

within the `meson.build` file add the following before the dependencies array:
```
ced = subproject('ced')
```
Within the `meson.build` file add the `ced` library as a dependency in the dependencies array.

also remove the following line: `include_directories : include_directories('3rdparty/ced'),`

For aubio (self-built) can do the same but isn't currently a showstopper if it doesn't happen.


As for cmake in our ubuntu1804 builds. I think we should build cmake ourselves and use that over the provided version.

Lord-Kamina:
I am addingna ced ‘meson.build‘ as part of my webserver branch. And I plan to have it build aubio by simply importing the CMake script

Baklap4:
> I am addingna ced meson.buildas part of my webserver branch.

@Lord-Kamina please make it in a separate pr 👍

Lord-Kamina:
I can't do it in a separate PR because my branch can't be merged if it doesn't build in meson.

OznOg:
> @OznOg If you get the pipelines up and running for this branch that'd be nice. As far as windows concerned it build correctly from this branch :)

Baklap4:
hmmm, why did you close it @OznOg ?

OznOg:
that was a mistake.

OznOg:
for the meson part I am not fluent enough with meson to be able to fix anything. Feel free to merge anything in this branch that would fix it up.

deuteragenie:
For meson, it is currently using aubio from system pkg and ced from its git repo, so we are fine.
I'll have a look at using .wraps later.

OznOg:
> For meson, it is currently using aubio from system pkg and ced from its git repo, so we are fine.
> I'll have a look at using .wraps later.

well, seems that it is not doing exactly that: it fails looking for the submodule....

Baklap4:
@OznOg you basicly have to do whats within my comment: https://github.com/performous/performous/pull/595#issuecomment-894395680

OznOg:
the relentless maintenance of 2 build chains...

OznOg:
> @OznOg you basicly have to do whats within my comment: [#595 (comment)](https://github.com/performous/performous/pull/595#issuecomment-894395680)
> 
> Instead of using `HEAD` as revision, kamina made a commit on performous/ced which you can use: https://github.com/performous/compact_enc_det/tree/meson

well must be missing something again...

Baklap4:
> well must be missing something again...

Ahh my bad, i think there needs to be a 'subproject' folder in the root and within that folder needs the ced.wrap file i think but @deuteragenie might know more

deuteragenie:
Yes, in a subproject folder.  Or alternatively, use the method in the OSX PR (get ird of 3rd party and pull ced from its repo with git) and I'll add the wrap version later:
https://github.com/performous/performous/pull/627/files

Baklap4:
>  Or alternatively, use the method in the OSX PR

That won't go up as the submodule is deleted here (as it should be within this pr)

deuteragenie:
> > Or alternatively, use the method in the OSX PR
> 
> That won't go up as the submodule is deleted here (as it should be within this pr)

Well, this is also what is done in that PR:

# Clean up the 3rdparty folder (aubio is installed with brew) and pull ced from its repository

          - rm -rf 3rdparty
          - git clone --depth 1 https://github.com/google/compact_enc_det.git 3rdparty/ced
          # Checkout should fail if (ever) google ced is updated, providing an opportunity to review the changes
          - git -C 3rdparty/ced checkout 37529e6

Baklap4 (approved):
LGTM thanks for this pr @OznOg! Seems there are some conflicts with this vs master though

OznOg:
I just rebased the branch and remove useless commits

Baklap4 (approved):
LGTM

PR #598, opened 2021-07-04
Remove uneeded constructor

deuteragenie:
Remove uneeded constructor in controllers.cc, which also suppresses a warning.
The default implicit constructor should work well enough.
Tested on Ubuntu.

### DIFF
--- a/game/controllers.cc
+++ b/game/controllers.cc
@@ -89,7 +89,6 @@ struct ButtonMap {
 	Button map;  // Generic action
 	Button negative, positive;  // Half-axis movement
 	Button up, down, left, right;  // Hat direction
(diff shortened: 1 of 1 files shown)


OznOg:
I think this was already fixed in https://github.com/performous/performous/commit/3a01558bdf4adc96230098157b8e36907cace9ae wasn't it?

deuteragenie:
> I think this was already fixed in [3a01558](https://github.com/performous/performous/commit/3a01558bdf4adc96230098157b8e36907cace9ae) wasn't it?

Ah, yes.  It just got committed while I was looking at it.  Thx, I will close this.

PR #599, opened 2021-07-05
Address warning "ISO C++ forbids variable length array infoLog"

deuteragenie:
- Address the warning in title
- Reorder slightly the logic to make the normal case (ie no infolog) appear first, rather than the special radeon case
- Log compilation of shaders at info level as this is quite an important event of the app lifecycle

### DIFF
--- a/game/glshader.cc
+++ b/game/glshader.cc
@@ -25,22 +25,27 @@ namespace {
 }
 
 /// Dumps Shader/Program InfoLog
-void Shader::dumpInfoLog(GLuint id) {
+void Shader::dumpInfoLog(GLuint id) { 
 	int maxLength;
 
 	if (glIsShader(id)) glGetShaderiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 	else glGetProgramiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 
-	char infoLog[maxLength];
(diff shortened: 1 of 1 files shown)


game/glshader.cc
```diff
@@ -25,22 +25,27 @@ namespace {
...
 /// Dumps Shader/Program InfoLog
-void Shader::dumpInfoLog(GLuint id) {
+void Shader::dumpInfoLog(GLuint id) { 
 	int maxLength;
 
 	if (glIsShader(id)) glGetShaderiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 	else glGetProgramiv(id, GL_INFO_LOG_LENGTH, &maxLength);
 
-	char infoLog[maxLength];
+	char *infoLog = new char[maxLength];
```
Tronic, line 34: Performous AFAIK never uses the `new` operation and certainly never the `new[]` which is just plain evil. Don't use manual memory management. Instead, in this case, use

```c++
std::vector<char> infoLog(maxLength);
// and where you need the data pointer
infoLog.data()
```

deuteragenie: Will change it, also based on this:
https://www.khronos.org/opengl/wiki/Shader_Compilation

deuteragenie:
Closed in favor of PR 600.

PR #605, opened 2021-07-12
Ignore all build folders in .gitignore

deuteragenie:
I suppose we don't want any build* folder ending up in git, so ignore any folder starting with build instead of only build32 and build64.  This is consistent with semaphore which uses build.gcc and build.clang for example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
-build
-build32
-build64
+build*
 CMakeLists.txt.user*
 *~
 *.mo
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-build*
+/[bB]uild*/
 CMakeLists.txt.user*
 *~
 *.mo


Lord-Kamina (approved):
Makes sense to me, at least.

.gitignore
```diff
@@ -1,6 +1,4 @@
-build
-build32
-build64
+build*
```
Baklap4, line 1: i'd rather use:
`[bB]uild*`

Also on visual studio the default is an `out` folder.

Baklap4 (changes requested):

.gitignore
Tronic: I'd suggest `/[bB]uild*/` with the first slash making it match only repository root and the final slash making it match folders only.

deuteragenie: Done.  I guess this can be merged into master now.

PR #606, opened 2021-07-12
Fix fs with boost 1 74

OznOg:
Fix compilation when using boost FS >= 1.74


Baklap4 (approved):
LGTM

Lord-Kamina:
@Tronic could you check whether this fixes the error you found previously with `performous_app_build.sh`?

OznOg:
I'm affraid, there is something wrong still; Just figured out that create_directory has inverted parameter with copy_directory, thus the code may be wrong as I would need to invert :
`                        else { create_directory(path, destinationDir / relativePathStr); }` in fc.cpp
I'll need to make some more checks,..
@Tronic You may check compilation anyway, as you problem is plain compilation pb AFAII
@Baklap4 please wait for my check to complete before merging

Baklap4 (approved):
Confirmed by other mac user (@ooshlablu) that this resolves this building issue.

Baklap4:
@OznOg rebase and merge when you're up for it :+1:

PR #608, opened 2021-07-13
Add git describe to version

deuteragenie:
Adds the git describe output, containing the git commit id etc. to the project version.
When started, the app then logs the git reference:

```
...
core/notice: Performous 2.x (git: 1.0-1128-gcfb61b90) starting...
  Internationalization: Enabled
  MIDI Hardware I/O:    Enabled
... 
```

This works when building locally, but on semaphore there appears to be an issue with reading from stdout (maybe).  
In this case, the git describe output is ignored (ie only Performous 2.x is displayed).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 #Project definition
 project('performous',
         'cpp',
-        version: '2.x',
+        version: '1.x',
         default_options : ['cpp_std=c++17','buildtype=debugoptimized'],
         meson_version: '>=0.53.0',
         license : 'GPL-2.0-or-later')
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -1,7 +1,7 @@
 #Project definition
 project('performous',
         'cpp',
-        version: '1.x',
+        version: '2.x',
```
Baklap4, line 4: For now keep this at 1.x

Baklap4 (changes requested):

meson.build
deuteragenie: Reverted

Baklap4 (approved):
LGTM But ci fails due to packagecloud. Can't merge until that is resolved: https://github.com/git-lfs/git-lfs/issues/4548

PR #611, opened 2021-07-13
Set MacOS variable properly

ooshlablu:
otherwise the input is nothing and causes build issues on later versions

### What does this PR do?

Fix a bug in CMakeLists.txt that causes build issues in later MacOS versions.  The current version results in the following error, because the output variable contains no output:
```
-- macOS deployment target not set; checking host... 
CMake Error at CMakeLists.txt:51 (if):
  if given arguments:

    "VERSION_GREATER_EQUAL" "10.15"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!
```


CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if (APPLE)
 			set(MACOS_SUPPORT_FILESYSTEM TRUE)
 		endif()
 	else()
-		execute_process(COMMAND "/usr/bin/sw_vers | grep ProductVersion | sed 's|ProductVersion: ||'" OUTPUT_VARIABLE HOST_MACOS_VERSION)
+		execute_process(COMMAND /usr/bin/sw_vers "-productVersion" OUTPUT_VARIABLE HOST_MACOS_VERSION)
```
Baklap4, line 48: Remove the trailing spaces/tabs

Baklap4 (changes requested):

CMakeLists.txt
ooshlablu: Whoops. All set

Baklap4 (approved):
LGTM

PR #612, opened 2021-07-13
Fixes for performous-app-build.sh

Lord-Kamina:
This was broken by the previous PR expanding build folders.

### More


.gitignore
```diff
@@ -1,4 +1,5 @@
 /[bB]uild*/
+/osx-utils/[bB]uild*/
```
Baklap4, line 2: hmm maybe instead of listing 'build' folders everywhere in the tree, let's get rid of the `/` in front of line 1.  This should stop every folder called  `build*` or `Build*` from being listed anywhere in the directory

`[bB]uild*/`

Baklap4 (changes requested):

.gitignore
Lord-Kamina: Other options would be to edit `performous_app_build.sh` so it uses `build` in the project root. Does anybody feel strongly about either option? @Tronic do you care either way?

Tronic: Making the build folder at project root would be nicer IMO.

ooshlablu:
I was able to build `Performous.app` using dependencies installed via homebrew with this PR, which I was not able to do before. When running the app though, I can't get the microphones to track, singing doesn't show up during the game and none of the notes can be hit. I used several microphones to test this out: the built-in mic on the mac, a USB mic that I use on performous on ubuntu 20.04 all the time, and a real mic connected to a USB preamp. Interestingly enough, performous can output through the USB preamp without any issues.
I don't think that anything in this PR would have caused the issue I am seeing.

Lord-Kamina:
@Tronic so the idea now would be to use the build directory in the system root, but still keep `osx-utils` as the place where the script and all its output lives.

I've added a couple new options, `--no-clean|-k` to enable rebuilds without purging the bundle first, and also support for a new `PERFORMOUS_SOURCE` variable, defaulted to `${CURRDIR}/..`

Can't get meson working still (due to cpprest and portmidi not liking package-config, although the first can be fixed with a small patch; not sure if it might break other platforms) but at least @ooshlablu was able to get a working .app using Homebrew from this PR. Please confirm when you have some time.

Tronic:
Now it builds, bundles and the bundle even runs. Thanks for your work!

Could you make selection of PREFIX=/usr/local automatic when /opt/local does not exist? I think nearly everybody uses Homebrew now, so defaulting to MacPorts prefix doesn't work too well.

Lord-Kamina:
> Now it builds, bundles and the bundle even runs. Thanks for your work!
> 
> Could you make selection of PREFIX=/usr/local automatic when /opt/local does not exist? I think nearly everybody uses Homebrew now, so defaulting to MacPorts prefix doesn't work too well.

Glad it's finally working for you.

I just added a bit of logic to auto-detect either MacPorts or Homebrew (and let the user pick if somehow both are installed).

osx-utils/performous-app-build.sh
```diff
@@ -30,6 +39,39 @@ function asksure {
...
+				then
+					PREFIXDIR="${brew_prefix}"
+				else
+					PREFIXDIR="${port_prefix}"
+			fi
+	fi
+}
+
+test -z ${PREFIXDIR} && detectPkgManager # Look for both Homebrew and Macports and use whichever is found. If for some reason, both are available, let the user choose. Alternatively, define this while invoking the script to bypass this behavior.
+test -z ${DEPLOYMENT_TARGET} && DEPLOYMENT_TARGET="10.14" # Change this if you want to target a different version of macOS.
```
Baklap4, line 69: instead of a static string i think it'd be better to use `sw_vers` for this to make it compatible with all systems instead of changing stuff in the directory. It also prevents useless commits/changes.

Baklap4 (changes requested):
Looks alright, but i have one comment to use `sw_vers` instead of a string which changes over time and across developers.
I can't test this as i do not own a mac

osx-utils/performous-app-build.sh
Lord-Kamina: Fixed. I also squashed everything because the commit list was getting too bloated.

Baklap4 (approved):
LGTM

PR #501, opened 2020-03-21
consolidate submodules

Baklap4:
# What does this PR do?

Makes the submodules be listed the same and moved to the 3rdparty folder.
Submodules are now pulled/updated the way you cloned the repo.
So if you cloned the repo with SSH the submodules will also be pulled by SSH
and if you cloned the repo with https the submodules will also be pulled by https

### Closes Issue(s)

Closes #461 
### Motivation

The lack of structure was bothering me.

PR #613, opened 2021-07-13
Fix submodule paths.

Lord-Kamina:
This should fix the issues with submodules on forks.

### More


Baklap4 (approved):
LGTM

OznOg:
I'm a bit late, but as far as I remember, ../ were added on purpose to allow testing modification in submodules, I'm affraid we get back in a previous bug...

Baklap4:
I think you're right @OznOg 
Hmmm the changes `../` were added as part of #501 

The problem what is being fixed here is people who forked the repo and didn't fork the submodules couldn't build.

Now both parties can build and soon (#595) submodules won't be a thing anymore
The ones who checked out using ssh will have to provide a username/password for the submodule to clone though

OznOg:
you are right, let's see with #595

deuteragenie:
Yes, let us see how it goes with aubio from system pkgs and ced from repo.  
On the test meson builds, it already works without any need for submodules.

PR #614, opened 2021-07-13
Boost.Predefs: remove dead code.

Lord-Kamina:
Drop support for Boost < 1.55. This was only really needed back when we still supported ubuntu 14.04.

### More


Baklap4 (approved):
Looking good. Let's get rid of some old code. 1.55 is good enough for now. But i think if possible we should target 1.66 or higher

Lord-Kamina:
> Looking good. Let's get rid of some old code. 1.55 is good enough for now. But i think if possible we should target 1.66 or higher

In practice I don't think we really support 1.55 any longer either but we can figure that out properly in another PR. As is, this is causing a bug with cmake 3.15+ falsely detecting boost as < 1.54.

PR #616, opened 2021-07-14
Uppercase cmake commands to lowercase

Baklap4:
### What does this PR do?

Consistent usage of the `set` cmake command. Upper and lowercase were mixed. Now all lowercase as it should be according to the docs.
Also according to cmake's own standard i've used a script to make everything lowercase which wasn't. Please see commit comment: https://gitlab.kitware.com/cmake/cmake/-/commit/77543bd

### More


ooshlablu:
Built, installed, and ran. Don't see any issues after trying out a song.

Lord-Kamina (approved):
LGTM

PR #617, opened 2021-07-14
Speed up builds by limiting the max execution time

Baklap4:
### What does this PR do?

Added max time of 15 minutes for normal builds and 10 minutes for speedy meson builds
A pipeline has a timeout of 30 minutes.

Example
Pipeline runs (20 minutes total already) -> all succeeds but macos i still busy. Macos will kill itself after 15 minutes of running within that 30 minute total pipeline time.

### Motivation

Builds should be fast. On average our build time is 6-8 minutes for linux and osx. the meson builds average around 5-6 minutes

### More


Baklap4:
seems the timelimits work:
image

deuteragenie:
I have an incoming brew based build for OSX which takes a bit less than 30 minutes. 
Will that fit ?

Baklap4:
> Will that fit ?

Nope. Might be worth checking into caching the folder homebrew uses. Not sure how big that is.

deuteragenie:
> > Will that fit ?
> 
> Nope. Might be worth checking into caching the folder homebrew uses. Not sure how big that is.
> For now this will be an improvement once homebrew is ready we can change it then

I agree to go forward with what you have now.  
We will see when I have the homebrew polished indeed. Go !

Lord-Kamina (changes requested):
According to this https://docs.semaphoreci.com/reference/pipeline-yaml-reference/#an-example-with-multiple-execution_time_limit

We might be able to separate the macos cache on a separate spec and maybe put the timeout there?

Baklap4:
> We might be able to separate the macos cache on a separate spec and maybe put the timeout there?

Like i mentioned before let's do that some other time when the meson build is merged.
For now 15 minutes is long enough.

Lord-Kamina:
Fair enough.

Lord-Kamina (approved):

PR #619, opened 2021-07-15
Remove three files that appear to be outdated / unmaintained / uneeded

deuteragenie:
- The docker files are seriously outdated, and it would be easier/cleaner for anybody to start from a blank slate if docker was to be used.
- The compiling instructions are maintained in the wiki.


Baklap4 (approved):
Can be re-added when needed again.

PR #620, opened 2021-07-15
Address 2 clang warnings

deuteragenie:
Remove two warnings returned by clang.

The one related to pitch detection needs careful attention, because it is maybe performance critical.
I tested it briefly and it seems not to make any difference.


Tronic:
> The one related to pitch detection needs careful attention, because it is maybe performance critical.

Heap allocations or zero-initialisation aren't that expensive, makes no difference here. Most other programming languages allocate, initialise and free all temporary objects that they ever use on the heap, and still do just fine :)

Baklap4:
@twollgam feel free to give this pr a test for those warning fixes 👍

twollgam:
I think the fixi is not correct and has to look like this in screen_intro.hh:

```
@@ -32,7 +32,9 @@ class ScreenIntro : public Screen {
 	std::unique_ptr<ThemeIntro> theme;
 	Menu m_menu;
 	bool m_first;
+#ifdef USE_WEBSERVER
 	bool m_drawNotice = false;
+#endif
 	AnimValue m_selAnim;
 	AnimValue m_submenuAnim;
 	AnimValue m_webserverNoticeTimeout;
```
I do not know where USE_WEBSERVER is configured, but simply removing m_drawNotice failed to compile in my local repository.

Baklap4:
`USE_WEBSERVER` is a cmake flag to enable the webserver. If cpprestsdk, libcrypto and libssl are found it's enabled per default unless explicitly disabled.

I think the PR just moves it from a place to somewhere down + includes the ifdef as stated by yourself. That should be alright right?

twollgam:
Hm, I am confused. Why I did not see that? Sorry ...

Baklap4 (approved):
LGTM

PR #621, opened 2021-07-16
Fix macOS cache on semaphore build

Lord-Kamina:
### More


Baklap4:
I think this is still a draft branch? (since it failed after it hit the 15 minute timer)

Lord-Kamina:
> I think this is still a draft branch? (since it failed after it hit the 15 minute timer)

Yeah, it took me quite a while because I had missed something. I think we're good now; though.

Baklap4:
I'm currently rerunning the pipeline for a few times. As of your initial it came down to 4 minutes for osx build which is nice!

Baklap4 (approved):
LGTM ci pipeline is about 4 minutes for osx which is pretty good compared to 15+ minutes. Let's see if it holds up during other PR's.

PR #625, opened 2021-07-18
Refactor platform names

Lord-Kamina:
This is a minor refactor; the sole purpose of this is to avoid macro conflicts.

### More


game/platform.cc
```diff
@@ -1,17 +1,17 @@
...
 
-Platform::platforms Platform::currentOS() {
-if (BOOST_OS_WINDOWS != 0) { return windows; }
-else if (BOOST_OS_LINUX != 0) { return linux; }
-else if (BOOST_OS_MACOS != 0) { return macos; }
-else if (BOOST_OS_BSD != 0) { return bsd; }
-else if (BOOST_OS_SOLARIS != 0) { return solaris; }
-else if (BOOST_OS_UNIX != 0) { return unix; }
+Platform::HostOS Platform::currentOS() {
+if (BOOST_OS_WINDOWS != 0) { return PERFORMOUS_WIN; }
```
Baklap4, line 5: Hmmm where are these `BOOST_OS_X` coming from? i guess since we link with boost? Is there a way in `std` to correctly determine OS?

Aside from that this is pretty much a rename so that's alright.

game/platform.hh
```diff
@@ -21,9 +21,9 @@ using STAT = struct stat;
 #endif
 
 struct Platform {
-enum platforms { windows, linux, macos, bsd, solaris, unix };
+enum HostOS { PERFORMOUS_WIN, PERFORMOUS_LINUX, PERFORMOUS_MAC, PERFORMOUS_BSD, PERFORMOUS_SOLARIS, PERFORMOUS_UNIX };
```
Baklap4, line 24: Within c++ there's two ways to declare an enum:
1. Plain old enum like shown
2. an Enum class 
The latter is used more and should be preferred because they cause fewer surprises that could potentially lead to bugs. It also makes refactoring way easier.
And as third we already use enum class with the a lot of our enums. Granted not yet all use it.

An example copied over from [stackoverflow](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum) on the usage:
```
enum Color { red, green, blue };                    // plain enum 
enum Card { red_card, green_card, yellow_card };    // another plain enum 
enum class Animal { dog, deer, cat, bird, human };  // enum class
enum class Mammal { kangaroo, deer, human };        // another enum class

void fun() {

    // examples of bad use of plain enums:
    Color color = Color::red;
    Card card = Card::green_card;

    int num = color;    // no problem

    if (color == Card::red_card) // no problem (bad)
        cout << "bad" << endl;

    if (card == Color::green)   // no problem (bad)
        cout << "bad" << endl;

    // examples of good use of enum classes (safe)
    Animal a = Animal::deer;
    Mammal m = Mammal::deer;

    int num2 = a;   // error
    if (m == a)         // error (good)
        cout << "bad" << endl;

    if (a == Mammal::deer) // error (good)
        cout << "bad" << endl;

}
```

If we actually wanna keep on using the plain enums we should be using typed enums: `enum Animal: unsigned int {dog, deer, cat, bird}` introduced in c++11

Baklap4 (changes requested):

game/platform.hh
Tronic: @Baklap4 Yes, this is better and avoids the ugly names too.

Baklap4: Just checking: with 'this' you mean the `enum class` variant right?

Tronic: Yes, enum class and short lower-case labels.

game/platform.cc
Lord-Kamina: There isn't, or wasn't last I checked.
These are part of the Boost.predef

game/platform.hh
Lord-Kamina: This was taken from a larger PR. I originally had used enum class (albeit with similarly long names) and @Tronic said they were too long.

I went for long, specific names for one specific circumstance (although there could be others), using gcc with `-std=gnu[11|14|17] treats `linux` as a macro and all sort of shit breaks.

Tronic: Right, cannot avoid macro expansion with namespacing. Oh, well. Anyway, the names can be more reasonable still. Cannot change all naming to use very unique names just to avoid macro collisions.

For instance, `<windows.h>` defines a macro named `max` which obviously spoils your efforts at using `std::max(...)`. Clean up the excess macros when you can (`-D WIN32_LEAN_AND_MEAN --std=c++17`).

Lord-Kamina: Yeah; in general MXE at least includes most things using `-D WIN32_LEAN_AND_MEAN`and many libraries also have also try to get around it by specifically checking for `max` and `min` which are particularly troublesome.

How about keeping it unscoped and changing to `OS_WIN, OS_MAC`and so on? I'm using caps mostly because enums elsewhere in the code are also using caps. I should probably also change `HostOS`to `hostOS`, though.

Baklap4 (approved):
LGTM

PR #626, opened 2021-07-18
Feature/drop waf

Lord-Kamina:
### What does this PR do?

Makes aubio build straight from CMake. Also, changes aubio includes to the hacky `aubio.h` to `aubio/aubio.h` as it should be.

This might need some minor retooling of the meson build.

### More

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,38 +97,39 @@ else()
 	set(AUBIO_CFLAGS "${CMAKE_C_FLAGS}")
 endif()
 
-find_package(BLAS REQUIRED MODULE)
-find_package(FFTW3 COMPONENTS single REQUIRED MODULE)
+#include(ExternalProject)
+#ExternalProject_Add(build-aubio-from-sources
+#	SOURCE_DIR ${CMAKE_SOURCE_DIR}/3rdparty/aubio/
+#        BINARY_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-build/
+#        INSTALL_DIR ${CMAKE_BINARY_DIR}/3rdparty/aubio-install/
+#        UPDATE_COMMAND ""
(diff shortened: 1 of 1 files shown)


Baklap4 (changes requested):
LGTM for normal linux builds
But please remove the commented code, we're using git here not some kind of dropbox ;)

ooshlablu:
I tried the meson build on ubuntu 18.04 and 20.04 earlier and both failed, but at different places and with different errors:

18.04:
```
The Meson build system
Version: 0.45.1
Source dir: /root/performous/performous
Build dir: /root/performous/performous/build
Build type: native build

meson.build:10:7: ERROR: lexer
summary({'Use Open CV': get_option('usewebcam'),
       ^
```

20.04:
```
Run-time dependency sdl2 found: YES 2.0.10
Run-time dependency aubio found: NO (tried pkgconfig and cmake)

meson.build:136:0: ERROR: Dependency "aubio" not found, tried pkgconfig and cmake
```

ooshlablu:
After installing `aubio` from apt in ubuntu 20.04, the meson build... builds and installs inside a docker container. I have not play tested a meson build on ubuntu 20.04 yet, but the cmake errors are gone.

Installing `aubio` from apt in ubuntu 18.04 doesn't make a difference. Same awful looking `lexer` error :-)

deuteragenie:
You need to install meson using pip.  Have a look at the semaphore.yml scripts.

ooshlablu:
Installing meson and ninja from pip did the trick

Baklap4:
@OznOg / @Lord-Kamina  The build seems to work at windows in regard of building aubio. The full build is still a no go though due to vcpkg being a bitch. Aubio is correctly build and or detected if already installed so that's good!

OznOg:
This can be dropped, I merged it in #595

PR #630, opened 2021-07-25
Remove boost regex

OznOg:
This PR remove completely boost regex from code. Was kept for redhat workaround on fedora 32, but it is now fixed.


game/players.cc
```diff
@@ -131,7 +131,7 @@ void Players::filter_internal() {
 			});
 		}
 // 		for (auto const& p: m_players) {
-// 			if (regex_search(p.name, regex(m_filter, regex_constants::icase))) filtered.push_back(p);
+// 			if (regex_search(p.name, regex(m_filter, std::regex_constants::icase))) filtered.push_back(p);
 // 		}
```
Baklap4, line 135: Hmmm why did we comment this code out? Let's remove it if we don't intend to use player filtering.

game/songparser.cc
```diff
@@ -145,7 +145,7 @@ void SongParser::guessFiles () {
 			file.clear();
 		}
 		if (file.empty()) { missing = true; }
-		regexps.emplace_back (p.second, regex_constants::icase);
+		regexps.emplace_back (p.second, std::regex_constants::icase);
```
Baklap4, line 148: remove space between method and braces

CMakeLists.txt
```diff
@@ -30,10 +30,6 @@ if(CMAKE_BUILD_TYPE STREQUAL "Release")
 	add_definitions("-DNDEBUG")
 endif()
 
-if (USE_BOOST_REGEX)
-        add_definitions(-DUSE_BOOST_REGEX)
-endif()
-
 option(USE_BOOST_FS "Whether to use boost filsystem instead of c++17" FALSE)
```
Baklap4, line 33: This option may be deleted too

Baklap4 (changes requested):
Looks good overal. Is this also an issue in lower versions of fedora? If yes do we still want to support that or get rid of that support?

OznOg:
> Looks good overal. Is this also an issue in lower versions of fedora? If yes do we still want to support that or get rid of that support?

fedora 31 is not supported anymore.

CMakeLists.txt
OznOg: there never was such an option (forgot to add it...)

game/players.cc
OznOg: done

game/songparser.cc
OznOg: done

Baklap4 (approved):
LGTM

PR #636, opened 2021-08-05
Added support for libxml++-5.0

Baklap4:
### What does this PR do?

As the title says, adds support for libxml++-5.0.

### Closes Issue(s)

None

### Motivation

vcpkg (windows) and arch come with libxml++-5.0. Within vcpkg i couldn't get 2.6 to work and 3.0 they didn't have so i've got to work around this using libxml++-5.0

### Additional Notes

People on ubuntu/debian/fedora/mac probably will have to download/compile libxml++-5 if they wanted to use it as it's pretty much not shipped with any OS per default.

### More additional Notes

Maybe the cmake module can be 'smarter' defined. If so please let me know i'm not a hero with cmake modules


game/libxml++-impl.hh
```diff
@@ -33,5 +33,19 @@ namespace xmlpp {
...
+		return element->add_child_element(name);
+	}
+
+	static inline const TextNode* get_first_child_text(const Element& element) {
+		return element.get_first_child_text();
+	}
+
+	static inline void set_first_child_text(Element* element, const std::string& content) {
+		return element->set_first_child_text(content);
+	}
```
Tronic, line 49: Could we avoid these wrappers by dropping support for 2.6 and then using the element accessors directly? They seem identical for 3.0 and 5.0, except for the string type.

image

Baklap4: 2.6 is sadly still in main debian repo's without 'easy' access to 3.0 :(

deuteragenie: 2.6 is the default on Ubuntu 18.04 so it is difficult to remove it without breaking stuff
It seems that the versioning of xmlpp is not exactly what it seems to be... meaning 5.0 is not necessarily "better" than 2.6 or 3.0.

Baklap4: Correct. 5.0 is just libxml++3 but without depending on glibmm
image

Baklap4: @Tronic I've asked around and came up with the new result

Lord-Kamina:
Tested on mac, works fine. LGTM

Lord-Kamina (approved):

PR #637, opened 2021-08-05
Update and cleaning of cmake modules

Baklap4:
### What does this PR do?

- Removed all unused cmake modules + fixed dependencies in apt-get statement in ci.
- Updated 'default' cmake modules to their latest version upstream.
- Added support for glibmm-2.68 and sigc++-3.0
- Fixed a typo in FindPThread module

### Closes Issue(s)

None

### Motivation

The module tree was pretty outdated and many deleted dependencies still had some find modules which weren't used. So a small cleanup.


Baklap4:
Not quite sure why osx build is failing with the message:
```
ld: framework not found -lCocoa
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
As it still links to it 🤔

Baklap4:
And the problem is fixed.. i forgot some quote....

deuteragenie:
> ... fixed dependencies in apt-get statement in ci.

According to https://docs.semaphoreci.com/ci-cd-environment/ubuntu-18.04-image/ 
Python is pre-installed and so is pip3

=> python3-dev python3-pip dependencies can be removed
=> sem-version python 3.8 can be upgraded to 3.9
=> sudo pip3 can be replaced by pip3 (ie without sudo), as it is run from a virtual env

As the meson build is limited to ensuring proper compilation and linking, you can use "meson --buildtype=plain ....", which in my environment saves another 20 seconds.

Baklap4:
> Python is pre-installed and so is pip3
> => python3-dev python3-pip dependencies can be removed
> => sem-version python 3.8 can be upgraded to 3.9
> => sudo pip3 can be replaced by pip3 (ie without sudo), as it is run from a virtual env

These are already fixed within another PR (#627) so let's await that :)

OznOg:
Hard to challenge thing in here without actually trying to compile.
Nevertheless, the removal of all those dead files is a good idea, thus LGTM

OznOg (approved):

Baklap4:
> Hard to challenge thing in here without actually trying to compile.

Yeah CI on mac fails. This is due to SDL2 so i'll have to do another commit on this first

OznOg (approved):
LGTM

PR #642, opened 2021-08-06
Prepare for AppImage requirements

deuteragenie:
- Add an initial AppImageBuilder file which will be used to generate the AppImage.  It has been tested on several Ubunutu versions and the resulting AppImages have been linted successfully for compatibility with many distros.

- Reorder the .desktop file items in accordance with the Freedesktop specs, review keywords. The file has also been validated with the freedesktop dedicated tool.

- Add a set of fixed icons based on the .svg, as .svg is not required to be supported by all distros, according to the freedesktop specs.  These icons will be installed at the correct location in the revised build scripts.  They can be used depending on context (browser, sidebars, menus etc.) by the OS.


Baklap4:
Can you create a little tutorial on how to use this within this PR and / or maybe wiki?

deuteragenie:
I'll update the wiki.

deuteragenie:
I described how to build the AppImage here:

https://github.com/performous/performous/wiki/Building-and-installing-from-source#Building_an_AppImage_Experimental

Baklap4:
Currently fails. Waiting for #627 then this can work

deuteragenie:
> Currently fails. Waiting for #627 to merge then this can work

How could this PR fail if it is only adding a few files that will be required later on if/when AppImage will be build ?

Baklap4:
@ooshlablu can you extend on this and add this AppImage as one of the artifacts to our current workflows in gh-actions?

Baklap4:
Closing this in favor of #736

PR #643, opened 2021-08-07
Add basic vscode config files

deuteragenie:
This PR adds a basic settings.json for VS Code which contains some edit conventions for this app.
It also adds a c_cpp_properties.json to enable intellisense / auto-completion on Linux (Mac and Windows are TBD).
Tested on a few different file types and looks fine.


Baklap4:
We had a little discussion on discord about this pr.
Resume: 
- we have an upcoming PR to fix stuff like line-endings and tabs and so on.
- And as many of us use different editors we didn't find it handy having a specific section for vscode.

My tip if you want to compile performous with vscode install the cmake add-ons. This will basicly configure everything for you without having to do anything

Closing this PR in favor of #717

PR #644, opened 2021-08-08
Include specific header for opencv

deuteragenie:
For webcam support, include the specific header required (ie <opencv2/videoio.hpp>
) rather than the generic one (ie #include <opencv2/opencv.hpp>)

This results in faster parse time.

When installing dependencies, we can probably only install opencv-videoio, which has a smaller footprint than the generic opencv and depends on opencv-core which is also required.

Tested on Ubuntu 21.04.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -17,7 +17,7 @@ blocks:
           - echo "Installing Performous dependencies..."
           - sudo apt -y update
           - sudo add-apt-repository -y ppa:litenstein/performous-bionic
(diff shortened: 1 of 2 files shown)


Baklap4:
> When installing dependencies, we can probably only install opencv-videoio, which has a smaller footprint than the generic opencv and depends on opencv-core which is also required.

Can you add the install of the specific library to this pr?

Baklap4 (approved):
LGTM

Baklap4:
@deuteragenie This PR broke the meson build

deuteragenie:
Let us wait until the OSX PR is merged first.

PR #645, opened 2021-08-09
Added skipping meson build

Baklap4:
### What does this PR do?

Skips the meson build when the current branch isn't `meson` as suggested by [semaphore](https://github.com/semaphoreci/semaphore/issues/68)

### Closes Issue(s)

* None

### Motivation

Meson is still experimental phase and our main building system is cmake. For the time being this will continue being our main building system. We'll have a separate branch called `meson` which we'll update every now and then to the latest master. This way we still support and know our meson build is successful. And can it not stop our PR's from getting merged.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -82,6 +82,8 @@ blocks:
     execution_time_limit:
       minutes: 10
     dependencies: []
+    skip:
+      when: "branch != 'meson'"
     task:
       prologue:
         commands:


OznOg:
is that completely skipped or "allowed to fail"?

Baklap4:
Completely skip. We'll use the [meson](https://github.com/performous/performous/tree/meson) branch to keep on testing with ci for meson. This branch will have to be brought up to date every now and then manually.

OznOg:
ok

OznOg (approved):

PR #649, opened 2021-08-10
Removed unused defines from config.cmake.hh

Baklap4:
### What does this PR do?

Removes unused defines from `config.cmake.hh`

### Closes Issue(s)

None

### Motivation

Why do we always keep unused stuff? Clean up when we can!


deuteragenie:
Ideally, all defines should be in this file.  For example USE_BOOST_FS is not listed...

Baklap4:
> Ideally, all defines should be in this file. For example USE_BOOST_FS is not listed...

Not sure if i agree with that statement. Those `USE_XXX` are added as defines throughout cmake itself using the [`add_definitions()`](https://cmake.org/cmake/help/latest/command/add_definitions.html) function. So to add them to this file would kinda make those calls obsolete?

Unless you want them to be always set within our codebase? Now they're conditionally set to 1 or NULL

deuteragenie:
I find it quite useful to have one central place which contains all defines and the doc / comments describing what the define is about.

cmake should then only set the *values* for existing defines, but not create new ones "out of the blue".

Baklap4:
For now let's keep them out and create a new pr with all these included

OznOg:
> I find it quite useful to have one central place which contains all defines and the doc / comments describing what the define is about.
> 
> cmake should then only set the _values_ for existing defines, but not create new ones "out of the blue".

I partly agree with that. I can agree that we may have a special header for setting "defined" values, but I quite dislike having a central file getting those definitions as it trigs a rebuilds of the whole project when just changing a side part of it.
I guess we could have a fs_config.hh generated that would have the FS related defines and just pull it from the "right" place (fro fs.hh probably).
Moreover, having config.hh file really is useful when creating libraries to let the user know which features where activated at build time, I think this may not be that helpful when creating "leaf" binary against no one will even link.

OznOg (approved):

PR #652, opened 2021-08-13
Fix autodetecting webcam

Baklap4:
### What does this PR do?

Fixes autodetecting the webcam.
It never was that when you passed `-1` that it'd autodetect it. At least since 11 years back it's been 0 according to the [opencv code](https://github.com/opencv/opencv/blob/master/modules/videoio/src/cap.cpp#L219-L220).

### Closes Issue(s)

None

### Motivation

We mustn't tell lies:
image

### Additional Notes

Try autodetecting with and without this fix.
On Windows it didn't work before this fix.


Baklap4:
Still works on Ubuntu 21.04

OznOg (commented):
I'm not a great fan of magic values, but it is not worse than before.
LGFM

Baklap4:
Changed to a field so we don't have a magic number :)

game/webcam.hh
```diff
@@ -47,6 +46,7 @@ class Webcam {
 	bool m_frameAvailable;
 	std::atomic<bool> m_running{ false };
 	std::atomic<bool> m_quit{ false };
+	const int m_autoDetect;
```
OznOg, line 49:
```suggestion
	const int m_autoDetect = 0;
```

game/webcam.cc
```diff
@@ -19,15 +19,15 @@ namespace cv {
 #endif
 
 Webcam::Webcam(int cam_id):
-  m_thread(), m_capture(), m_writer(), m_frameAvailable(false)
+  m_thread(), m_capture(), m_writer(), m_frameAvailable(false), m_autoDetect(0)
```
OznOg, line 22:
```suggestion
  m_thread(), m_capture(), m_writer(), m_frameAvailable(false)
```
see below

OznOg (approved):

PR #653, opened 2021-08-13
Pitchwave order

### LINKED ISSUE #199: Microphone colors rendering order
I think when singing, the pitchwave of the player with the most points should be on the top.
Now the order is strange yellow>purple>green>red>blue>lightblue it seems hard coded.

If you are playing with 6-7 players, not everyone can see their colors, just those that are on the top so it should be those who has the most points :)

image

Baklap4:
### What does this PR do?

Fixes the pitchwave order. The highest score gets on top.
We used to just iterate over the players and draw them one after another.

### Closes Issue(s)

Closes #199

### Motivation

The best will be king of the hill no more no less.

### Additional Notes

I'm not so sure about the copying the `database.cur` to a `std::list<Player>` but i couldn't get sort to work on `database.cur` ...

### DIFF
--- a/game/notegraph.cc
+++ b/game/notegraph.cc
@@ -203,9 +203,9 @@ namespace {
 void NoteGraph::drawWaves(Database const& database) {
 	if (m_vocal.notes.empty()) return; // Cannot draw without notes
 	UseTexture tblock(m_wave);
-	std::list<Player> players = database.cur;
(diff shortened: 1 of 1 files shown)


Baklap4:
thanks for testing @ooshlablu i removed the 'needs testing' label

OznOg (approved):

PR #656, opened 2021-08-17
Re enabled ppa

Baklap4:
### What does this PR do?

Re enabled ppa
Using libxml++3
Reverted only component install of opencv (#644)

### Motivation

Wanna use libxml3++


Baklap4:
This creates the bug that deb packages can't be build due to conflicts with ICU. But i'd rather have that then failing pipelines.

OznOg (approved):

PR #657, opened 2021-08-17
New sorting option for songs by score

### LINKED ISSUE #774: Trying to change sort criteria in screen_songs kicks you to main menu

Performous: 1.2.0
Operating System: macOS 11.6.6

### What is the problem?

Sorting can't be changed

### Steps to reproduce

Go into the song selection screen and try to change the sort ordering.

### Additional context

Using git bisect, I've traced the issue to 5ca3d79b6e5273e98201669092b093d509eb344d, but I have no idea what change in that commit could be responsible for this.

_No response_

twollgam:
### What does this PR do?

This pr adds a new option for sorting songs by the maximum song score.

### Closes Issue(s)

Closes #774 

### Motivation

That allows to sort the songs that are interesting to the front of the list. Songs never sung are at the end.


Baklap4:
Looks like a nice feature! Still gotta test it :)

game/scorewindow.hh
```diff
@@ -0,0 +1,29 @@
...
+class ScoreWindow {
+  public:
+	ScoreWindow(Instruments& instruments, Database& database);
+    
+	/// draws ScoreWindow
+	void draw();
+	bool empty();
+    
+  private:
+	Database& m_database;
```
OznOg, line 21: should/can this be const?

game/songs.hh
```diff
@@ -77,11 +77,12 @@ class Songs {
 	size_t loadedSongs() const { return m_songs.size(); }
 
   private:
-  	void LoadCache();
+	void LoadCache();
 	void CacheSonglist();
 
 	class RestoreSel;
-	typedef std::vector<std::shared_ptr<Song> > SongVector;
+	using SongPtr = std::shared_ptr<Song>;
```
OznOg, line 84: this kinda using on pointers is usually fishy and here seems really little used.

game/songs.hh
```diff
@@ -77,11 +77,12 @@ class Songs {
...
+	using SongVector = std::vector<SongPtr>;
```
OznOg, line 85: putting the container type in the name is usually a bad idea as you dons express the "what" but the "how". Songs? SongCollection?

game/songs.cc
```diff
@@ -509,39 +516,37 @@ void Songs::sortChange(int diff) {
...
-		  case 4: std::sort(m_filtered.begin(), m_filtered.end(), customComparator(&Song::genre)); break;
-		  case 5: std::sort(m_filtered.begin(), m_filtered.end(), customComparator(&Song::path)); break;
-		  case 6: std::sort(m_filtered.begin(), m_filtered.end(), customComparator(&Song::language)); break;
+		  case 1: std::sort(begin, end, customComparator(&Song::collateByTitle, !descending)); break;
+		  case 2: std::sort(begin, end, customComparator(&Song::collateByArtist, !descending)); break;
+		  case 3: std::sort(begin, end, customComparator(&Song::edition, !descending)); break;
+		  case 4: std::sort(begin, end, customComparator(&Song::genre, !descending)); break;
+		  case 5: std::sort(begin, end, customComparator(&Song::path, !descending)); break;
+		  case 6: std::sort(begin, end, customComparator(&Song::language, !descending)); break;
+		  case 7: std::sort(begin, end, [this, descending](SongPtr const& a, SongPtr const& b){
```
OznOg, line 545: capture descending by ref?

game/songitems.cc
```diff
@@ -71,17 +71,19 @@ void SongItems::addSong(std::shared_ptr<Song> song) {
...
-int SongItems::lookup(std::shared_ptr<Song> song) const {
-	for (auto const& s: m_songs) {
-		if (song->collateByArtistOnly == s.artist && song->collateByTitleOnly == s.title) return s.id;
-	}
+int SongItems::lookup(std::shared_ptr<Song> const& song) const {
+	for (auto const& s: m_songs)
+		if (song->collateByArtistOnly == s.artist && song->collateByTitleOnly == s.title) 
+			return s.id;
+
 	return -1;
```
OznOg, line 79: this id seems to eventually be used as a unsigned; maybe return an std::optional<> ?

Baklap4:
Just tested it. It seems to work but what happens in the case where there are 2 songs with the same highscore? Which song will be shown first?

Also a remark: we don't show the actual score (within the songbrowser) which makes the sorting seem random at first glance.
Maybe within a new pr add the highest score + playername to the songbrowser window.

game/scorewindow.hh
twollgam: Should, yes, but can not, because ScoreWindow sort the scores inside the Database class. Bad designed but I want not to refactor it at this point. Except @OznOg you has an idea for an easy solution.

game/songs.hh
twollgam: Any suggestion?

game/songs.cc
twollgam: Would work, but as descending is a bool a reference will have no benefit. Or am I wrong?

If reference it should de a const ref.

twollgam:
@Baklap4  all songs with equal score get random order. I like it but any suggestion? Sort by song name?

game/songitems.cc
twollgam: I pushed the change. I am not sure if it is a good solution with std::optional. @OznOg please take a look.

Baklap4:
Let's see if we can rebase this and then get it to merge. New functionality is pretty nice!

Baklap4:
@OznOg can you resolve the discussions if they're addressed?

meson.build
```diff
@@ -0,0 +1,273 @@
+#Project definition
```
ooshlablu, line 1: I believe we got rid of the Meson stuff. This file was probably in there when you originally started working on this PR though. It should be removed.

ooshlablu:
Tested it with vocals and guitar on Ubuntu 18.04 and 20.04, and everything seems to work as expected. I can't really comment on the code, so I leave that up to others for review.

twollgam:
@OznOg  can you do a code review again please. Would be nice if this pr will be merged.

game/songs.hh
OznOg: using unique_ptr?

OznOg (approved):

PR #659, opened 2021-08-18
Bugfix: Clamping score to 10k points

### LINKED ISSUE #577: Crash when achieved score > 10000

Bug

### What did you do?

- selected a full-band song
- only played guitar part
- scored more that 10 000 points (flawless) - I think it was 10 460 or so
- got "Virtuoso" evaluation
- then a crash with:
```
core/error: Exiting due to fatal error: Invalid score value
FATAL ERROR: Invalid score value
core/error: Invalid score value
```

score clipped at 10 000 points

### What did you see instead?
crash

```
Performous 1.1+
```

ArchLinux, x86_64

Baklap4:
### What does this PR do?

Fixes a bug where the score could reach more than 10k points. This caused the game to crash with a runtime exception: "Invalid score value" when the highscore is reached.
Initially we thought the bug appeared while singing. But later we found out it's during guitars.

### Closes Issue(s)

Closes #577 

### Motivation

Fixes a bug

### More

### Additional Notes

I've only added a clamp around the `getscore()` method. I'm not sure if this is visually enough. We don't want the game to show more than 10k points at any given time: playing time, highscore time

### DIFF
--- a/game/instrumentgraph.hh
+++ b/game/instrumentgraph.hh
@@ -13,6 +13,7 @@
 #include "menu.hh"
 #include "screen.hh"
 #include "fs.hh"
+#include "util.hh"
 
 /// Represents popup messages
 class Popup {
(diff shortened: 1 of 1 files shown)


twollgam (commented):
Question: Can the first parameter of clamp only be m_score * m_scoreFactor  without the greater zero check?

I think that clamp will take attention to negative values too due to the lower bound of 0.0.

Baklap4:
> Can the first parameter of clamp only be m_score * m_scoreFactor without the greater zero check?

@twollgam not sure if i understand this question correctly, but i think you're suggesting to remove the if statement since clamp should already take care of this right?

`clamp(x, y, z)` -> `x` = value to clamp, `y` is lower bounds, `z` is higher bounds

So yeah i'll remove that check :)

twollgam (approved):
Looks good to me.

PR #663, opened 2021-08-31
Fix keyboard instrument/nav keys getting stuck if released while holding mod keys

### LINKED ISSUE #232: Keypressed Loop
While sorting out songs i came acros this 'bug'

# Act
Within the song browser i'm holding the Right arrow. Then i press ctrl + R 

# Thought it'd do
Reload the whole song library.

# It does the following
- Reloads the song library and keeps on browsing songs to the right, which heavenly speeds up the animation.

# To stop the loop
- Press the same arrow key and quickly press the oposite arrow key.

Tronic:
Fix #232

### DIFF
--- a/game/controllers-keyboard.cc
+++ b/game/controllers-keyboard.cc
@@ -11,6 +11,7 @@ namespace input {
 	class Keyboard: public Hardware {
 		std::set<unsigned> m_pressed;
 		bool m_guitar, m_keytar, m_drumkit, m_dancepad;
+		unsigned m_mod;
 		/// Set the value of an instrument bool from a config variable and return a log message snippet if value was changed.
 		static std::string setMode(bool& var, std::string const& name) {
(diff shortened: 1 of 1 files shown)


Baklap4:
will test this out later this week! Thanks for the pr 👍

Baklap4 (changes requested):
LGTM but i'd use the private member instead of the created variable for the rest of the method. If we're not using the member variable than we should remove the `m_mod = mod` line

The bug is fixed with this :)

Baklap4 (approved):
LGTM @ooshlablu is still testing though let's await that and then merge it :)

ooshlablu:
I can confirm this fix works on Ubuntu 18.04 and 20.04.

Baklap4:
Thanks for the fix @Tronic 👍

PR #664, opened 2021-08-31
Move video FFMPEG init back to thread

Tronic:
- Reverting part of 977fb50d7e529730b100519ac1692a0c5ed02d09
- Catch errors to avoid crash

No I/O should be done in main thread because it may block, so the initialisation needs to be done in a separate thread, keeping main thread always running smoothly. The current master was also not catching any loading errors, and thus would crash the game whenever a video failed to load.

Probably a similar fix needs to be done to audio ffmpeg, but I cannot look into that right now.

### DIFF
--- a/game/video.cc
+++ b/game/video.cc
@@ -37,7 +37,7 @@ void Video::push(Bitmap&& f) {
 	m_queue.emplace_back(std::move(f));
 }
 
-Video::~Video() { 
+Video::~Video() {
 	{
 		std::lock_guard<std::mutex> l(m_mutex);
 		m_quit = true;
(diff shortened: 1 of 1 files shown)


game/video.cc
```diff
@@ -47,40 +47,46 @@ Video::~Video() {
...
+						break;
+					}
 				}
-				m_cond.wait(l, [this]{ return m_quit || m_seek_asked; });
-			} catch (std::exception& e) {
-				UnlockGuard<decltype(l)> unlocked(l); // release lock for possibly blocking calls
-				std::clog << "ffmpeg/error: " << file << ": " << e.what() << std::endl;
-				if (++errors > 2) { std::clog << "ffmpeg/error: FFMPEG terminating due to multiple errors" << std::endl; break; }
 			}
+		} catch (std::exception& e) {
```
OznOg, line 88: maybe just keep the try/catch around VideoFFmpeg if it is the only place to throw; here it is quite hard to tell where the exception can come from.

OznOg (approved):

OznOg:
Did you have particular issue with this or did you see that while reading the code?

game/video.cc
Tronic: I'd do that but then the `ffmpeg` variable lives only inside that scope and cannot be used in later code. All the alternatives are somewhat ugly (like not using `auto` but manually defining the variable).

Tronic:
I had the game crashing "randomly" in song browser because of this uncaught exception on one specific song. Digging into the code I noticed the other problem (construction of ffmpeg would cause a slight pause in song browser animation).

Back in the day I spent a lot of sweat to move cover image loading, ffmpeg and everything to worker threads to allow the song browser to run at constant 60 FPS without missing any frames. If audio loading now happens in main thread, that also needs to be moved to a worker thread (as I believe it was originally). If the song collection was on a HDD in sleep, the UI would freeze for about 20 seconds while the disk spins up, if any I/O occurs in the main thread.

deuteragenie:
I can only confirm that browsing is a resource hog.  In particular, browsing like mad can consume MORE cpu than playing the actual game/song.  So indeed, particular care needs to be put in the song browser.

ooshlablu:
@Tronic I've noticed that crash in the song browser as well.

Baklap4:
Are we leaving the `catch (std::exception& e)` in or not? @OznOg @Tronic 
If not let's merge it

Personally i wouldn't leave it in as other errors shouldn't pop up here

OznOg:
> I can only confirm that browsing is a resource hog. In particular, browsing like mad can consume MORE cpu than playing the actual game/song. So indeed, particular care needs to be put in the song browser.

I agree, but my experience is that aubio computation causes hangs/cpu load on my system. I agree that we should work on making this smooth as possible, but I also challenge optimization based on "feelings". 

That was the purpose of my question.

game/video.cc
OznOg: ok

Baklap4:
> but I also challenge optimization based on "feelings".

Keep doing that, that's a good perk to have!

PR #668, opened 2021-09-03
Changed from map to unordered map and added a hash function for fs::path

Baklap4:
### What does this PR do?

Changes covers from a map type to unordered_map to get a O(1) find operation instead of OLog(n)
Also cleans up the insert statement.

### Closes Issue(s)

None

### Motivation

Better faster stronger

### Additional Notes

Did some benchmarking. It's not much but every little bit counts.

### DIFF
--- a/game/fs.hh
+++ b/game/fs.hh
@@ -100,10 +100,9 @@ Paths getThemePaths();  ///< Get the data/theme file search path (includes curre
 Paths getPathsConfig(std::string const& confOption);  ///< Return expanded list of paths specified by a path config option
 
 template <>
-class std::hash<fs::path>
+struct std::hash<fs::path>
 {
-public:
-    size_t operator()(const fs::path& path) const
+    size_t operator()(const fs::path& path) const noexcept
     {
         return fs::hash_value(path);
     }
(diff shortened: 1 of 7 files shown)


OznOg (approved):

game/fs.hh
```diff
@@ -99,3 +99,12 @@ Paths const& getPaths();  ///< Get the data file search path
 Paths getThemePaths();  ///< Get the data/theme file search path (includes current and default themes in addition to data folders)
 Paths getPathsConfig(std::string const& confOption);  ///< Return expanded list of paths specified by a path config option
 
+template <>
+class std::hash<fs::path>
```
OznOg, line 103: std::hash seems to be a struct (AFAIR, windows care about struct/class), and this would prevent the need for `public:` ; may also be noexcept

OznOg (changes requested):

game/fs.hh
Baklap4: Changed so it's a struct instead of class and added the noexcept

game/screen_players.hh
```diff
@@ -1,5 +1,5 @@
 #pragma once
-
+#include <unordered_map>
```
OznOg, line 2: would be a better idea to include after local includes

Baklap4: Fixed :)

OznOg (approved):

PR #669, opened 2021-09-04
Converted plain enums to enum classes

Baklap4:
### What does this PR do?

Converts every plain enum to an enum class.

### Closes Issue(s)

None

### Motivation

With enumerations you most likely want to have strongly typed classes with when necessairy a backingtype (default is unsigned int). 
This improves readability, refactoring and possibly wrong assignments/checks. Take the following snippet in account:
```
enum Color { red, green, blue };                    // plain enum 
enum Card { red_card, green_card, yellow_card };    // another plain enum 
enum class Animal { dog, deer, cat, bird, human };  // enum class
enum class Mammal { kangaroo, deer, human };        // another enum class

void fun() {

    // examples of bad use of plain enums:
    Color color = Color::red;
    Card card = Card::green_card;

    int num = color;    // no problem

    if (color == Card::red_card) // no problem (bad)
        cout << "bad" << endl;

    if (card == Color::green)   // no problem (bad)
        cout << "bad" << endl;

    // examples of good use of enum classes (safe)
    Animal a = Animal::deer;
    Mammal m = Mammal::deer;

    int num2 = a;   // error
    if (m == a)         // error (good)
        cout << "bad" << endl;

    if (a == Mammal::deer) // error (good)
        cout << "bad" << endl;

}
```
In the above example a wrong assignment + check is showcased which compiles just fine but resulting in errors.

### Additional Notes

There were 2 enums i couldn't get converted. These are located within https://github.com/performous/performous/blob/feature/enumClass/game/controllers.hh#L18 and https://github.com/performous/performous/blob/feature/enumClass/game/controllers.hh#L27

Within `controllers.hh/cc` there's some black magic going on creating a define function which uses `controllers-buttons.ii`. I did fix that with `DevType` but couldn't get it to work for `ButtonId` and `NavMenu`. So it'd be nice if someone else can make a commit to fix those and commit them to this PR. You can get these errors bij adding the `class` keyword to the above mentioned lines.

### Testing Notes

The enums are basicly everywhere throughout the codebase. Special testing care has to be taken to the controllers (keyboard, guitar, drums etc).


game/controllers.hh
```diff
@@ -12,31 +12,31 @@
...
-		NAV_UP, NAV_DOWN, NAV_LEFT, NAV_RIGHT, NAV_MOREUP, NAV_MOREDOWN, NAV_VOLUME_UP, NAV_VOLUME_DOWN
+	enum class NavButton {
+		NONE /* No NavEvent emitted */, SOME /* Major gameplay button with no direct nav function, used for joining instruments */,
+		START, CANCEL, PAUSE,
+		REPEAT = 0x80 /* Anything after this is auto-repeating */,
+		UP, DOWN, LEFT, RIGHT, MOREUP, MOREDOWN, VOLUME_UP, VOLUME_DOWN
 	};
 	/// Alternative orientation-agnostic mapping where A axis is the one that is easiest to access (e.g. guitar pick) and B might not be available on all devices
-	enum NavMenu { NAVMENU_NONE, NAVMENU_A_PREV, NAVMENU_A_NEXT, NAVMENU_B_PREV, NAVMENU_B_NEXT };
+	enum class NavMenu { NAVMENU_NONE, NAVMENU_A_PREV, NAVMENU_A_NEXT, NAVMENU_B_PREV, NAVMENU_B_NEXT };
```
OznOg, line 25: I guess NAVMENU_ can be removed as well

OznOg (approved):

game/controllers.hh
Baklap4: probably yes, it's the same :o

OznOg (approved):
So great!

PR #674, opened 2021-09-10
Updated filename to match zip file

Baklap4:
### What does this PR do?

Tries to fix an issue with the meson.zip which was renamed to master.zip

### Motivation

Meson build failed


Baklap4:
Closing this since meson support is currently very fragile. And a poll is running in discord to either support it fully or don't support it at all.

PR #677, opened 2021-09-11
Add german translations for rc1.2

twollgam:
This commit provides most of the missing german translations for rc1.2


twollgam:
Translation is not complete but as far as I  can do it. Should be reviewed.

Baklap4:
Can you change the target branch to be feature/ConsistentLanguagePot and also rebase on top of that?
The current master branch doesn't contain some fixes presented on that branch :)

Issue #679, opened 2021-09-13
Add support for enum translation


twollgam:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

The file schema.xml contains mappings from text to integer. The description of every item resists inside an "enum" xml tag. All those descriptions are english and while the is no translation yet they are displayed in english even if all other texts are displayed in another language.

Tasks:

- [x] Implement i18n support for enum
- [ ] Implement i18n support for stringvalue
- [x] Extend the xml_gettext.py that it extracts the enum-tags from schema.xml 
- [ ] Extend the xml_gettext.py that it extracts the  stringvalue-tags from schema.xml 
- [x] Add example translation

Baklap4:
Can you verify that strings in a stringlist and/or optionlist are translatable? if not they should be added too.

twollgam:
@Baklap4 the stringvalue tag is rarely used in schema.xml and I did not find them in the configuration screens.  Only I found "Word collation" but that did not work (anymore). It displays < 6 items >. That is not what I expect.

Lord-Kamina:
> @Baklap4 the stringvalue tag is rarely used in schema.xml and I did not find them in the configuration screens. Only I found "Word collation" but that did not work (anymore). It displays < 6 items >. That is not what I expect.

Word collation is working as expected (It's the list of words that are ignored for sorting purposes, such as "The", "A" and such) but the interface does not allow directly editing a stringlist.

PR #682, opened 2021-09-15
Add support for enum translation

twollgam:
### What does this PR do?

The file schema.xml contains mappings from text to integer. The description of every item resists inside an "enum" xml tag. All those descriptions are english and while there is no translation yet they are displayed in english even if all other texts are displayed in another language.

### Closes Issue(s)

#679 Add support for enum translation

### Motivation

Full localization: Translate string that are currently available only in english.


Baklap4:
Converted to draft + removed `WIP:` from the title

Baklap4:
Can you please revert the line-ending change in `i18n.hh`? As this currently shows that the file changed completely while this is not the case i believe

twollgam:
I merged current master into this PR. All po-files have had conflicts so that a reviewer should have a look a them.

Baklap4:
instead of merging master into this, please rebase it:
execute `git rebase -i master` on your own branch and then force push the branch
This will put your commits on top of latest master

OznOg:
@Baklap4 didn't you fix that actually? seems to me the lang is correctly translated in the menu and is no more in the schema. Maybe the descrption should be updated after rebase because I sadly don't understand it anymore.

Baklap4:
> @Baklap4 didn't you fix that actually? seems to me the lang is correctly translated in the menu and is no more in the schema. Maybe the descrption should be updated after rebase because I sadly don't understand it anymore.

I didn't. The thing is that everything within schema.xml should also be translated. The languages is a dynamic so therefor can't be added to the schema.xml see: https://github.com/performous/performous/blob/master/data/config/schema.xml#L98-L104

For the languages to be translateable we've added them within the code as shown here https://github.com/performous/performous/blob/master/game/i18n.hh#L99-L118

However for things like `notegraphscalingmode` which is listed within the schema.xml isn't translateable at this very moment

Baklap4:
This branch is in need to rebasing, aside from that it looks good to go i think.

OznOg:
Yes I eventually figured out what was being fixed here. I think we should merge it.

game/configuration.cc
```diff
@@ -195,7 +195,8 @@ void ConfigItem::removeAllEnums() {
 
 void ConfigItem::selectEnum(std::string const& name) {
 	auto it = std::find(m_enums.begin(), m_enums.end(), name);
-	if (it == m_enums.end()) throw std::runtime_error("Enum value " + name + " not found in " + m_shortDesc);
+	if (it == m_enums.end())
```
OznOg, line 198: useless part or broken indent (if you was to keep 2 lines)

twollgam: :+1: Fixed that.

lang/performous.pot
```diff
@@ -1,2205 +1,2205 @@
...
+msgid "Harmonic 1"
+msgstr ""
+
+#: ../game/song.hh:28
+msgid "Harmonic 2"
+msgstr ""
+
+#: ../game/song.hh:28
+msgid "Harmonic 3"
+msgstr ""
```
Baklap4, line 2205: It might be that i'm just overlooking things, but this PR should fix that enumerations within schema.xml gets translated, but i don't see `../data/config/schema.xml:133` (https://github.com/performous/performous/blob/master/data/config/schema.xml#L132-L137) within the main `performous.pot` ? I do see them in the `de.po`

lang/xml_gettext.py
```diff
@@ -56,7 +56,26 @@ def _start(self, *args, **kwargs):
...
             for k, v in parent[elem].attrib.items():
-                info += f"#.   {k}: {v}\n"
+                info += f"#. {k}: {v}\n"
+            po[text] = (
+                f'\n{info}'
+                f'#: {filename}:{elem.lineno}\n'
+                f'msgid "{text}"\n'
+                'msgstr ""\n'
+            )
+        if elem.tag in ("enum"):
```
Baklap4, line 66: not sure if these needs to be picked up by `xml_gettext.py`
```
<ui unit=" ms" />
```

ms being short for milisecond, but can have a different translation in other languages

twollgam: Pushed 2 commits to support this.

lang/performous.pot
twollgam: Fixed

Baklap4:
Will take a final look later tonight, but i think this is soon to be merged 👍

OznOg (approved):

Baklap4:
@twollgam @OznOg any of you  know why Windows build step gives out all those warnings within `configuration.cc`? Is this actually a warning or is msvc just tripping?

PR #688, opened 2021-09-27
Hotfixes/1.2.0 rc1

Baklap4:
### What does this PR do?

A couple of hotfixes for the 1.2.0 rc-1

### Closes Issue(s)

None.

### Motivation

Some stuff was broken on release.


cmake/Modules/FindAubio.cmake
```diff
@@ -1,14 +1,15 @@
-
 if (NOT USE_SELF_BUILT_AUBIO)
     find_package(PkgConfig REQUIRED)
     if(PKG_CONFIG_FOUND)
         # The version we look for is legacy, may need some fine tunning
         pkg_check_modules(AUBIO IMPORTED_TARGET aubio>=0.4.9)
+            if(AUBIO_FOUND)
+                add_library(aubio ALIAS PkgConfig::AUBIO)
```
ooshlablu, line 7: This is causing an error for cmake < 3.19. I get the following error on Ubuntu 18.04 and 20.04, and on Fedora 32:
```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Checking for module 'aubio>=0.4.9'
--   Found aubio, version 0.4.9
CMake Error at cmake/Modules/FindAubio.cmake:7 (add_library):
  add_library cannot create ALIAS target "aubio" because target
  "PkgConfig::AUBIO" is imported but not globally visible.
  ```

ooshlablu (changes requested):
The error can be resolved by upgrading cmake on the operating systems affected, but that would mean deviating far from the norm on at least 2 operating systems that are still supported by their developers. I have not tried on Ubuntu 21.04, but if this will build fine with Ubuntu 22.04, then maybe it is worth the change. Perhaps updating the docs to explicitly reflect that a version of cmake  external to the Ubuntu repos is required, and producing packages for consumption on the affected versions of Ubuntu will suffice.

ooshlablu:
I did just run a build in an Ubuntu 21.10 container, and it builds and packages without issue. CMake version is `3.18.4`, so I think 22.04 shouldn't have any issue. Fedora 32 has been deprecated with 33 and 34 being the currently supported versions.

ooshlablu (approved):
Packages build for Ubuntu 18.04 & 20.04 and Fedora 32 - 34 fine.

Tested with singing and guitar/bass on both versions of Ubuntu, and everything works as expected.

ooshlablu (approved):
Tested the latest commit on Ubuntu 18.04 and 20.04 with singing and guitars. Everything seems :+1: from my side.

PR #694, opened 2022-01-01
Fix possible invalid memory access

OznOg:
The function tellg() can return -1 making the overall vector size to be
0 and thus accessing the back element is invalid.

g++ 11 complains about it and reports a warning

This patch adds an extra check in case tellg fails

### DIFF
--- a/game/glshader.cc
+++ b/game/glshader.cc
@@ -16,6 +16,7 @@ namespace {
 		if (!f) throw std::runtime_error(std::string("Couldn't open ") + filepath);
 		f.seekg(0, std::ios::end);
 		size_t size = f.tellg();
+		if (size == static_cast<decltype(size)>(-1)) throw std::runtime_error(std::string("Cannot get size of file ") + filepath);
 		f.seekg(0);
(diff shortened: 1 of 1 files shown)


game/glshader.cc
```diff
@@ -16,6 +16,7 @@ namespace {
 		if (!f) throw std::runtime_error(std::string("Couldn't open ") + filepath);
 		f.seekg(0, std::ios::end);
 		size_t size = f.tellg();
+                if (size == static_cast<decltype(size)>(-1)) throw std::runtime_error(std::string("Cannot get size of file ") + filepath);
```
twollgam, line 19: Use 2 tabs here for indent.

twollgam (changes requested):
LGTM except the indent.

Baklap4:
I'll try to compile it later today

game/glshader.cc
OznOg: grumpf

Baklap4 (approved):
LGTM

PR #635, opened 2021-08-05
Try out docker

ooshlablu:
### What does this PR do?
Add docker pulls from my repo https://github.com/ooshlablu/performous-docker to start enabling builds for other platforms besides what semaphore has to offer at the moment.

PR #708, opened 2022-04-03
Try out docker builders for Linux on Semaphore

ooshlablu:
Replacement for https://github.com/performous/performous/pull/635


Baklap4:
Closing this in favor of #715 since within GH-Actions we make use of our docker containers for the linux build.
Docker builds have been relocated from dockerhub to github actions aswell and is maintained here: https://github.com/performous/performous-docker

Since we're going to phase out semaphore this is probably not needed anymore.

PR #709, opened 2022-04-03
Fix races on m_songs in Songs

OznOg:
The m_songs member was not locked correctly.

This was leading to random crashes while loading many songs and/or while applying filters while loading many songs.

### DIFF
--- a/game/songs.cc
+++ b/game/songs.cc
@@ -48,12 +48,12 @@ void Songs::reload() {
 
 void Songs::reload_internal() {
 	{
-		std::lock_guard<std::mutex> l(m_mutex);
+		std::unique_lock<std::shared_mutex> l(m_mutex);
 		m_songs.clear();
 		m_dirty = true;
 	}
 	LoadCache();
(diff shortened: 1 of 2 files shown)


Baklap4:
YES FINALLY! i had this sometimes, but never could really get a lead on to where it happened

Baklap4:
Is there any way to actually test this properly?
As to me it happened mostly at random times :/

OznOg:
> Is there any way to actually test this properly? As to me it happened mostly at random times :/
> 
> Maybe, if possible, add a unittest for this since with regular play it's hard to actually reproduce it

Like any race, I can be quite hard to do on purpose. You may add some sleeps on the code when loading the cache to make the race more "likely". Creating the race itself can be done with some breakpoints on gdb.

Baklap4 (dismissed):
LGTM

OznOg:
I'm affraid there's still something wrong with the filter restore..... Do not merge for now....

OznOg:
> I'm affraid there's still something wrong with the filter restore..... Do not merge for now....

well actually not, the Restore works only on filtered not songs directly.

Baklap4 (approved):
I can't un-dismiss hahah so LGMT

PR #710, opened 2022-04-03
Adds a skip option while iterating through directories

### LINKED ISSUE #700: Errors in folder browser

Bug

### What did you do?

Tried browsing folders in Windows 10.

1) Browsing folders and not being able to enter unauthorized folders.

2) This is just usability but IMO, but navigation could use some improving. Either `Escape` should go one folder up and only exit the screen if on top level of the hierarchy, or maybe we should set up left and right arrows to go one folder up and down respectively. Might also be a good idea to display folders in a different color.

### What did you see instead?

Folders for which I don't have permission (mostly OS shortcuts) not only show on the list, but they are clickable as folders, whereupon performous gives an iterator error, correctly saying permission denied, but then proceeds to change into the folder anyway. Not sure if this could then cause further breakage, with the game thinking it's inside a folder it doesn't actually have access to.

```
(paste your output here)
```

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```
(paste your output here)
```

Baklap4:
### What does this PR do?

Uses the [Directory Options](https://en.cppreference.com/w/cpp/filesystem/directory_options) to skip directories we don't have permission for without throwing an exception.

**This doesn't remove the entry from the directory list though.**

### Closes Issue(s)

Partly Closes #700

### DIFF
--- a/game/screen_paths.cc
+++ b/game/screen_paths.cc
@@ -98,7 +98,7 @@ void ScreenPaths::generateMenuFromPath(fs::path path) {
 	
 	// Extract list of all directories
 	std::list<fs::path> directories;
-	for (const auto &di : fs::directory_iterator(path)) {
+	for (const auto &di : fs::directory_iterator(path, fs::directory_options::skip_permission_denied)) {
 		auto &p = di.path();
 		if (fs::is_directory(p) && (showHiddenfolders || p.filename().c_str()[0] != '.')) {
 			directories.emplace_back(p);


yoda-jm:
Can we catch the specific error while iterating and display a nice message into the console (something alike):
WARNING: cannot browse folder XXXX (reasong: e.what())

Baklap4:
Seems like a good suggestion. Will add it in a bit

Baklap4:
Might be even better to rework this screen as a whole. Like in the opening of this PR stated, The user can still end up in a directory for which he doesn't have permission. A quick flash message shows the error popping up. But after that the screen remains blank. This is not a good user experience.

Ideally we want the following behavior:
1. Start in $HOME folder
2. Filter out all directories which gives permission errors
    - This is the tricky part, as we can only determine it by going into the folder. Reading permissions/user/group is not a thing? 
3. Show directory tree to user

This way the user will never be able to accidentally end up in an sort of empty screen and we also did 'tackle' the permission error.

Baklap4:
Converting to draft since the scope of this has changed and stuff needs to get changed to make this work properly.

game/screen_paths.cc
```diff
@@ -98,7 +98,7 @@ void ScreenPaths::generateMenuFromPath(fs::path path) {
 	
 	// Extract list of all directories
 	std::list<fs::path> directories;
-	for (const auto &di : fs::directory_iterator(path)) {
+	for (const auto &di : fs::directory_iterator(path, fs::directory_options::skip_permission_denied)) {
```
OznOg, line 101: seems that this option interrupts the iterating; passing an error code seems to allow to continue iterating

Baklap4:
Closing this so it can be explained more within the issue #700

PR #713, opened 2022-04-03
Nlohmann migration

yoda-jm:
### What does this PR do?

Refactor to use nlohmann::json instead of cpprest json library for songs cache

### Motivation

nlohmann::json is standalone hpp file, this will also help when we want to move to an http server lib that is header only.


game/songs.cc
```diff
@@ -82,181 +87,160 @@ void Songs::reload_internal() {
 	doneLoading = true;
 }
 
-#ifdef USE_WEBSERVER
+const std::string SONGS_CACHE_JSON_FILE = "songs.json";
+
 void Songs::LoadCache() {
-	fs::path songsMetaFile = getCacheDir() / "Songs-Metadata.json";
+	const fs::path songsMetaFile = getCacheDir() / SONGS_CACHE_JSON_FILE;
```
Lord-Kamina, line 93: I'd prefer if you put the JSON read/parse code into its own function, like I did in #545 (plus some comments made by @Baklap4 in there, re: throwing and closing the file, that I just realized I never addressed)

Lord-Kamina (changes requested):
Also, I really don't like just adding json.hpp onto the project like that, especially not under the `game` folder.

game/songs.cc
yoda-jm: The goal is to keep the code as much as possible as it was previously:
- to ease review (code change is difficult to check)
- to stay in line with PR goad (it is just changing the lib used to parse and produce the json songs cache.

About the placement of json.hpp please give me a place where to put it. I think in the game folder it is the best since:
- it is only used for the game
- most discribution do not provide a package for this header
- it enables a nice include : #include "nlohmann/json.hpp"

game/song.cc
```diff
@@ -12,62 +12,61 @@ extern "C" {
...
+	artist = song.contains("artist") ? song.at("artist").get<std::string>() : "";
+	title = song.contains("title") ? song.at("title").get<std::string>() : "";
+	language = song.contains("language") ? song.at("language").get<std::string>() : "";
+	edition = song.contains("edition") ? song.at("edition").get<std::string>() : "";
+	creator = song.contains("creator") ? song.at("creator").get<std::string>() : "";
+	genre = song.contains("genre") ? song.at("genre").get<std::string>() : "";
+	cover = song.contains("cover") ? song.at("cover").get<std::string>() : "";
+	background = song.contains("background") ? song.at("background").get<std::string>() : "";
+	video = song.contains("videoFile") ? fs::path(song.at("videoFile").get<std::string>()) : "";
+	midifilename = song.contains("midFile") ? fs::path(song.at("midFile").get<std::string>()) : "";
```
Baklap4, line 28: i think this key is, and was, a typo to begin with: `midFile` -> `midiFile`

game/songs.cc
Baklap4: I think @Lord-Kamina has a valid point, but so dos @yoda-jm in terms of the JSON read/parse.
For now let's just plain rewrite it and let's (@Lord-Kamina can you do this?) make an issue to improve/refactor the json capabilities like discussed in #545

> most discribution do not provide a package for this header

Not sure where you did get this info but almost every major system does have it: ubuntu, debian, alpine, arch, freebsd, vcpkg: https://pkgs.org/search/?q=nlohmann
So i think we can depend on the package being installed by the packagemanager. We will need a cmake module for it though.

```
# CMakeLists.txt
find_package(nlohmann_json 3.2.0 REQUIRED)
...
add_library(foo ...)
...
target_link_libraries(foo PRIVATE nlohmann_json::nlohmann_json)
```

game/songs.cc
```diff
@@ -82,181 +87,160 @@ void Songs::reload_internal() {
...
+			songObject["cover"] = song->cover.string();
+		}
+		if(!song->background.string().empty()) {
+			songObject["background"] = song->background.string();
+		}
+		if(!song->music[TrackName::BGMUSIC].string().empty()) {
+			songObject["songFile"] = song->music[TrackName::BGMUSIC].string();
+		}
+		if(!song->midifilename.string().empty()) {
+			songObject["midFile"] = song->midifilename.string();
```
Baklap4, line 173: `midFile` typo -> `midiFile`

game/songs.cc
```diff
@@ -82,181 +87,160 @@ void Songs::reload_internal() {
...
-	fs::path cacheDir = getCacheDir() / "Songs-Metadata.json";
+	fs::path cacheDir = getCacheDir() / SONGS_CACHE_JSON_FILE;
 
 	try {
-		std::stringstream stream;
-		jsonRoot.serialize(stream);
 		std::ofstream outFile(cacheDir.string());
-    	outFile << stream.rdbuf();
-    	outFile.close();
+		outFile << jsonRoot.dump(4);
```
Baklap4, line 237: Define the 'magic number' by using a const variable.
`const int spacesCount = 4` or a better name.

Baklap4 (changes requested):

game/songs.cc
Baklap4: I think if you add a cmake module and add the dependency to the list of dependencies within game/cmakelists.txt#103
we don't actually have to include the version as it'll probably work with nlohmann3 and nlohmann2

The module should look something like this i think:
FindJson.cmake
```
# - Try to find nlohmann_json
# Once done, this will define
#
#  Json_FOUND - system has nlohmann_json
#  Json_INCLUDE_DIRS - the nlohmann_json include directories
#  Json_LIBRARIES - link these to use nlohmann_json
#
# See documentation on how to write CMake scripts at
# http://www.cmake.org/Wiki/CMake:How_To_Find_Libraries

include(LibFindMacros)
libfind_pkg_detect(Json nlohmann_json FIND_PATH nlohmann/json.hpp FIND_LIBRARY nlohmann_json)
set(Json_VERSION ${Json_PKGCONF_VERSION})
libfind_process(Json)
```

Baklap4: outstanding issue at @Tronic LibFindMacros.cmake > https://github.com/Tronic/cmake-modules/issues/5

Lord-Kamina:
@yoda-jm could you add a fallback logic similar to what we've got in https://github.com/performous/performous/blob/master/cmake/Modules/FindAubio.cmake ?

cmake/Modules/FindJson.cmake
```diff
@@ -0,0 +1,15 @@
...
+#  Json_LIBRARIES - link these to use nlohmann_json
+#
+# See documentation on how to write CMake scripts at
+# http://www.cmake.org/Wiki/CMake:How_To_Find_Libraries
+
+include(LibFindMacros)
+
+libfind_pkg_detect(Json nlohmann_json FIND_PATH nlohmann/json.hpp)
+set(Json_VERSION ${Json_PKGCONF_VERSION})
+libfind_process(Json)
```
OznOg, line 15: I don't really get the point of this file. why not simply use `find_package(nlohmann_json 3.10.5 REQUIRED)`

Baklap4: This is why we use our own modules instead of a 'simple' find_package call indeed :)

game/song.cc
```diff
@@ -12,62 +12,61 @@ extern "C" {
...
-	music[TrackName::PREVIEW] = song.has_field("Preview") ? fs::path(song.at("Preview").as_string()) : "";
-	music[TrackName::GUITAR] = song.has_field("Guitar") ? fs::path(song.at("Guitar").as_string()) : "";
-	music[TrackName::BASS] = song.has_field("Bass") ? fs::path(song.at("Bass").as_string()) : "";
-	music[TrackName::DRUMS] = song.has_field("Drums") ? fs::path(song.at("Drums").as_string()) : "";
-	music[TrackName::KEYBOARD] = song.has_field("Keyboard") ? fs::path(song.at("Keyboard").as_string()) : "";
-	music[TrackName::GUITAR_COOP] = song.has_field("Guitar_coop") ? fs::path(song.at("Guitar_coop").as_string()) : "";
-	music[TrackName::GUITAR_RHYTHM] = song.has_field("Guitar_rhythm") ? fs::path(song.at("Guitar_rhythm").as_string()) : "";
+
+Song::Song(nlohmann::json const& song): dummyVocal(TrackName::LEAD_VOCAL), randomIdx(rand()) {
+	path = song.count("txtFileFolder") > 0 ? fs::path(song.at("txtFileFolder").get<std::string>()) : "";
```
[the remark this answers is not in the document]
Baklap4: Seems like a good suggestion :)

cmake/Modules/FindJson.cmake
OznOg: ok

game/song.cc
```diff
@@ -7,67 +7,70 @@
...
+	}
+
+	if (getJsonEntry<bool>(song, "keyboardTracks").value_or(false)) {
+		instrumentTracks.insert(make_pair(TrackName::KEYBOARD, InstrumentTrack(TrackName::KEYBOARD)));
 	}
-	
-	if (song.has_field("KeyboardTracks")) {
-			instrumentTracks.insert(make_pair(TrackName::KEYBOARD, InstrumentTrack(TrackName::KEYBOARD)));
+
+	if (getJsonEntry<bool>(song, "drumTracks").value_or(false)) {
```
OznOg, line 59: careful field was "DrumTracks" and now is "drumTracks"; Moreover, are you sure this is  boolean value? shouldn't it be :
`if (getJsonEntry<std::string>(song, "drumTracks"))` ?

yoda-jm: This was changed in a previous commit for the following reasons:
- use camlCase json names (compliant with Google recommandations)
- use boolean value instead of int for this field since we only encode a boolean.

OznOg: ok, my bad

yoda-jm:
> @yoda-jm could you add a fallback logic similar to what we've got in https://github.com/performous/performous/blob/master/cmake/Modules/FindAubio.cmake ?
> 
> Edit: be sure to use `GIT_SHALLOW` because the nlohmann repo has >200 mb worth of examples and tests

done

yoda-jm:
> Also, I really don't like just adding json.hpp onto the project like that, especially not under the `game` folder.

done

Lord-Kamina (approved):
LGTM

Baklap4 (approved):
LGTM

PR #714, opened 2022-04-04
Migrates boost-format to fmt

Baklap4:
### What does this PR do?
Removed boost-format added fmt.
Updated the various places to use fmt
Updated CI to use `libfmt-dev`

### Closes Issue(s)

Closes partly #39 

~~Depends on https://github.com/performous/performous-docker/pull/20~~ Has been merged :)

### Motivation

libfmt is way faster than boost is and the syntax is quite a bit nicer.
Also removing boost completely has been on our roadmap, since it's a very big dependency.

### Additional Notes

Do take a special look to the `configuration.cc` -> `numericFormat` template method. I'm not quite sure if i did the right conversion. During runtime i didn't see any problems.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -21,8 +21,7 @@ sudo apt-get install git-core cmake build-essential gettext help2man \
    libepoxy-dev libsdl2-dev libcairo2-dev libpango1.0-dev librsvg2-dev \
    libboost-all-dev libavcodec-dev libavformat-dev libswscale-dev libswresample-dev \
    libpng-dev libjpeg-dev libxml++2.6-dev portaudio19-dev \
-   libopencv-dev libportmidi-dev libcppnetlib-dev libjsoncpp-dev
-   nlohmann-json3-dev
+   libopencv-dev libportmidi-dev libcpprest-dev nlohmann-json3-dev libfmt-dev
 ```
 
 Notice: Dependency problems may prevent installation of portaudio19-dev. At least with Ubuntu 13.04 this can be solved by first installing libjack-jackd2-dev, even though that package is not actually needed for Performous.
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -227,7 +227,7 @@ jobs:
           chmod +x appimage-builder-x86_64.AppImage
           sudo mv appimage-builder-x86_64.AppImage /usr/local/bin/appimage-builder
           sudo apt update
-          sudo apt-get install -y --no-install-recommends git cmake build-essential gettext help2man libopenblas-dev libfftw3-dev libicu-dev libepoxy-dev libsdl2-dev libfreetype6-dev libpango1.0-dev librsvg2-dev libxml++2.6-dev libavcodec-dev libavformat-dev libswscale-dev libjpeg-dev portaudio19-dev libglm-dev libboost-filesystem-dev libboost-iostreams-dev libboost-locale-dev libboost-system-dev libboost-program-options-dev libssl-dev libcpprest-dev libportmidi-dev libopencv-dev libaubio-dev nlohmann-json3-dev
+          sudo apt-get install -y --no-install-recommends git cmake build-essential gettext help2man libopenblas-dev libfftw3-dev libicu-dev libepoxy-dev libsdl2-dev libfreetype6-dev libpango1.0-dev librsvg2-dev libxml++2.6-dev libavcodec-dev libavformat-dev libswscale-dev libjpeg-dev portaudio19-dev libglm-dev libboost-filesystem-dev libboost-iostreams-dev libboost-locale-dev libboost-system-dev libboost-program-options-dev libssl-dev libcpprest-dev libportmidi-dev libopencv-dev libaubio-dev nlohmann-json3-dev libfmt-dev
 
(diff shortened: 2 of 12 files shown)


yoda-jm:
Don't know the current code compatibility we want, but C++ 20 std::format could also be nice (see https://en.cppreference.com/w/cpp/utility/format/format)

Baklap4:
Currently we're using c++14 if i'm correct
Besides fmt uses c++20 when available

yoda-jm:
Since we are already using boost heavily (like everywhere), I am not sure this change worse being done.
The next step however is if we want to remove a boost usage to go for an std variant (this is possible if we bump minimum C++ requireent from C++17 to C++20).

Lord-Kamina:
If we do end-up using RESTinio, that also uses fmt. Just a heads up.


RE: Removing boost. I'm pretty sure that's not gonna happen anytime soon. We depend on it for localization and I think the ship's already sailed for the STL on that. I am pretty sure at some point I read that boost::locale was NOT going to ever be merged.

Baklap4:
> Since we are already using boost heavily (like everywhere)

That's kinda exaggerated. We are only using a couple of boost modules. The amount we use boost has been drastically decreased since 1.1. See https://github.com/performous/performous/issues/39#issuecomment-429954430 for what's already been ticked off and removed

Ultimatum we'd only have to use boost for just the localization stuff and program options since there's no good equivalent for those yet.

The current installed packages are: `libboost-program-options-dev` `libboost-iostreams-dev` `libboost-system-dev` `libboost-locale-dev`

yoda-jm:
> > Since we are already using boost heavily (like everywhere)
> 
> That's kinda exaggerated. We are only using a couple of boost modules. The amount we use boost has been drastically decreased since 1.1. See [#39 (comment)](https://github.com/performous/performous/issues/39#issuecomment-429954430) for what's already been ticked off and removed
> 
> Ultimatum we'd only have to use boost for just the localization stuff and program options since there's no good equivalent for those yet.
> 
> The current installed packages are: `libboost-program-options-dev` `libboost-iostreams-dev` `libboost-system-dev` `libboost-locale-dev`

Here is the list of stuff we should try to remove:

$ git grep boost | grep include | cut -d: -f2- | sort | uniq
#include <boost/algorithm/string/case_conv.hpp>
#include <boost/algorithm/string/classification.hpp>
#include <boost/algorithm/string.hpp>
#include <boost/algorithm/string/replace.hpp>
#include <boost/asio.hpp>
#include <boost/filesystem.hpp>
#include <boost/format.hpp>
#include <boost/iostreams/device/file_descriptor.hpp>
#include <boost/iostreams/stream.hpp>
#include <boost/locale.hpp>
#include <boost/predef/os.h>
#include <boost/program_options.hpp>
#include <boost/range/adaptor/reversed.hpp>
#include <boost/range.hpp>
#include <boost/range/iterator_range.hpp>

Baklap4:
@yoda-jm better to move the boost deps discussion to #39

Baklap4:
Converting to draft since it needs to be updated and some stuff has to be addressed.

yoda-jm:
Can you revert the change done by adding CRLF to end of line (dos2unix / unix2dos).
I think we should stay with the Unix end of line format.

Baklap4:
Blocked by #762 but otherwise good for review

yoda-jm (approved):
LGTM, if anybody wants to double-check

ooshlablu (changes requested):
While testing, the game runs ok until I launch some (but not all) songs, which freezes the screen (but still continues to play the music). `ERROR: cannot switch from automatic to manual argument indexing` is constantly dumped to the terminal that the process is run in (but not to the logs). I can reproduce this behavior with both the AppImage and deb packages on Ubuntu 18.04 & 20.04.

cmake/Modules/FindFmt.cmake
```diff
@@ -0,0 +1,15 @@
+# - Try to find Fmt
+# Once done, this will define
+#
+#  Fmt_FOUND - system has fmt
+#  Fmt_INCLUDE_DIRS - the fmt include directories
+#  Fmt_LIBRARIES - link these to use fmt
```
OznOg, line 6: I still dislike detecting dependencies like this without using the canonical way of doing it in modern cmake.The imported target should be fmt::fmt, and all this file be one line...
As I seem to be the only one with this opinion, I will disagree and commit.

OznOg (dismissed):

ooshlablu:
Just tested again with the latest build and the error I was seeing before is gone. Works in both AppImage and debs on Ubuntu 18.04 & 20.04

cmake/Modules/FindFmt.cmake
Baklap4: How would you like to see this find module?

ooshlablu (approved):
CI/CD items LGTM

cmake/Modules/FindFmt.cmake
Baklap4: Resolved as discussed on discord. Opened https://github.com/performous/performous/issues/767 added to resolve the other issues regarding dependencies

OznOg (approved):

ooshlablu:
Tested again earlier, didn't notice any changes since the last commit. :+1:

Issue #514, opened 2020-03-27
Code cleanup/style


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Question/feature

### What did you expect to see?

I expected some clean code all written in the same style.

This issue is open to discussion. Let's try to get to a point where we can say: **let's do performous this way.**

# Suggestions
1. Make/use/find code conventions for latest c++
2. Use a tool to force these conventions on us, popular tools are; editorconfig, [cpplint](https://github.com/cpplint/cpplint)
3. Use these tools within the CI
4. Treat warning as errors, forced upon us by cmakelist
5. Tabs vs spaces, spaces vs tabs
6. .. some other stuff? 

LEt's see what everyone wants to bring in :)

**Again this is open to discussion**

@performous/core-owners @performous/contributors

OznOg:
yes unifying would be a good idea; as tools there is also uncrustify, cang-format (indent).
I'm not a great fan of tabs, but seems to be quite "standard" in the code, so let's not mess everything up just for spacings.
-Wall -Wextra -Wall -Werror should already be fine to have.
The code may also be passed to coverity, this is available online for opensource projects, the tool may find some relevant problems.
One really nice to have would be unit tests and/or validation tests, so that we are more confident with the code being pushed.

Tronic:
Performous coding style: https://github.com/performous/performous/wiki/Coding-style

I would avoid all tools that try to make lines of particular length because they lose all the thought that has been put into manual formatting. In particular, I dislike splitting `if` and what follows it like `return something` on separate lines, especially when many similar ifs follow each other because the interleaved conditions and responses are far harder to read than almost identical lines following each other, and if there are no curly braces (which then add more noise) it is also quite easy to mess up.

Also worth a notice is that all auto-indent tools have been more or less broken by design. Most of them think that a tab is equivalent to some integer number of spaces, which is just wrong (my editor is set to use tab width of π columns to avoid such misconception). Also, none of them know of the *small indent* (two spaces) convention used for splitting long lines (previously also `case`, `public`, `private` but because of tool issues we no longer indent them).

Overall, I find Performous code style quite consistent, and thus see only marginal use for extra linting, especially since compilers are quite good at finding issues nowadays. Enforcing warning-free compiles, consistent indentation, removal of end-of-line whitespace and other such things in CI couldn't hurt, as long as the tools don't break existing code and practices.

twollgam:
[...]

PR #715, opened 2022-04-05
GitHub actions

ooshlablu:
### What does this PR do?
Adds GitHub actions to automate building PR's, master and rc branches and tags.
Fixes an issue with the `FindSDL2` cmake module where it found debug library in release mode.
Fixes versioning of our binaries. Versioning scheme can be found below.
Adds artifacts to PR's
Adds anonymous downloadable artifacts through our README.md with the help of [Nightly.Link](https://github.com/oprypin/nightly.link)

#### Workflows + Artifacts

- PR workflow: https://github.com/ooshlablu/performous/actions/runs/2177182680
    - Run is linked to this PR: https://github.com/ooshlablu/performous/pull/10 where also the artifacts appeared.
- Master workflow: https://github.com/ooshlablu/performous/actions/runs/2177243726
- Tag workflow: https://github.com/ooshlablu/performous/actions/runs/2177795224
    - Release that's been made automaticly with all attached artifacts: https://github.com/ooshlablu/performous/releases/tag/6.7.8

#### PR's, Master, RC-branches and Tags

- Automates installing dependencies
- Automates building for:
    - Ubuntu 1804
    - Ubuntu 2004
    - Ubuntu 2204
    - Fedora 33
    - Fedora 34
    - Fedora 35
    - Mac OSX
    - Windows
- Generates artifacts for the above mentioned platforms with the following version syntax:
    - PR's: 
        - `Performous-<Major>.<Minor>.<Patch>-<PR_Number>-<Short_SHA>-alpha`
            - `Performous-1.2.0-715-0883ada-alpha`
    - Merge to master: 
        - `Performous-<Major>.<Minor>.<Patch>-<Nr_Of_Commits_Since_last_release>-<Short_SHA>-beta`
            - `Performous-1.2.0-40-0883ada-beta`
        - `Performous-latest`
            - `Performous-latest`
    - Tags: 
        - `Performous-<Major>.<Minor>.<Patch>`
            - `Performous-1.2.1`
- Upon tagging a Github Release is automaticly made with the name `Performous <tag>`. Since we're tagging in semver notation this will become: `Performous <Major>.<Minor>.<Patch>` -> `Performous 1.2.3`
- Within the README.md you can always download the Latest version for each OS.
    - Note: At this moment GitHub has issues with Workflow filtering. Due to this Nightly might be unable to correcty identify the latest succesful build. For more info check these links:
        - https://github.com/oprypin/nightly.link/issues/26#issuecomment-1097162423
[...]

PR #717, opened 2022-04-06
Add git hooks

twollgam:
### What does this PR do?

It adds some git hooks devided in necessary and optional.

The necessary scripts check if tabs instead spaces are used and there
 is no trailing whitespace.

The optional scripts autocorrect tabs and trailing whitespace.

### Closes Issue(s)

Partly closes: https://github.com/performous/performous/issues/514

### Motivation

The hooks help to commit correctly styled code.

### Additional Notes

Actually there is only a powershell install script, which also work on linux with pwsh.

### Tasks

- [ ] Missing extensions: xsl, frag, vert, geom, yml, cmake, config, hpp, inc, ebuild


twollgam:
If this PR is merged I will update the style guide in the wiki how to install the git hooks and what they do.

misc/git-hook/install.ps1
```diff
@@ -0,0 +1,40 @@
+param([switch]$all, [switch]$clear)
+
+$gitDir = git rev-parse --git-commom-dir
```
Baklap4, line 3: typo: `--git-common-dir`

Baklap4 (changes requested):
Will test these hooks later today

Baklap4:
Hmmm i'm having double feelings about the githooks.
It requires the developer to install these into the repository as it doesn't get configured upon clone the repository.
After install it blocks and/or modifies your commit if it doesn't agree with the style itself.

This is a good thing in general, however like i pointed it it requires an installation/configuration from the developer.
However the actual issue is a 'server' problem. The server (which can also be described in other words: performous-core owners) want to enforce a specific styling. Therefore it is/can become a server problem which the server should resolve.

I'd suggest to wait with this PR until we have #715 merged. And then add a style-checker job. If the issue can be automaticly resolved a bot could create this style-fixing-commit with the approach you already made in these scripts. If the issue cannot be automaticly resolved, the developer should add another commit resolving the issues. This is indicated by a failing github action which need to be resolved eitherway.

To aid the developer to not make these mistakes in the first place, we still can suggest to install these commit hooks

@twollgam what do you think about that approach?

twollgam:
@Baklap4 I agree what belongs the server side. Not to offer server-side hooks may be a decision  for security reasons but it denies solutions which could reject commits. If we can make a server-side solution I would prefer this.

A solution that only checks and remarks false style is not what I expect. Those checks must run early. I do not want something where I have to do "extra style commits".

I use the hooks in this PR and they are not perfect yet but I like to be warned and I like the "auto correction" ;-)

BTW: once installed, the hooks update themself from repo.

Baklap4:
> Not to offer server-side hooks may be a decision for security reasons but it denies solutions which could reject commits. 

That's correct it's for security reasons github doesn't want you to execute 'random' code on their servers ;)

> A solution that only checks and remarks false style is not what I expect. Those checks must run early. I do not want something where I have to do "extra style commits".

Well in a way this is the same as when it rejects it locally. The gh-actions would fail, so the developer needs to do something to fix it. This can be done by either amending the relevant commits and force-pushing _or_ by creating a last commit 'cleanup'. Which is in a way pretty much the same locally: it rejects your commit, you have to make changes and then commit the change

Like i said before the installing of hooks require the developer to do something, it'd be encouraged, but if not installed the ci/cd will tell you to clean up your "mess"

yoda-jm:
The scripts are nice, but I think I would have put them in a workflow preventing merge if the files your are committing contains illegal code

Baklap4:
> The scripts are nice, but I think I would have put them in a workflow preventing merge if the files your are committing contains illegal code

Best would be to have them on either side. So within the workflow and installed in your local repository.

@twollgam can you rebase this branch and try adding the checks to the workflow so we check it on both sides (server and client)? If the check is not succesful the action will result in a failure, and by making it a required action we prevent it from getting merged.

misc/git-hook/necessary/pre-commit
```diff
@@ -0,0 +1,75 @@
...
+if [ -f "./functions.bash" ]
+then
+    . ./functions.bash
+else
+    . $GITDIR/hooks/functions.bash
+fi
+
+echo ">>>>>>>> modifying begin <<<<<<<<"
+
+files=`find $GITDIR/hooks/precommit/modify/* -type f -maxdepth 0 -mindepth 0`
```
Baklap4, line 23: This line gives a 'no such file or directory' issue when the optional tools are not installed.
image

misc/git-hook/necessary/precommit/check/indent
```diff
@@ -0,0 +1,38 @@
...
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+    against=HEAD
+else
+    # Initial commit: diff against an empty tree object
+    against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+# We should check only added or modified C/C++ source files  for tabs.
+changed_files=$(git diff-index --cached $against | grep -E '[MA]	.*\.(c|cpp|cc|cxx|h|hh)$' | cut -d'	' -f 2)
```
Baklap4, line 14: Please include other text based files used within the repo

misc/git-hook/necessary/precommit/check/lineendings
```diff
@@ -0,0 +1,37 @@
...
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+    against=HEAD
+else
+    # Initial commit: diff against an empty tree object
+    against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+# We should check only added or modified files of these given types for line endings.
+changed_files=$(git diff-index --cached $against | grep -E '([MA]	.*\.(c|cpp|cc|cxx|h|hh)$)' | cut -f 2)
```
Baklap4, line 14: Please include other text based files in the repo

misc/git-hook/optional/precommit/modify/remove_trailing_spaces
```diff
@@ -0,0 +1,34 @@
...
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+    against=HEAD
+else
+    # Initial commit: diff against an empty tree object
+    against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+changed_files=$(git diff-index --cached $against | grep -i -E '[MA]	.*\.(c|cc|cpp|h|hh)$' | cut -d'	' -f 2)
```
Baklap4, line 23: Please include other text-based files in the repo

misc/git-hook/optional/precommit/modify/spaces2tabs
```diff
@@ -0,0 +1,34 @@
...
+
+if git rev-parse --verify HEAD >/dev/null 2>&1
+then
+    against=HEAD
+else
+    # Initial commit: diff against an empty tree object
+    against=4b825dc642cb6eb9a060e54bf8d69288fbee4904
+fi
+
+changed_files=$(git diff-index --cached $against | grep -i -E '[MA]	.*\.(c|cc|cpp|h|hh)$' | cut -d'	' -f 2)
```
Baklap4, line 23: Please include other text-based files in the repo.

Baklap4 (changes requested):
Hmm after installing the hooks (with option 'all' defined) i saw it didn't work with other text based files due to the extension constraint. Please add the following to the list:
- .md
- .html
- .css
- .js
- .xml
- .json
- .txt
- .desktop
- .sh
- .ps1
- .bash
- other text based files

misc/git-hook/install.ps1
```diff
@@ -0,0 +1,40 @@
...
+
+$gitHooksDir = join-path $gitDir "hooks"
+
+if (!(test-path $gitHooksDir)) {
+    throw "$gitHooksDir not present"
+}
+
+if($clear) {
+    write-host "cleaning old hooks dir: $githooksdir"
+    Remove-Item (join-path $githooksdir "*") -force -recurse
```
Baklap4, line 24: If clear is given, exit out of here.

Baklap4, line 24: Also clear removes all the 'default sample git hooks'. Might be wise to not remove those

twollgam:
> Hmm after installing the hooks (with option 'all' defined) i saw it didn't work with other text based files due to the extension constraint. Please add the following to the list:
> 
> * .md
> * .html
> * .css
> * .js
> * .xml
> * .json
> * .txt
> * .desktop
> * .sh
> * .ps1
> * .bash
> * other text based files

@Baklap4 what would you do expect to be checked for this files? No white space at end of lines - ok, but other formatting like tabs too? I split the checks into seperate scripts so that checks can be done for different types of files.

Baklap4:
> @Baklap4 what would you do expect to be checked for this files? No white space at end of lines - ok, but other formatting like tabs too? I split the checks into seperate scripts so that checks can be done for different types of files.

Yeah both can done and 4 spaces/1tab seems pretty much the norm out there anyway regardless of what kind of textfile there is

misc/git-hook/necessary/precommit/check/lineendings.bash
```diff
@@ -0,0 +1,28 @@
...
+
+IFS=$'\n' # bash specific
+for file in $changed_files
+do
+	result=`cat "$file" | grep -E -n '[[:space:]]$'`
+
+	if [ -n "$result" ]
+	then
+		echo "file with line ending issue: $file"
+		echo "$result"
```
Baklap4, line 16: Might be wise to not echo this. As large files tend to clutter the output a lot

misc/git-hook/necessary/precommit/check/indent.bash
```diff
@@ -0,0 +1,28 @@
...
+
+IFS=$'\n' # bash specific
+for file in $changed_files
+do
+	result=`cat "$file" | grep -P -n '^[\t]*(    | {1,3}\t)'`
+
+	if [ -n "$result" ]
+	then
+		echo "file with indent issue: $file"
+		echo "$result"
```
Baklap4, line 16: Might be wise to not output this as this can become quite noisy in log output.

misc/git-hook/optional/precommit/modify/spaces2tabs.bash
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
Baklap4, line 1: Hmm not sure how i didn't notice this before.

I know i said i prefer tabs before. But then i saw that different browsers show the git diff differently when tabs or spaces are used.
Which leads me to think it'd be nice to conserve spaces instead of tabs.
This is due to the fact that tabs are rendered on different widths in different programs whilst spaces have a fixed width.

Within the editor of your choice you can set to show spaces as tabs if it meets a certain value.

After installing these scripts we'll run into problems when converting from tabs to spaces for the files you've changed. But i think that's alright. Not sure if @OznOg or @Lord-Kamina have any concerns regarding this?

Baklap4 (changes requested):

misc/git-hook/optional/precommit/modify/spaces2tabs.bash
Baklap4: Majority of the code base is tabs if that makes for an easier decision on this

misc/git-hook/necessary/precommit/check/indent.bash
twollgam: OK, but in common you want to know which lines has issues and must be fixed. If the codebase is without issues the noise should be low.

twollgam: Should we limit the output? 8 issues per file for example?

misc/git-hook/necessary/functions.bash
```diff
@@ -0,0 +1,95 @@
+getTextFileExtensions()
+{
+    echo "c|cpp|cc|cxx|h|hh|ps1|psm1|bash|sh|svg|xml|md|html|css|js|json|txt|desktop"
+}
+
+getSourceFileExtensions()
+{
+    echo "c|cpp|cc|cxx|h|hh|ps1|psm1|bash|sh|js|"
+}
```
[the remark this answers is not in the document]
twollgam: Removed it.

misc/git-hook/necessary/precommit/check/lineendings.bash
twollgam: See next comment.

Baklap4:
stumbling upon this old pr again. Within the format project we use pre-commit. This project already defines a lot of standard commit hooks which we can integrate. Maybe that's nice to do aswell? https://github.com/UltraStar-Deluxe/format/blob/main/.pre-commit-config.yaml

twollgam:
@Baklap4 I removed the modifying hooks and I added an install script for bash. I think we should start with and expand the hooks if needed.

Baklap4 (approved):
LGTM , we might wanna check if these work on the different OS'ses?

twollgam:
I have checked on ubuntu and win 11. So the bash install script and the hooks should work in every git environment.

PR #720, opened 2022-04-12
Deps build logic

yoda-jm:
### What does this PR do?

Make it clearer when we want to fetch external source code and build them when compiling Performous.
The following things will be done also later in the PR:
- try to display a message advertising for the usage of "ALLOW_*" flags when lib is not found,
- update the semaphore configuration to match this new logic (put the allow only when required).

### Motivation

I always prefer explicitly specifying what I want to do (and helping people do this when possible)


game/CMakeLists.txt
```diff
@@ -99,6 +99,17 @@ target_link_libraries(performous PRIVATE ${LibEpoxy_LIBRARIES})
...
+option(ALLOW_SELF_BUILT_AUBIO "allow downloading and building aubio from sources if not available on system" FALSE)
+option(FORCE_SELF_BUILT_AUBIO "force downloading and building aubio from sources" FALSE)
+find_package(Aubio 0.4.9 REQUIRED)
+message(STATUS "Aubio includes: ${Aubio_INCLUDE_DIRS}")
+target_include_directories(performous SYSTEM PRIVATE ${Aubio_INCLUDE_DIRS})
+target_link_libraries(performous PRIVATE ${Aubio_LIBRARIES})
+add_definitions(${Aubio_DEFINITIONS})
+
+option(ALLOW_SELF_BUILT_JSON "allow downloading and building nlohmann_json from sources if not available on system" FALSE)
+option(FORCE_SELF_BUILT_JSON "force downloading and building nlohmann_json from sources" FALSE)
```
OznOg, line 111: Thise 2 boolean seems to conflict each other, what would mean ALLOW=FALSE and FORCE=TRUE? Maybe would be better to use a single variable wit 3 values (SEFL_BUILD_XXX={NEVER,AUTO,ALWAYS}?

game/CMakeLists.txt
```diff
@@ -99,6 +99,17 @@ target_link_libraries(performous PRIVATE ${LibEpoxy_LIBRARIES})
 find_package(PkgConfig REQUIRED)
 pkg_check_modules(deps REQUIRED IMPORTED_TARGET glib-2.0 gio-2.0 gobject-2.0)
 
+option(ALLOW_SELF_BUILT_AUBIO "allow downloading and building aubio from sources if not available on system" FALSE)
+option(FORCE_SELF_BUILT_AUBIO "force downloading and building aubio from sources" FALSE)
+find_package(Aubio 0.4.9 REQUIRED)
+message(STATUS "Aubio includes: ${Aubio_INCLUDE_DIRS}")
+target_include_directories(performous SYSTEM PRIVATE ${Aubio_INCLUDE_DIRS})
+target_link_libraries(performous PRIVATE ${Aubio_LIBRARIES})
```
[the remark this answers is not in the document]
Lord-Kamina: Cmake creates the variable. That's standard operation for cmake in module mode.

game/CMakeLists.txt
Baklap4: I'd agree with @OznOg here, 1 variable with 3 values

Baklap4 (changes requested):

cmake/Modules/FindAubio.cmake
```diff
@@ -1,21 +1,39 @@
...
-        # The version we look for is legacy, may need some fine tunning
-        pkg_check_modules(AUBIO IMPORTED_TARGET GLOBAL aubio>=0.4.9)
-            if(AUBIO_FOUND)
-                add_library(aubio ALIAS PkgConfig::AUBIO)
-            endif()
-    endif()
-endif()
+include(LibFindMacros)
+
+function(_buildAubioFromSource)
```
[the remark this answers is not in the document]
yoda-jm: this is now somewhat abstract (probably need some more work, but at least there is a ground to discuss it)

yoda-jm:
Things that could be nice:
- tag Aubio head (for example 0.4.9-performous or something similar), it would be nicer in the cmake result of performous when displaying the version used
- do the same for CED (there is no tag upstream, maybe something like 20211202-git ?)
- improve usage of libfetch_git_pkg (name, checking variables, ...)

cmake/Modules/LibFetchMacros.cmake
```diff
@@ -0,0 +1,41 @@
+#
+
+include(FetchContent)
+
+function (libfetch_git_pkg PREFIX)
```
Baklap4, line 5: Some form of documentation of what this function does would be nice for this.

game/CMakeLists.txt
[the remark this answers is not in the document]
yoda-jm: I choose to have the exact same logic as the loop under

cmake/Modules/LibFetchMacros.cmake
```diff
@@ -0,0 +1,60 @@
...
+# - ${PREFIX}_VERSION
+# - ${PREFIX}_INCLUDE_DIRS
+# - ${PREFIX}_FOUND
+# The following attribute are required:
+# - the PREFIX
+# - REPOSITORY <address>: the git repository address
+# - REFERENCE <reference>: a git reference for this repository (tag or hash or branch or ...
+# The following attribute are optional:
+# - FIND_PATH <header path>: the path of a file that should be located.
+# Eg: libfetch_git_pkg(Json NAME json REPOSITORY http://performous.org/json.git REFERENCE master FIND_PATH json/json.hpp)
```
Baklap4, line 14:
```suggestion
# Simple function to abstract fetching a dependency from Git.
# It produces similar output as what libfind_pkg_detect would return:
# - ${PREFIX}_VERSION
# - ${PREFIX}_INCLUDE_DIRS
# - ${PREFIX}_FOUND
# The following attribute are required:
# - PREFIX
# - REPOSITORY <address>: the git repository address
# - REFERENCE <reference>: a git reference for this repository (tag, branch, or hash)
# The following attributes are optional:
# - FIND_PATH <header path>: the path of a file that should be located.
# Eg: libfetch_git_pkg(Json NAME json REPOSITORY https://github.com/performous/json.git REFERENCE master FIND_PATH json/json.hpp)
```

cmake/Modules/LibFetchMacros.cmake
```diff
@@ -0,0 +1,60 @@
+include(FetchContent)
```
Baklap4, line 1: It's a best practice to set minimum versions for scripts. Since we're using `FetchContent` the minimum version should be 3.11.
```suggestion
cmake_minimum_required(VERSION 3.11)
include(FetchContent)
```

Baklap4 (commented):
A couple of typos and maybe nice to add a minimum cmake version. But i think that's all then :)

cmake/Modules/FindAubio.cmake
```diff
@@ -1,21 +1,36 @@
...
-    include(FetchContent)
-    FetchContent_Declare(aubio-sources
-      GIT_REPOSITORY https://github.com/performous/aubio.git
-      GIT_TAG        14fec3da6749fbcc47b56648d7a38296eccd9499
-      SOURCE_DIR aubio-src
-    )
-    FetchContent_MakeAvailable(aubio-sources)
+if(SELF_BUILT_AUBIO STREQUAL "ALWAYS")
+	message(STATUS "aubio forced to build from source")
+	libfetch_git_pkg(Aubio
```
OznOg, line 8: I don't really get the purpose of the libfetch_git_pkg function, basically you replace:
 ```
  FetchContent_Declare(aubio-sources
      GIT_REPOSITORY https://github.com/performous/aubio.git
      GIT_TAG        14fec3da6749fbcc47b56648d7a38296eccd9499
)
     FetchContent_MakeAvailable(aubio-sources)
```
by ```

libfetch_git_pkg(Aubio
        	REPOSITORY ${SELF_BUILT_GIT_BASE}/aubio.git
		REFERENCE  ${Aubio_GIT_VERSION}
	)
```
what is the expected benefit from it?

yoda-jm: The goal is:
- having a single way to have a library available (it mostly behave like our FindMacros package (returns the same kind of variables: found, includes, ...)
- hiding all the folder logic (for example the naming of the folder where it is downloaded and things like that),
- discover automatically the file (probably useful for headers only packages)
- hide the fact that FetchContent can fetch anything (this is especially crafter for git)
- avoid FetchContent_Declare/_MakeAvailable logic

.github/workflows/build_and_release.yml
```diff
@@ -127,6 +127,12 @@ jobs:
             EXTRA_CMAKE_ARGS="-DUSE_BOOST_REGEX=1"
           fi
 
+          if [[ "${ID}" == "ubuntu" && "${VERSION_ID}" == "18.04" ]]; then
+            # Ubuntu 18.04 has system Aubio 0.4.5, this is not enough
+            # because performous requires a minimum version of 0.4.9.
+            EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DSELF_BUILT_AUBIO=FORCE"
```
Baklap4, line 133: I guess this can be detected within the FindAubio cmake module right? The error message even gives you the version number it did find.

Doing a simple if there and then fall back to self-built seems appropiate

yoda-jm: I think I prefer not using AUTO as much as possible. If a specific target requires the "ALWAYS" target put it, else we should leave it to NEVER.
Other "AUTO" are currently only set until the dependency is added to the docker image (or dependency download)

yoda-jm:
Summary of what is used in workflows and why:
- CED (compact_enc_det): we always compile it from sources because it is not packaged for any distribution
- Json (nlohmann_json):
-- compiling from sources (AUTO) in Linux images (because it is not installed in the docker images)
-- compiling from sources (AUTO) in Mac image (because it is not installed by brew)
-- using system version in Windows image (the dependency is installed)
- Aubio
-- compiling from sources (ALWAYS) in Linux Ubuntu 18.04 (because version is too old there, but it is installed in docker image)
-- using system version for other Linux images
-- compiling from sources (ALWAYS) in Mac image (because linking with system library fails)
-- compiling from sources (AUTO) in Windows image (because dependency is not installed)

AUTO: means that updating depency should fix the issue (as soon as the image is updated, it should use the system version)
ALWAYS: means that there is something bigger to dig there (espacially Aubio/Mac)
NEVER: all the other cases, this is the default behavior: use system version

cmake/Modules/FindAubio.cmake
OznOg: I'm not really fan of such "syntaxic suggar" wrappers as it make life more complicated to people that know cmake and even to those who try to understand, but if I'm the only one to dislike it, go on...

cmake/Modules/FindAubio.cmake
```diff
@@ -1,21 +1,25 @@
...
+    FetchContent_Declare(aubio_from_source
       GIT_REPOSITORY https://github.com/performous/aubio.git
+      GIT_SHALLOW    TRUE
       GIT_TAG        14fec3da6749fbcc47b56648d7a38296eccd9499
       SOURCE_DIR aubio-src
     )
-    FetchContent_MakeAvailable(aubio-sources)
+    FetchContent_MakeAvailable(aubio_from_source)
+    set(Aubio_INCLUDE_DIRS ${aubio_from_source_SOURCE_DIR}/include/)
+    set(Aubio_LIBRARIES "-laubio")
```
OznOg: you should  not use variables for targets, this is really discouraged by cmake

yoda-jm: I revert back to former code style to detect Aubio system installation

Baklap4:
It seems to work properly here without changing stuff. That's good. However we do have a cmake warning added because of these changes:
```
CMake Warning (dev) at cmake/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
1> [CMake]   The package name passed to `find_package_handle_standard_args` (GLIB) does
1> [CMake]   not match the name of the calling package (Glib).  This can lead to
1> [CMake]   problems in calling code that expects `find_package` result variables
1> [CMake]   (e.g., `_FOUND`) to follow a certain pattern.
1> [CMake] Call Stack (most recent call first):
1> [CMake]   cmake/Modules/FindGlib.cmake:109 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
1> [CMake]   cmake/Modules/LibFindMacros.cmake:14 (find_package)
1> [CMake]   cmake/Modules/FindGlibmm.cmake:11 (libfind_package)
1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
1> [CMake]   cmake/Modules/LibFindMacros.cmake:14 (find_package)
1> [CMake]   cmake/Modules/FindLibXML++.cmake:12 (libfind_package)
1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
1> [CMake]   game/CMakeLists.txt:110 (find_package)
1> [CMake] This warning is for project developers.  Use -Wno-dev to suppress it.
1> [CMake] 
1> [CMake] -- Found GLIB: C:/Users/Arjan/source/repos/performous/performous2/build/x64-debug/vcpkg_installed/x64-windows/include/glib-2.0;C:/Users/Arjan/source/repos/performous/performous2/build/x64-debug/vcpkg_installed/x64-windows/debug/lib/glib-2.0/include (found version "2.70.1") 
```

yoda-jm:
> It seems to work properly here without changing stuff. That's good. However we do have a cmake warning added because of these changes:
> 
> ```
> CMake Warning (dev) at cmake/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
> 1> [CMake]   The package name passed to `find_package_handle_standard_args` (GLIB) does
> 1> [CMake]   not match the name of the calling package (Glib).  This can lead to
> 1> [CMake]   problems in calling code that expects `find_package` result variables
> 1> [CMake]   (e.g., `_FOUND`) to follow a certain pattern.
> 1> [CMake] Call Stack (most recent call first):
> 1> [CMake]   cmake/Modules/FindGlib.cmake:109 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
> 1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
> 1> [CMake]   cmake/Modules/LibFindMacros.cmake:14 (find_package)
> 1> [CMake]   cmake/Modules/FindGlibmm.cmake:11 (libfind_package)
> 1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
> 1> [CMake]   cmake/Modules/LibFindMacros.cmake:14 (find_package)
> 1> [CMake]   cmake/Modules/FindLibXML++.cmake:12 (libfind_package)
> 1> [CMake]   C:/Users/Arjan/source/repos/performous/vcpkg/scripts/buildsystems/vcpkg.cmake:805 (_find_package)
> 1> [CMake]   game/CMakeLists.txt:110 (find_package)
> 1> [CMake] This warning is for project developers.  Use -Wno-dev to suppress it.
> 1> [CMake] 
> 1> [CMake] -- Found GLIB: C:/Users/Arjan/source/repos/performous/performous2/build/x64-debug/vcpkg_installed/x64-windows/include/glib-2.0;C:/Users/Arjan/source/repos/performous/performous2/build/x64-debug/vcpkg_installed/x64-windows/debug/lib/glib-2.0/include (found version "2.70.1") 
> ```
> 
> This warning can also be found here: https://github.com/performous/performous/runs/6098083200?check_suite_focus=true#step:6:669

This seems to be linked to the removal of FetchContent_MakeAvailable(ced-sources) from top-level CMakeLists.txt. I have no explanation on why yet.

Baklap4:
> This seems to be linked to the removal of FetchContent_MakeAvailable(ced-sources) from top-level CMakeLists.txt. I have no explanation on why yet.

Kind of weird, but glad it's fixed :)

Within discord i asked if it's possible to squash these 29 commits into a couple neat commits as some are fixes upon fixes
As much as i'm concerned this can be merged after that 👍

yoda-jm:
> > This seems to be linked to the removal of FetchContent_MakeAvailable(ced-sources) from top-level CMakeLists.txt. I have no explanation on why yet.
> 
> Kind of weird, but glad it's fixed :)
> 
> Within discord i asked if it's possible to squash these 29 commits into a couple neat commits as some are fixes upon fixes As much as i'm concerned this can be merged after that +1

This should now be done (I have rebuild a complete branch based on the final diff)

Baklap4 (approved):
LGTM

For an upcoming pr please fix the `FindAubio.cmake` on windows vcpkg we don't have an `aubio.pc` (aubio pkg-config file) So it should be found manually with `libfind_pkg_detect`

PR #721, opened 2022-04-15
Some fixes and refactors for the cache

Lord-Kamina:
### More

### DIFF
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -66,16 +66,16 @@ namespace {
 	using Lock = std::lock_guard<std::mutex>;
 }
 
-	BinaryBuffer readFile(fs::path const& path) {
-		BinaryBuffer ret;
-		fs::ifstream f(path, std::ios::binary);
-		f.seekg(0, std::ios::end);
-		ret.resize(f.tellg());
-		f.seekg(0);
-		f.read(reinterpret_cast<char*>(ret.data()), ret.size());
-		if (!f) throw std::runtime_error("File cannot be read: " + path.string());
-		return ret;
-	}
+BinaryBuffer readFile(fs::path const& path) {
+	BinaryBuffer ret;
+	fs::ifstream f(path, std::ios::binary);
+	f.seekg(0, std::ios::end);
+	ret.resize(f.tellg());
+	f.seekg(0);
+	f.read(reinterpret_cast<char*>(ret.data()), ret.size());
+	if (!f) throw std::runtime_error("File cannot be read: " + path.string());
+	return ret;
+}
 
 void copyDirectoryRecursively(const fs::path& sourceDir, const fs::path& destinationDir)
 {
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -128,20 +128,20 @@ Paths getThemePaths() {
 	std::string theme = config["game/theme"].getEnumName();
 	Paths paths = getPaths();
 	Paths infixes = { 
-					  themes / theme,
-					  themes / theme / www,
-					  themes / theme / www / js,
-					  themes / theme / www / css,
-					  themes / theme / www / images,
-					  themes / theme / www / fonts,
+						themes / theme,
+						themes / theme / www,
+						themes / theme / www / js,
+						themes / theme / www / css,
+						themes / theme / www / images,
+						themes / theme / www / fonts,
 
-					  themes / def,
-					  themes / def / www,					  
-					  themes / def / www / js,
-					  themes / def / www / css,				  
-					  themes / def / www / images,			  
-					  themes / def / www / fonts,
-					  fs::path() };
+						themes / def,
+						themes / def / www,						
+						themes / def / www / js,
+						themes / def / www / css,					
+						themes / def / www / images,				
+						themes / def / www / fonts,
+						fs::path() };
 	if (!theme.empty() && theme != def) infixes.push_front(themes / theme);
 	// Build combinations of paths and infixes
 	Paths themePaths;
@@ -164,21 +164,21 @@ fs::path findFile(fs::path const& filename) {
 		if (fs::exists(p)) return p.string();
 	}
 	std::string logmsg = "fs/error: Unable to locate data file, tried:\n";
-	for (auto const& p: list) logmsg += "  " + p.string() + '\n';
+	for (auto const& p: list) logmsg += "	 " + p.string() + '\n';
 	std::clog << logmsg << std::flush;
 	throw std::runtime_error("Cannot find file \"" + filename.string() + "\" in Performous theme or data folders");
 }
 
 Paths listFiles(fs::path const& dir) {
 	if (dir.is_absolute()) throw std::logic_error("listFiles expects a folder name without path.");
-	std::set<fs::path> found;  // Filenames already found
-	Paths files;  // Full paths of files found
+	std::set<fs::path> found;	 // Filenames already found
+	Paths files;	// Full paths of files found
 	for (fs::path path: getThemePaths()) {
 		fs::path subdir = path / dir;
 		if (!fs::is_directory(subdir))
 			continue;
 		for (const auto &d : fs::recursive_directory_iterator(subdir)) {
-			fs::path name = d.path().filename();  // FIXME: Extract full path from current folder, not just the filename
+			fs::path name = d.path().filename();	// FIXME: Extract full path from current folder, not just the filename
 			// If successfully inserted to "found", it wasn't found before, so add to paths.
 			if (found.insert(name).second) files.push_back(d);
 		}
@@ -205,7 +205,7 @@ Paths getPathsConfig(std::string const& confOption) {
 	Lock l(mutex);
 	Paths ret;
 	for (auto const& str: config[confOption].sl()) {
-		ret.splice(ret.end(), cache.pathExpand(str));  // Add expanded paths to ret.
+		ret.splice(ret.end(), cache.pathExpand(str));	 // Add expanded paths to ret.
 	}
 	return ret;
 }
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -128,20 +128,20 @@ Paths getThemePaths() {
 	std::string theme = config["game/theme"].getEnumName();
 	Paths paths = getPaths();
 	Paths infixes = { 
-						themes / theme,
-						themes / theme / www,
-						themes / theme / www / js,
-						themes / theme / www / css,
-						themes / theme / www / images,
-						themes / theme / www / fonts,
+		themes / theme,
+		themes / theme / www,
+		themes / theme / www / js,
+		themes / theme / www / css,
+		themes / theme / www / images,
+		themes / theme / www / fonts,
 
-						themes / def,
-						themes / def / www,						
-						themes / def / www / js,
-						themes / def / www / css,					
-						themes / def / www / images,				
-						themes / def / www / fonts,
-						fs::path() };
+		themes / def,
+		themes / def / www,
+		themes / def / www / js,
+		themes / def / www / css,
+		themes / def / www / images,
+		themes / def / www / fonts,
+		fs::path() };
 	if (!theme.empty() && theme != def) infixes.push_front(themes / theme);
 	// Build combinations of paths and infixes
 	Paths themePaths;
@@ -164,21 +164,21 @@ fs::path findFile(fs::path const& filename) {
 		if (fs::exists(p)) return p.string();
 	}
 	std::string logmsg = "fs/error: Unable to locate data file, tried:\n";
-	for (auto const& p: list) logmsg += "	 " + p.string() + '\n';
+	for (auto const& p: list) logmsg += " " + p.string() + '\n';
 	std::clog << logmsg << std::flush;
 	throw std::runtime_error("Cannot find file \"" + filename.string() + "\" in Performous theme or data folders");
 }
 
 Paths listFiles(fs::path const& dir) {
 	if (dir.is_absolute()) throw std::logic_error("listFiles expects a folder name without path.");
-	std::set<fs::path> found;	 // Filenames already found
-	Paths files;	// Full paths of files found
+	std::set<fs::path> found; // Filenames already found
+	Paths files; // Full paths of files found
 	for (fs::path path: getThemePaths()) {
 		fs::path subdir = path / dir;
 		if (!fs::is_directory(subdir))
 			continue;
 		for (const auto &d : fs::recursive_directory_iterator(subdir)) {
-			fs::path name = d.path().filename();	// FIXME: Extract full path from current folder, not just the filename
+			fs::path name = d.path().filename(); // FIXME: Extract full path from current folder, not just the filename
 			// If successfully inserted to "found", it wasn't found before, so add to paths.
 			if (found.insert(name).second) files.push_back(d);
 		}
@@ -205,7 +205,7 @@ Paths getPathsConfig(std::string const& confOption) {
 	Lock l(mutex);
 	Paths ret;
 	for (auto const& str: config[confOption].sl()) {
-		ret.splice(ret.end(), cache.pathExpand(str));	 // Add expanded paths to ret.
+		ret.splice(ret.end(), cache.pathExpand(str)); // Add expanded paths to ret.
 	}
 	return ret;
 }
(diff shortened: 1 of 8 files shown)


game/fs.cc
```diff
@@ -77,6 +77,25 @@ namespace {
...
+			try {
+			file.open(filename.string());
+			if (!file.is_open()) throw std::runtime_error("Can't open file.");
+				json = nlohmann::json::parse(file);
+			} catch(nlohmann::json::exception const& e) {
+				std::clog << "json/error: " << e.what() << std::endl;
+			} catch(std::exception const& e) {
+				std::clog << "fs/error: " << e.what() << std::endl;
+			}
+			if (file.is_open()) {
```
Baklap4, line 93: No need to check if it's open. Just always close it :)

game/songs.cc
```diff
@@ -233,16 +218,18 @@ void Songs::CacheSonglist() {
 	}
 
 	fs::path cacheDir = getCacheDir() / SONGS_CACHE_JSON_FILE;
-
+	std::ofstream outFile;
```
Baklap4, line 221: Maybe also add a WriteJson() method within fs.cc

game/songs.cc
```diff
@@ -91,24 +89,11 @@ const std::string SONGS_CACHE_JSON_FILE = "songs.json";
...
-		} catch(std::exception const& e) {
-			std::clog << "songs/error: " << e.what() << std::endl;
-			file.close();
-			return;
-		}
-	} else {
-		std::clog << "songs/info: Could not open songs meta cache file " << songsMetaFile.string() << std::endl;
+	auto jsonRoot = readJSON(songsMetaFile);
+	if (jsonRoot.empty()) {
+		std::clog << "songs/info: Could not parse songs meta cache file " << songsMetaFile.string() << std::endl;
```
Baklap4, line 94: Duplicate message since readJson checks for parsing the cache.

game/fs.cc
```diff
@@ -77,6 +77,25 @@ namespace {
...
+			} catch(nlohmann::json::exception const& e) {
+				std::clog << "json/error: " << e.what() << std::endl;
+			} catch(std::exception const& e) {
+				std::clog << "fs/error: " << e.what() << std::endl;
+			}
+			if (file.is_open()) { 
+				file.close();
+			}
+		return json;
+	}
```
Baklap4, line 97: It might be the browser, but the indents are looking weird in this method.

game/fs.cc
```diff
@@ -77,6 +77,25 @@ namespace {
 		return ret;
 	}
 
+	nlohmann::json readJSON(fs::path const& filename) {
+		nlohmann::json json;
```
Baklap4, line 81: `json` isn't assigned/initialized a value. Thus a calling `.empty()` on the caller side (without checking for NULL) can result in a error right?

game/songs.cc
```diff
@@ -91,24 +89,11 @@ const std::string SONGS_CACHE_JSON_FILE = "songs.json";
...
-			jsonRoot = nlohmann::json::parse(buffer);
-		} catch(std::exception const& e) {
-			std::clog << "songs/error: " << e.what() << std::endl;
-			file.close();
-			return;
-		}
-	} else {
-		std::clog << "songs/info: Could not open songs meta cache file " << songsMetaFile.string() << std::endl;
+	auto jsonRoot = readJSON(songsMetaFile);
+	if (jsonRoot.empty()) {
```
Baklap4, line 93: Probs need a null-check? since `readJSON` can return an unitialized/unassigned json

Baklap4 (changes requested):

game/fs.cc
Lord-Kamina: I thought it would cause an error if it wasn't open but you're correct.

game/songs.cc
Lord-Kamina: Sound reasonable.

game/songs.cc
```diff
@@ -233,16 +213,18 @@ void Songs::CacheSonglist() {
...
 		const int spacesCount = 4;
 		outFile << jsonRoot.dump(spacesCount);
-		outFile.close();
+		std::clog << "songs/info: saved " + std::to_string(jsonRoot.size()) + " songs to the cache at " + cacheDir.string() << std::endl;
 	} catch (std::exception const& e) {
 		std::clog << "songs/error: Could not save " + cacheDir.string() + ": " + e.what() << std::endl;
-		return;
 	}
+	if (outFile.is_open()) outFile.close();
+	return;
```
OznOg, line 227: useless return

game/songs.cc
```diff
@@ -233,16 +213,18 @@ void Songs::CacheSonglist() {
...
+		if (!outFile.is_open()) throw std::runtime_error("Can't open file.");
 		const int spacesCount = 4;
 		outFile << jsonRoot.dump(spacesCount);
-		outFile.close();
+		std::clog << "songs/info: saved " + std::to_string(jsonRoot.size()) + " songs to the cache at " + cacheDir.string() << std::endl;
 	} catch (std::exception const& e) {
 		std::clog << "songs/error: Could not save " + cacheDir.string() + ": " + e.what() << std::endl;
-		return;
 	}
+	if (outFile.is_open()) outFile.close();
```
OznOg, line 226: fstream closes the file on destructor

game/songs.cc
```diff
@@ -233,16 +213,18 @@ void Songs::CacheSonglist() {
 	}
 
 	fs::path cacheDir = getCacheDir() / SONGS_CACHE_JSON_FILE;
-
+	std::ofstream outFile;
 	try {
-		std::ofstream outFile(cacheDir.string());
```
OznOg, line 238: why not creating the stream in the try catch block? you could use the constructor instead of open

game/fs.hh
```diff
@@ -53,6 +54,7 @@ static inline fs::path absolute(const fs::path& p, const fs::path& base) {
 
 typedef std::vector<std::uint8_t> BinaryBuffer;
 
+nlohmann::json readJSON(fs::path const& filename); ///< Reads a JSON file into memory.
```
OznOg, line 57: sorry, but I don't think putting this function here is a good idea. this file is meant as a wrapper for filesystem and should eventually be removed. BTW pulling a dep like json in a file that is for file manipulation is pulling a build dep in many places where json is not used. I think yo ushould put it in the place where json is used or create a special jsonhelper header/cpp is really needed.

cmake/Modules/FindJson.cmake
```diff
@@ -12,7 +12,7 @@ include(LibFindMacros)
 
 set(Json_FIND_REQUIRED FALSE)
 set(Json_FIND_QUIETLY TRUE)
-libfind_pkg_detect(Json nlohmann_json FIND_PATH nlohmann/json.hpp)
+libfind_pkg_detect(Json nlohmann_json VERSION >= 3.6.0 FIND_PATH nlohmann/json.hpp)
```
Baklap4, line 15: Won't work for ubuntu1804.
Either add the build steps in this module or add nlohman_json 3.6.0 as backport to your PPA

game/songs.cc
Lord-Kamina: Because then I can't close it outside of the try block.

game/songs.cc
Lord-Kamina: Oversight

game/fs.hh
Lord-Kamina: I think you're wrong because fs.cc/hh existed before implementing filesystem and they do already have other functions in them. Whether we might eventually transition away from it and move all those functions elsewhere is a valid point but IMO we'll cross that bridge when we get to it.

game/songs.cc
OznOg: as the destructor closes the file, you don't actually need to close it ouside the try block

yoda-jm:
I don't see why all of these is required:
- Json should be a header only library, why do we need to add the include of the libraries in CMake
- There is a lot of indent change that are not related to this PR
- I am not a bug fan of pulluting the fs library with json related code (I was already reluctant to have it in a core structure such as Songs) (it also complexify the CMake file),
- please don't use contains instead of count since it was carefully chosen in order to lower the dependency version,
- also this patch since to activate the cache reading in all Performous versions whereas it was only activated when compiling using webserver feature (this was carefully chosen because it seemed that the cache loading still had problems).

Lord-Kamina:
> I don't see why all of these is required:
> 
> - Json should be a header only library, why do we need to add the include of the libraries in CMake
> 

That's a fair point, and it might be unnecessary.

> - There is a lot of indent change that are not related to this PR
> 

Yes. It just seemed awkward to make a PR to correct the indent of code that hasn't been touched in ages, when I'm changing something literally next to it.


> - I am not a bug fan of pulluting the fs library with json related code (I was already reluctant to have it in a core structure such as Songs) (it also complexify the CMake file),
> 

The reason I'm doing this is because my reworked webserver uses json elsewhere and thus we'd have to begin duplicating code.

I put it in fs because it seemed the most fitting.

Where else would you put it? Its own file?

> - please don't use contains instead of count since it was carefully chosen in order to lower the dependency version,
> 

And I tried at first, but truly what's the point? What system/configuration are you trying to support out of the box here? Ubuntu 18.04? That has nlohmann 2, the last version of which was released **five** years ago. Nlohmann 3 is faster, more robust and supports exceptions properly. If you can show me a single relevant case where using 3.6.0 instead of an earlier 3.x version will make it easier for anyone, I'll change it back to count.

> - also this patch since to activate the cache reading in all Performous versions whereas it was only activated when compiling using webserver feature (this was carefully chosen because it seemed that the cache loading still had problems).

Does it? @Baklap4 correct me if I'm wrong but I think the only actual reason the cache depended on webserver is because they were both supplied by CppRest (which came with several other dependencies as well)

game/song.cc
```diff
@@ -15,7 +15,7 @@ extern "C" {
 }
 
 template <typename T> std::optional<T> getJsonEntry(nlohmann::json const& json, const char *name) {
```
Baklap4, line 17: Template should live in json.hh/cc

yoda-jm: Agreed, and please also keep the count() > 0 (it enable you to remove the check in FindJson.cmake

game/songs.cc
```diff
@@ -41,25 +40,23 @@ void Songs::reload() {
 	}
 	// Run loading thread
 	m_loading = true;
-	if (m_thread) m_thread->join();
```
yoda-jm, line 44: I would prefer to have all this threading logic done in a separate PR (changed from line 40 to 63).
Replacing loading/reading with json wrapper is ok.

cmake/Modules/FindJson.cmake
```diff
@@ -15,7 +15,7 @@ set(Json_FIND_QUIETLY TRUE)
 libfind_pkg_detect(Json nlohmann_json FIND_PATH nlohmann/json.hpp)
 libfind_process(Json)
 
-if (Json_FOUND)
+if (Json_FOUND AND Json_PKGCONF_VERSION VERSION_GREATER_EQUAL 3.6.0)
```
yoda-jm, line 18: can probably be removed if using count() instead of contains()

game/songs.cc
yoda-jm: You can also maybe rebase in order to have 3 separate commits (in this PR):
- 1 for include order + spaces/indent
- 1 for json wrapper
- 1 for the threading logic (I would still prefer a separate PR for this, as it is more touchy subject)

cmake/Modules/FindJson.cmake
Lord-Kamina: I can remove it but this was intentional. We still need at least 3.x because of exceptions among other things. 

That's why I thought to just go with contains, because I doubt any disto that has 3.x will have a version older than 3.6.0.

If you're still not convinced, I'll lower the check to whatever version it is that exceptions were actually introduced in, but removing it will lead to a failed build in 18.04 because it will find an incompatible version in system headers.

game/songs.cc
Lord-Kamina: That sounds doable.

cmake/Modules/FindJson.cmake
Baklap4: 1804 support will drop when they anounce EOL which is in a couple of days. Ubuntu 2004 and 2204 do support 3.6

game/song.cc
Baklap4: The template still lives in game/song.cc please move it over to game/json.hh

Baklap4:
> Does it? @Baklap4 correct me if I'm wrong but I think the only actual reason the cache depended on webserver is because they were both supplied by CppRest (which came with several other dependencies as well)

@Lord-Kamina They were both supplied by CppRest so that's why webserver support 'depended' on json. Now with nlohmann-json it can be added as an option or become part of the core system. If it becomes part of the core system the lib should be required, else can be optional.

Lord-Kamina:
I removed the change to `reload_internal` entirely.
I still left three commits, with the removal of the `#IFDEF WEBSERVER` bit on its own commit.

ooshlablu:
Played a few songs with vocals + guitar using the AppImage and packages on Ubuntu 18.04 & 20.04. Didn't notice any effects on gameplay, which is to be expected I believe. If anything, the cached songs are loaded faster, and load in a few seconds rather than the 30 seconds + that it used to take.

ooshlablu (approved):
Cmake and Actions changes LGTM

ooshlablu (dismissed):
LGTM

Baklap4 (changes requested):
Looks alright some minor issues
And for inline comments in this pr please use just 1 space in between :)

game/fs.cc
```diff
@@ -164,21 +164,21 @@ fs::path findFile(fs::path const& filename) {
 		if (fs::exists(p)) return p.string();
 	}
 	std::string logmsg = "fs/error: Unable to locate data file, tried:\n";
-	for (auto const& p: list) logmsg += "  " + p.string() + '\n';
+	for (auto const& p: list) logmsg += "	 " + p.string() + '\n';
```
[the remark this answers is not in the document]
Lord-Kamina: Sure, but I think it was because it's a list with multiple lines maybe?

game/json.cc
```diff
@@ -0,0 +1,36 @@
+#include "json.hh"
+
```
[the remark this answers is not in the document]
Lord-Kamina: Oh. I  did it on purpose. Maybe we should discuss this properly for the style guide but I like separating headers in different categories as follows:

* Own header.
* Other performous headers.
* 3rd party libraries
* STL headers.

And within each category, sorted alphabetically. Do you have an opinion on the general idea, and/or whether you like that ordering or would propose a different one?

game/songs.cc
```diff
@@ -1,28 +1,29 @@
...
 #include <cstdlib>
 #include <iostream>
+#include <fstream>
 #include <regex>
 #include <stdexcept>
 
-#include "fs.hh"
 #include <fmt/format.h>
+#include <nlohmann/json.hpp>
 #include <unicode/stsearch.h>
```
Baklap4, line 25: Change this order to be:
1. local headers
2. 3rd party headers
3. std headers

Baklap4 (changes requested):
As per dicussion earlier i think only the reorder of headers is needed, then this gets approval :)

game/songs.cc
```diff
@@ -95,24 +94,8 @@ const std::string SONGS_CACHE_JSON_FILE = "songs.json";
...
-			file.close();
-			return;
-		}
-	} else {
-		std::clog << "songs/info: Could not open songs meta cache file " << songsMetaFile.string() << std::endl;
-		return;
-	}
-
+	auto jsonRoot = readJSON(songsMetaFile);
+	if (jsonRoot.empty()) return;
```
Baklap4, line 98: Seems the `nlohmann/json.hpp` is still needed as `.empty()` isn't abstracted away

Lord-Kamina: Actually I think you're right; it was an oversight. It compiles fine on mac without the include.

Baklap4: It should since this file includes `json.hh` which in turn includes this `nlohmann/json.hpp`, but then again having a file call something which is not directly listed as an include is nasty imho

Lord-Kamina: Yes, but IMO half the point of json.hh existing is not having to include nlohmann everywhere.

Baklap4 (approved):
LGTM

PR #727, opened 2022-04-19
Fix build with GCC 12

sergiomb2:
### Motivation

allow to build with gcc 12 and in Fedora 36

https://gcc.gnu.org/gcc-12/porting_to.html
 The following headers are used less widely in libstdc++ and may need to be included explicitly when compiled with GCC 12: 

please test and confirm this fix is good .

Thank you

### DIFF
--- a/game/libda/sample.hpp
+++ b/game/libda/sample.hpp
@@ -62,10 +62,12 @@ namespace da {
 	static inline int conv_to_s24_fast(sample_t s) { return static_cast<int>(s * max_s24); }
 	static inline int conv_to_s32_fast(sample_t s) { return static_cast<int>(s * max_s32); }
 
-	template <typename ValueType> class step_iterator: public std::iterator<std::random_access_iterator_tag, ValueType> {
+	template <typename ValueType> class step_iterator {
 		ValueType* m_pos;
 		std::ptrdiff_t m_step;
 	  public:
+        using iterator_category = std::random_access_iterator_tag;
+        using value_type = ValueType;
 		step_iterator(ValueType* pos, std::ptrdiff_t step): m_pos(pos), m_step(step) {}
 		ValueType& operator*() { return *m_pos; }
 		step_iterator operator+(std::ptrdiff_t rhs) { return step_iterator(m_pos + m_step * rhs, m_step); }


Baklap4:
Build seems to work, yet we're not targetting gcc 12 anywhere within the pipelines. @OznOg / @ooshlablu Do any of you have gcc 12 installed?

@ooshlablu do one of the ubuntu images use gcc 12?

ooshlablu:
> @ooshlablu do one of the ubuntu images use gcc 12?

Not currently. Ubuntu 22.04 and Fedora 35 (both latest for their distros) use gcc 11 which is what's currently in their repos.

I'm not sure what the future plans are for either distros tho, so it may come up sooner rather than later.

Baklap4:
Hmmm might be worth looking into https://github.com/performous/performous-docker/issues/11 and upgrade the way we handle those dockerfiles. Instead of directly installing gcc-8 for example in ubuntu1804 we could also say `apt install gcc` which pulls the latest version for the specific distro. This in turn combined with the other things specified there makes sure we're always using the latest compiler on the specific distro supplied by its packagemanager.

sergiomb2:
Fedora 36 with gcc12 should be available in a week or two .... 
https://fedorapeople.org/groups/schedule/f-36/f-36-key-tasks.html

ooshlablu:
Tested on Ubuntu 18.04 (gcc 8) and Ubuntu 20.04 (gcc 11), and everything seems to work as expected. Tested with vocals and guitar/bass. :+1:

Baklap4:
Blocked by https://github.com/performous/performous-docker/pull/16

Baklap4:
@OznOg i might be confused with the ffmpeg 5.0.0 branch but did you want any more change here?
It does build now with gcc 12 on fedora 36 since https://github.com/performous/performous-docker/pull/16 got merged

OznOg:
Could you give the original build error, I cannot figure out what is the link between the patch and the g++12 differences

ooshlablu:
Also, just as a measurement: We merged in Fedora 36 support recently, so all PRs are currently building with it, and we don't see any errors related to gcc 12. 
We did see compilation errors on Fedora 36 with ffmpeg 5, which were resolved here: https://github.com/performous/performous/pull/752

Baklap4:
> Also, just as a measurement: We merged in Fedora 36 support recently, so all PRs are currently building with it, and we don't see any errors related to gcc 12.
>We did see compilation errors on Fedora 36 with ffmpeg 5, which were resolved here: https://github.com/performous/performous/pull/752

So this one can get closed then i guess?

Baklap4:
Closing this since the aren't any errors related to gcc-12 as we can see within fedora-36 (ci build)

Please re-open if it's still an issue

PR #729, opened 2022-04-22
Use upstream ced?

OznOg:
use upstream ced

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,13 +83,14 @@ endif()
 
 include(FetchContent)
 FetchContent_Declare(ced-sources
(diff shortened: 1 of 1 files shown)


OznOg:
vanilla ced is not usable... let's forget it

Baklap4:
Wouldn't toss it in the ring that easily. I'll try have a look at it tomorrow :)

OznOg:
Actually, the upstream ced project is almost dead and would need more patches to be handy. In the end, I really doubt it is useful to try to upstream things, we'd better assume that our fork is now the one up to date and don't try to backport things.

PR #732, opened 2022-04-23
Adds debian to the build pipeline

Baklap4:
### What does this PR do?

Adds debian to the build pipelines to check if we can also build on there.

### Motivation

It's a good thing to have support for more platforms and be sure that we can build on them.

### More


ooshlablu (dismissed):

.github/workflows/build_and_release.yml
```diff
@@ -137,7 +139,7 @@ jobs:
           case ${ID} in
```
ooshlablu, line 139: The `if` statement above needs to be set like it is here: https://github.com/performous/performous-docker/blob/master/build_performous.sh#L67-L71

(I tried to comment on the actual line, but you can't comment on lines that didn't change?)

ooshlablu (changes requested):
Don't forget the README :-)

.github/workflows/build_and_release.yml
```diff
@@ -137,7 +139,7 @@ jobs:
...
             'fedora')
           PACKAGE_TYPE='RPM';;
-            'ubuntu')
+            'ubuntu|debian')
```
ooshlablu, line 142: I think this needs to be specified like `'ubuntu'|'debian')`. That works in the build script in the containers at least.

ooshlablu (changes requested):
Still need updates to README also

.github/workflows/build_and_release.yml
ooshlablu: Hmmm, based upon the build failure, it looks like cpack doesn't come with TAR out of the box anymore: `CPack Error: Could not create CPack generator: TAR`. It might be worth changing the `PACKAGE_TYPE='TAR';;` statement below to be `PACKAGE_TYPE='TGZ;;`

Baklap4: > Hmmm, based upon the build failure, it looks like cpack doesn't come with TAR out of the box anymore: CPack Error: Could not create CPack generator: TAR. It might be worth changing the PACKAGE_TYPE='TAR';; statement below to be PACKAGE_TYPE='TGZ';;

lol not sure how i didnt spot this before but cpack never came with TAR package type:
```
7Z - 7zip - (.7z)
TBZ2 (.tar.bz2)
TGZ (.tar.gz)
TXZ (.tar.xz)
TZ (.tar.Z)
TZST (.tar.zst)
ZIP (.zip)
```

Baklap4: TAR issue is resolved on master. Within this PR i've changed the case statement and updated the readme ;)

ooshlablu (approved):

PR #736, opened 2022-04-25
Build AppImage

ooshlablu:
### What does this PR do?

Enables AppImage build in the GH-Actions workflow.

### Closes Issue(s)

Supercedes https://github.com/performous/performous/pull/642


Baklap4:
Not sure but does an AppImage compiled on 20.04 work on 22.04, fedora and other systems? Or do we need to have an output for each system like with the deb/rpm packages?

.github/workflows/build_and_release.yml
```diff
@@ -200,6 +200,85 @@ jobs:
...
 
+  # Create the AppImage
+  AppImage:
+    name: Create the AppImage
+    needs:
+      - determine_version
+      - create_release
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install the AppImage bundler and Performous deps
```
Baklap4, line 211: This step can be made into a dockerfile. I think we'll be alright by just including the AppImageCrafter builder thingy and supplying a build-arg with it.

.github/workflows/build_and_release.yml
```diff
@@ -200,6 +200,85 @@ jobs:
...
+        id: upload_assets
+        if: ${{ github.event_name != 'pull_request' && github.ref_type == 'tag' }}
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ needs.create_release.outputs.upload_url }}
+          asset_path: ${{ env.ARTIFACT_PATH }}
+          asset_name: ${{ env.ARTIFACT_NAME }}
+          asset_content_type: application/octet-stream
```
Baklap4, line 279: This is all the same compared to a normal linux deploy any chance we can not duplicate this?

.github/workflows/build_and_release.yml
```diff
@@ -200,6 +200,85 @@ jobs:
...
+        id: build_appimage
+        run: |
+          cd performous
+          mkdir build
+          cd build
+          PACKAGE_VERSION=${{ needs.determine_version.outputs.complete_version }}
+          cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DENABLE_WEBSERVER=ON -DENABLE_WEBCAM=ON -DPERFORMOUS_VERSION=$PACKAGE_VERSION ..
+          make -j$(nproc) install DESTDIR=../AppDir
+          cd ..
+          appimage-builder --recipe AppImageBuilder.yml --skip-test
```
Baklap4, line 235: Within `AppImageBuilder.yml` we're listing a version `1.x` this should become `$PACKAGE_VERSION` any chance we can override that value from the cli or by tokenizing it like on osx?

AppImageBuilder.yml
```diff
@@ -0,0 +1,82 @@
...
+      - libpng16-16
+      - libportaudio2
+      - libportmidi0
+      - librsvg2-2
+      - libsdl2-2.0-0
+      - libssl1.1
+      - libstdc++6
+      - libswresample3
+      - libswscale5
+      - libxml++2.6-2v5
```
Baklap4, line 51: This breaks portability with other systems than ubuntu 20.04 i think

Baklap4 (changes requested):
I have some concerns regarding the portability of this

AppImageBuilder.yml
ooshlablu: It shouldn't. Those dependencies are being installed inside of the AppImage

.github/workflows/build_and_release.yml
```diff
@@ -200,6 +200,85 @@ jobs:
...
+
+          WORK_DIR=$(pwd)
+          PACKAGE_PATH=$(ls ${WORK_DIR}/*.AppImage)
+          PACKAGE_NAME=$(echo ${PACKAGE_PATH} | sed 's/\(performous\).*$/\1/')
+          PACKAGE_SUFFIX=".AppImage"
+          NEW_PACKAGE_NAME="${PACKAGE_NAME}-${PACKAGE_VERSION}-${ID}_${VERSION_ID}${PACKAGE_SUFFIX}"
+          MASTER_NEW_PACKAGE_NAME="${PACKAGE_NAME}-latest-${ID}_${VERSION_ID}${PACKAGE_SUFFIX}"
```
Baklap4, line 242: The endresult artifact has naming problems: `Artifact performous-1.2.0-736-85dd735-alpha-_.AppImage has been successfully uploaded!`

.github/workflows/build_and_release.yml
ooshlablu: I think we are hoping that YAML templates will be available at some point to take care of the duplication.

.github/workflows/build_and_release.yml
ooshlablu: It can't be because even building the AppImage has a dependency on `fuse` which is not straightforward to set up.

ooshlablu:
> Not sure but does an AppImage compiled on 20.04 work on 22.04, fedora and other systems? Or do we need to have an output for each system like with the deb/rpm packages?

Yes, the AppImage produced should work on any version of Linux that supports running them even though it was built on Ubuntu 20.04. For instance I just tested the image produced during this run on an Ubuntu 18.04 machine and it works.

It would be great is someone with Fedora can verify.

.github/workflows/build_and_release.yml
ooshlablu: Yeah, it looks like I'll have to do the tokenizing to fix it.

Baklap4:
@OznOg can you try to run the produced AppImage within Fedora?

AppImageBuilder.yml
Baklap4: It's yaml, and each operation is executed in the bash shell. So referencing the version should work :)

.github/workflows/build_and_release.yml
Baklap4: It's yaml, and each operation is executed in the bash shell. So referencing the version should work :)

.github/workflows/build_and_release.yml
ooshlablu: Yep, fixed.

.github/workflows/build_and_release.yml
```diff
@@ -200,6 +200,86 @@ jobs:
...
+      - create_release
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Install the AppImage bundler and Performous deps
+        id: fetch_deps
+        run: |
+          wget -O appimage-builder-x86_64.AppImage https://github.com/AppImageCrafters/appimage-builder/releases/download/v1.0.0-beta.1/appimage-builder-1.0.0-677acbd-x86_64.AppImage
+          chmod +x appimage-builder-x86_64.AppImage
+          sudo mv appimage-builder-x86_64.AppImage /usr/local/bin/appimage-builder
+          sudo apt-get install -y --no-install-recommends git cmake build-essential gettext help2man libopenblas-dev libfftw3-dev libicu-dev libepoxy-dev libsdl2-dev libfreetype6-dev libpango1.0-dev librsvg2-dev libxml++2.6-dev libavcodec-dev libavformat-dev libswscale-dev libjpeg-dev portaudio19-dev libglm-dev libboost-filesystem-dev libboost-iostreams-dev libboost-locale-dev libboost-system-dev libboost-program-options-dev libssl-dev libcpprest-dev libportmidi-dev libopencv-dev libaubio-dev
```
Baklap4, line 217: missing `nlohmann-json3-dev`

AppImageBuilder.yml
```diff
@@ -0,0 +1,82 @@
...
+      - libavutil56
+      - libboost-iostreams1.71.0
+      - libboost-locale1.71.0
+      - libboost-program-options1.71.0
+      - libc6
+      - libcairo2
+      - libcpprest2.10
+      - libepoxy0
+      - libfontconfig1
+      - libgcc-s1
```
Baklap4, line 33: missing `nlohmann-json3-dev` (it's currently building do to self-build)

Baklap4 (commented):
Still missing 2 dependencies heheh

Baklap4 (approved):
LGTM

PR #738, opened 2022-04-26
Build appimage

ooshlablu:
### What does this PR do?
Fix file excludes, copy libnsl.so.1 to the root so it stops complaining


yoda-jm (approved):
perfect if it works

PR #741, opened 2022-04-29
always use self_built aubio on mac

Lord-Kamina:
I made the cmake build so I wouldn't have to install this particular lib when building performous.

### More


yoda-jm:
I don't understand why we don't want the system lib. If the packaging has a problem then we should fix the packaging.

Baklap4:
We really should settle this debate once and for all 😂 
We now have a couple of options around aubio and ced. (Auto, never, always). This fixes the issues some were having with submodules and downloading while building.

However the actual problem: Order of how we want to look for our dependencies hasn't been solved.
So what will we do with that?

My stance on this would be (on Auto mode):
1. Search system wide installed packages
    - apt for debian based
    - dnf for fedora based
    - pacman for arch based
    - homebrew/macports for macos
    - vcpkg for windows 
2. If not found _**or required version is not found**_ download and build from git (preferably from the main trunk instead of a forked version we maintain) with the newly added function: `libfetch_git_pkg`.

This would mean that every module we currently have should/need to have the options added so we can specify it for every dependency we have. And this would also mean that we need to specify a minimum version we want to support of a library. By specifying the minimum version of a library the argument "the version we use is waaay too old" gets neglected.

Within the Github workflow and our local builds we can then use 'Auto' for every dependency since we've stated a minimum version. This will either then use the lib provided from the system or built it from scratch using `libfetch_git_pkg`

If for some reason stuff doesn't work and we'll have to override the behavior we can use "Never" to only use system libraries or "Always" to only use git based self-built libraries. This is currently the case for Aubio on windows for instance. Since the vcpkg based install doesn't get found with pkg-config (since it doesn't come with a `.pc` file on vcpkg) therefore we have to build it ourselves.

Lord-Kamina:
> I don't understand why we don't want the system lib. If the packaging has a problem then we should fix the packaging.

Because there is no "system" lib, only installing from one of two (actually three) different package managers that might have different versions and build settings.

I originally modified the aubio build script so it would be configured with the least amount of dependencies while still being suitable for what we want it.

I agree it's a packager's responsibility, but I'm the macOS packager. 😂

Lord-Kamina:
The macports version does use fftw3 but also depends on libsamplerate and libsndfile

Baklap4:
Ok, so if i get this correct:
Homebrew's version of Aubio doesn't use FFTW3
MacPorts version of Aubio does use FFTW3
Selfbuilt version of Aubio does use FFTW3

So clearly we want to use an Aubio with FFTW3 (according to this PR).
I'm not entirely sure why that is as both builds seem to work? So that's where the confusion comes into play i guess

The windows build uses FFTW3 (since we force to build it there)
The linux build doesn't use FFTW3 (as far as i can tell) but yet works fine?

My question probably is what is FFTW3 and why do we need to use it as opposed to other methods?

Lord-Kamina:
> Ok, so if i get this correct:
> 
> Homebrew's version of Aubio doesn't use FFTW3
> 
> MacPorts version of Aubio does use FFTW3
> 
> Selfbuilt version of Aubio does use FFTW3
> 
> 
> 
> So clearly we want to use an Aubio with FFTW3 (according to this PR).
> 
> I'm not entirely sure why that is as both builds seem to work? So that's where the confusion comes into play i guess
> 
> 
> 
> The windows build uses FFTW3 (since we force to build it there as the system supplied one cannot be found properly. However the version provided by vcpkg doesn't use FFTW3 as far as i can tell: https://github.com/microsoft/vcpkg/blob/master/ports/aubio)
> 
> The linux build doesn't use FFTW3 (as far as i can tell) but yet works fine?
> 
> 
> 
> My question probably is what is FFTW3 and why do we need to use it as opposed to other methods?

Building without fftw3 will cause runtine crashes.

Baklap4:
> Building without fftw3 will cause runtime crashes.

So current master gives runtime crashes right now?
Are these crashes within Performous or are these within Aubio? If the latter, it might already have been fixed upstream (we're 250 commits behind or so)

Lord-Kamina:
> > Building without fftw3 will cause runtime crashes.
> 
> 
> 
> So current master gives runtime crashes right now?
> 
> Are these crashes within Performous or are these within Aubio? If the latter, it might already have been fixed upstream (we're 250 commits behind or so)

Just replying for the sake of posterity. The reason is we're using NPOT window/sample sizes (testing, it's what gave me best results) and those require fftw3.

Baklap4 (approved):
still looks good to me ;)

Baklap4 (approved):

PR #742, opened 2022-04-29
Refactor Menus

Lord-Kamina:
Our current code is essentially creating new Menu and MenuOption objects permanently, when there's no real good reason to do that.

This was originally part of a humoungous PR involving overhauling the text rendering, but I am in the process of splitting it in usable PRs.

### More

### DIFF
--- a/game/dancegraph.cc
+++ b/game/dancegraph.cc
@@ -103,7 +103,8 @@ void DanceGraph::setupJoinMenu() {
 	m_menu.clear();
 	updateJoinMenu();
 	// Populate root menu
-	m_menu.add(MenuOption(_("Ready!"), _("Start performing!")));
+	auto ready = std::make_unique<MenuOption>(_("Ready!"), _("Start performing!"));
+	m_menu.add(std::move(ready));
 	{ // Create track selector
 		ConfigItem::OptionList ol;
 		int i = 0, cur = 0;
@@ -114,7 +115,9 @@ void DanceGraph::setupJoinMenu() {
 		}
 		m_selectedTrack = ConfigItem(ol); // Create a ConfigItem from the option list
 		m_selectedTrack.select(cur); // Set the selection to current level
-		m_menu.add(MenuOption("", _("Select track")).changer(m_selectedTrack)); // MenuOption that cycles the options
+		auto selectTrack = std::make_unique<MenuOption>("", _("Select track"));
+		selectTrack->changer(m_selectedTrack);
+		m_menu.add(std::move(selectTrack)); // MenuOption that cycles the options
 		m_menu.back().setDynamicName(m_trackOpt); // Set the title to be dynamic
 	}
 	{ // Create difficulty opt
@@ -130,10 +133,14 @@ void DanceGraph::setupJoinMenu() {
 		}
 		m_selectedDifficulty = ConfigItem(ol); // Create a ConfigItem from the option list
 		m_selectedDifficulty.select(cur); // Set the selection to current level
-		m_menu.add(MenuOption("", _("Select difficulty")).changer(m_selectedDifficulty)); // MenuOption that cycles the options
+		auto selectDifficulty = std::make_unique<MenuOption>("", _("Select difficulty"));
+		selectDifficulty->changer(m_selectedDifficulty);
+		m_menu.add(std::move(selectDifficulty)); // MenuOption that cycles the options
 		m_menu.back().setDynamicName(m_difficultyOpt); // Set the title to be dynamic
 	}
-	m_menu.add(MenuOption(_("Quit"), _("Exit to song browser")).screen("Songs"));
+	auto quit = std::make_unique<MenuOption>(_("Quit"), _("Exit to song browser"));
+	quit->screen("Songs");
+	m_menu.add(std::move(quit));
 }
 
 void DanceGraph::updateJoinMenu() {
--- a/game/guitargraph.cc
+++ b/game/guitargraph.cc
@@ -160,13 +160,17 @@ void GuitarGraph::setupJoinMenuDifficulty() {
 	}
 	m_selectedDifficulty = ConfigItem(ol); // Create a ConfigItem from the option list
 	m_selectedDifficulty.select(cur); // Set the selection to current level
-	m_menu.add(MenuOption("", _("Select difficulty")).changer(m_selectedDifficulty)); // MenuOption that cycles the options
+	auto selectDifficulty = std::make_unique<MenuOption>("", _("Select difficulty"));
+	selectDifficulty->changer(m_selectedDifficulty);
+	m_menu.add(std::move(selectDifficulty)); // MenuOption that cycles the options
 	m_menu.back().setDynamicName(m_difficultyOpt); // Set the title to be dynamic
 }
 
 void GuitarGraph::setupJoinMenuDrums() {
 	setupJoinMenuDifficulty();
-	m_menu.add(MenuOption(_("Lefty-mode"), "").changer(m_leftymode));
+	auto leftyMode = std::make_unique<MenuOption>(_("Lefty-mode"), "");
+	leftyMode->changer(m_leftymode);
+	m_menu.add(std::move(leftyMode));
 	m_menu.back().setDynamicComment(m_leftyOpt);
 }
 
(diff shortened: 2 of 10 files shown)


ooshlablu:
Tested with singing and guitar/bass on a few songs. I didn't notice anything different game play or experience wise. :+1:

game/dancegraph.cc
```diff
@@ -103,7 +103,8 @@ void DanceGraph::setupJoinMenu() {
 	m_menu.clear();
 	updateJoinMenu();
 	// Populate root menu
-	m_menu.add(MenuOption(_("Ready!"), _("Start performing!")));
+	auto _ready = std::make_unique<MenuOption>(_("Ready!"), _("Start performing!"));
```
Baklap4, line 106: please get rid of the `_` prefix for local variables.

game/screen_songs.cc
```diff
@@ -529,17 +529,16 @@ void ScreenSongs::drawMenu() {
 	for (MenuOptions::const_iterator it = m_menu.begin(); it != m_menu.end(); ++it) {
```
OznOg, line 529: use range for loop to get rid of (*it)

game/menu.hh
```diff
@@ -95,18 +95,17 @@ public:
...
 	const MenuOptions::const_iterator begin() const { return menu_stack.back()->begin(); }
 	const MenuOptions::const_iterator end() const { return menu_stack.back()->end(); }
-	const MenuOptions getOptions() const { return *menu_stack.back(); }
+	const MenuOptions& getOptions() const { return *menu_stack.back(); }
 
 	Dimensions dimensions;
 
 private:
-	MenuOptions root_options;
+	bool m_open;
```
OznOg, line 107: why did  m_open moved here?

game/menu.cc
```diff
@@ -4,9 +4,7 @@
 #include "fs.hh"
 #include "game.hh"
 
-MenuOption::MenuOption(std::string const& nm, std::string const& comm, MenuImage img):
-  type(), value(), newValue(), callback(), image(img), name(nm), comment(comm), namePtr(), commentPtr()
-{}
+MenuOption::MenuOption(std::string const& nm, std::string const& comm, MenuImage img): type(), value(), newValue(), options(std::make_shared<MenuOptions>()), callback(), image(img), name(nm), comment(comm), namePtr(), commentPtr() {}
```
OznOg, line 7: no need to initialize a shared_ptr to null, it is already done in the defautl constructor

game/menu.cc
```diff
@@ -31,10 +29,10 @@ bool MenuOption::isActive() const {
 }
 
 
-Menu::Menu(): dimensions(), m_open(true) { clear(); }
+Menu::Menu(): dimensions(), m_open(true), root_options(std::make_shared<MenuOptions>()) { clear(); }
```
OznOg, line 32: same, If yo ureally want to stress this is set to nullptr, set it at nullptr at its declaration.

game/menu.cc
```diff
@@ -92,10 +90,10 @@ void Menu::action(int dir) {
 }
 
 void Menu::clear(bool save_root) {
-	if (!save_root) root_options.clear();
+	if (!save_root) root_options->clear();
 	menu_stack.clear();
 	selection_stack.clear();
-	menu_stack.push_back(&root_options);
```
OznOg, line 98: not sure this is realy what we want to have: should menu_stack **own** the root_options? It seems to me the sub menu is just using it, so using a raw pointer can be ok (or reference wrapper if you really don't like pointers), or if you really fear dangling pointers, you may use weak_ptr here.

game/menu.hh
```diff
@@ -37,7 +37,7 @@ public:
 	/// Make the option set a given value for ConfigItem and close the menu.
 	MenuOption& setter(ConfigItem& val, ConfigItem newval) { type = Type::SET_AND_CLOSE; value = &val; newValue = newval; return *this; }
 	/// Make the option open a submenu
-	MenuOption& submenu(MenuOptions opts) { type = Type::OPEN_SUBMENU; options = opts; return *this; }
+	MenuOption& submenu(MenuOptions opts) { type = Type::OPEN_SUBMENU; options = std::make_shared<MenuOptions>(std::move(opts)); return *this; }
```
OznOg, line 40: option =  std::make_shared<MenuOptions>(std::move(opts)) => option = std::move(opts);

game/menu.hh
```diff
@@ -95,18 +95,17 @@ public:
...
 	const MenuOptions::const_iterator end() const { return menu_stack.back()->end(); }
-	const MenuOptions getOptions() const { return *menu_stack.back(); }
+	const MenuOptions& getOptions() const { return *menu_stack.back(); }
 
 	Dimensions dimensions;
 
 private:
-	MenuOptions root_options;
+	bool m_open;
+	std::shared_ptr<MenuOptions> root_options;
```
OznOg, line 108: std::shared_ptr<MenuOptions> root_options = nullptr; If you really want to explicitely show it was initialized

game/screen_songs.cc
```diff
@@ -514,7 +514,7 @@ void ScreenSongs::drawMenu() {
 	if (m_menu.empty()) return;
 	// Some helper vars
 	ThemeInstrumentMenu& th = *m_menuTheme;
-	const auto cur = &m_menu.current();
+	MenuOption& cur = m_menu.current();
```
OznOg, line 517: why removing the const and the auto?

game/screen_songs.cc
```diff
@@ -529,17 +529,16 @@ void ScreenSongs::drawMenu() {
 	for (MenuOptions::const_iterator it = m_menu.begin(); it != m_menu.end(); ++it) {
 		// Pick the font object
 		SvgTxtTheme* txt = &th.option_selected;
-		if (cur != &*it)
-			txt = &(th.getCachedOption(it->getName()));
+		if (&cur != &*it->get()) txt = &(th.getCachedOption((*it)->getName()));
```
OznOg, line 532: if (&cur != it->get())

game/instrumentgraph.cc
```diff
@@ -53,10 +53,17 @@ bool InstrumentGraph::dead() const { return m_jointime != m_jointime || m_dead >
 
 void InstrumentGraph::setupPauseMenu() {
 	m_menu.clear();
-	m_menu.add(MenuOption(_("Resume"), _("Back to performing!")));
-	m_menu.add(MenuOption(_("Rejoin"), _("Change selections")).changer(m_rejoin));
-	m_menu.add(MenuOption(_("Restart"), _("Start the song\nfrom the beginning")).screen("Sing"));
-	m_menu.add(MenuOption(_("Quit"), _("Exit to song browser")).screen("Songs"));
+	auto _resume = std::make_unique<MenuOption>(_("Resume"), _("Back to performing!"));
```
OznOg: if you don't use _resume, add call directly make_unique in the m_menu.add

game/guitargraph.cc
```diff
@@ -160,13 +160,17 @@ void GuitarGraph::setupJoinMenuDifficulty() {
 	}
 	m_selectedDifficulty = ConfigItem(ol); // Create a ConfigItem from the option list
 	m_selectedDifficulty.select(cur); // Set the selection to current level
-	m_menu.add(MenuOption("", _("Select difficulty")).changer(m_selectedDifficulty)); // MenuOption that cycles the options
+	auto _selectDifficulty = std::make_unique<MenuOption>("", _("Select difficulty"));
```
OznOg: m_menu.add(std::make_unique<MenuOption>("", _("Select difficulty")).changer(m_selectedDifficulty));  ? why do you need to split that in 3 lines ?

game/guitargraph.cc
```diff
@@ -160,13 +160,17 @@ void GuitarGraph::setupJoinMenuDifficulty() {
...
+	_selectDifficulty->changer(m_selectedDifficulty);
+	m_menu.add(std::move(_selectDifficulty)); // MenuOption that cycles the options
 	m_menu.back().setDynamicName(m_difficultyOpt); // Set the title to be dynamic
 }
 
 void GuitarGraph::setupJoinMenuDrums() {
 	setupJoinMenuDifficulty();
-	m_menu.add(MenuOption(_("Lefty-mode"), "").changer(m_leftymode));
+	auto _leftyMode = std::make_unique<MenuOption>(_("Lefty-mode"), "");
+	_leftyMode->changer(m_leftymode);
```
OznOg: same

OznOg:
I don't really get the purpose of this patch. You seem to talk about not recreating things, but looks like it tries to fix some ownership problem on menu options. Could you give some more context please?
BTW, I think that using unique_ptr in menu intefaces is making the code harder to read, if you really want to use shared ptr in internal, maybe you should use some pimpl idom to "hide" it from user.

Lord-Kamina:
> I don't really get the purpose of this patch. You seem to talk about not recreating things, but looks like it tries to fix some ownership problem on menu options. Could you give some more context please? BTW, I think that using unique_ptr in menu intefaces is making the code harder to read, if you really want to use shared ptr in internal, maybe you should use some pimpl idom to "hide" it from user.

As discussed on Discord, options are being copied (which is what I was missing in my explanation) a bazillion times on master. The PR makes it so each option is built once and destroyed once, when entering and quitting the screen. 

You said this is chiefly because of my usage std::move, and that we could scrap the smart pointers. Sounds valid, I can't entirely remember whether the pointers themselves were relevant for the rest of the PR where I originally did this.

I'm currently trying to divide that into useable chunks

PR #743, opened 2022-04-29
Use weakptr to store previous selection

OznOg:
No need for custom solution to know if the song is still there, use a
weak_ptr to do it.


game/songs.cc
```diff
@@ -299,26 +299,20 @@ void Songs::reload_internal(fs::path const& parent) {
...
   public:
 	/// constructor
-	RestoreSel(Songs& s): m_s(s), m_sel(s.empty() ? nullptr : &s.current()) {}
-	/// resets song to given song
-	void reset(Song const* song = nullptr) { m_sel = song; }
+	RestoreSel(Songs& s): m_s(s), m_sel(s.currentPtr()) {}
 	~RestoreSel() {
 		int pos = 0;
-		if (m_sel) {
+		if (auto song = m_sel.lock()) {
```
Baklap4, line 313: This reads pretty weird (or i'm not used to this notation), think it'd be best to do the assigning above the if statement.

OznOg: That's the idiomatic way to use weak ptrs: you use lock to get a shared_ptr from it, and check for nullness at the same time. If null, then the object was destroyed so you know it is not there anymore, if not null you have a guaranty that it will not vanish while using it.

OznOg: see https://en.cppreference.com/w/cpp/memory/weak_ptr

Baklap4: Thanks for the explanation :)

Baklap4 (approved):
LGTM

PR #744, opened 2022-04-29
Fix/conversion warnings

Baklap4:
### What does this PR do?

Fixes conversion warnings

### Closes Issue(s)

None

### Motivation

I hated to see 200k warnings within visual studio.


yoda-jm:
I took a look and saw a lot of conversion from double to float. I would go the other way to keep the double precision

Lord-Kamina:
> I took a look and saw a lot of conversion from double to float. I would go the other way to keep the double precision

But save for a couple things, we don't need the double precision.

PR #745, opened 2022-04-30
Initialize and destroy Portaudio once

OznOg:
Portaudio requires to call init and shutdown *once* per application,
but we were calling it everythime we were changing the audio subsystem.

This patch fixes that by making sure it is called only once.


Baklap4 (approved):
LGTM

Didn't run into audio issues, however* i only have one audio-output device not multiple

PR #747, opened 2022-04-30
Instantly translate menus as lang changes

OznOg:
This pr makes it possible to change the translation of the menu
dynamicaly when changing the current language. This prevents the need to
restart or save the config.

https://user-images.githubusercontent.com/820313/166120410-3ec5750b-8b01-4f67-ad95-fd459671b03c.mp4

### DIFF
--- a/game/i18n.cc
+++ b/game/i18n.cc
@@ -5,6 +5,10 @@
 
 TranslationEngine::TranslationEngine(const char *package) : m_package(package) {
 	initializeAllLanguages();
+	/* set all languages in configuration.
+	 * They are kept untranslated internally which prevents having issues
+	 * when changing languages (and need to update internal stuff with
+	 * translations) */
 	populateLanguages(GetAllLanguages());
 
 	std::clog << "locale/debug: Checking if language is set in config." << std::endl;
@@ -43,7 +47,6 @@ void TranslationEngine::setLanguage(const std::string& language, bool fromSettin
 	try {
 		std::locale::global(m_gen(m_currentLanguage.first));
 		std::cout << "locale/notice: Current language is: '" << m_currentLanguage.second << "'" << std::endl;
-		populateLanguages(GetAllLanguages(true), true);
 	}
 	catch (std::runtime_error& e) {
 		std::clog << "locale/warning: Unable to detect locale, will try to fallback to en_US.UTF-8. Exception: " << e.what() << std::endl;
@@ -52,7 +55,7 @@ void TranslationEngine::setLanguage(const std::string& language, bool fromSettin
 }
 
 std::string TranslationEngine::getLanguageByHumanReadableName(const std::string& language) {
-	if (language == _("Auto")) {
+	if (language == "Auto") {
 		return boost::locale::util::get_system_locale(true);
 	}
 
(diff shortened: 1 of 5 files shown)


Baklap4:
Sadly things got broken in regard to localizations (at least on windows) 
The locale system detects that we're changing languages as seen in the logs:
```
add locale path: .                                                                                                      
add locale path: ./lang                                                                                                 
add locale path: C:\Users\Arjan\source\repos\performous\performous2\build\x64-debug\lang                                
locale/notice: Setting language to: 'de_DE.UTF-8'                                                                      
locale/notice: Current language is: 'German'
```

But UI-wise nothing gets changed. Saving + restart doesn't change the language either...

Baklap4 (changes requested):
See comment above

game/i18n.cc
```diff
@@ -89,25 +92,27 @@ std::map<std::string, std::string> TranslationEngine::GetAllLanguages(bool refre
...
+		{ "fr_FR.UTF-8", translate_noop("French") },
+		{ "hu_HU.UTF-8", translate_noop("Hungarian") },
+		{ "it_IT.UTF-8", translate_noop("Italian") },
+		{ "ja_JP.UTF-8", translate_noop("Japanese") },
+		{ "nl_NL.UTF-8", translate_noop("Dutch") },
+		{ "pl_PL.UTF-8", translate_noop("Polish") },
+		{ "pt_PT.UTF-8", translate_noop("Portuguese") },
+		{ "sk_SK.UTF-8", translate_noop("Slovak") },
+		{ "sv_SE.UTF-8", translate_noop("Swedish") },
+		{ "zh_CN.UTF-8", translate_noop("Chinese") }
```
Baklap4, line 115: Please refrain from using `translate_noop`. Use `_("you english translation key")` instead

`translate_noop` is solely used within the request_provider to provide the webfront-end its translations. Rethinking that, that should also use `_()` since they're just static strings

game/screen_intro.cc
```diff
@@ -181,15 +181,15 @@ void ScreenIntro::populateMenu() {
...
+			configmain.push_back(MenuOption(submenu.shortDesc, submenu.longDesc, imgConfig).submenu(opts));
 		} else {
-			configmain.push_back(MenuOption(_(submenu.shortDesc), _(submenu.longDesc), imgConfig).screen(submenu.name));
+			configmain.push_back(MenuOption(submenu.shortDesc, submenu.longDesc, imgConfig).screen(submenu.name));
 		}
 	}
-	m_menu.add(MenuOption(_("Configure"), _("Configure audio and game options."), imgConfig).submenu(configmain));
-	m_menu.add(MenuOption(_("Quit"), _("Leave the game."), imgQuit).screen(""));
+	m_menu.add(MenuOption(translate_noop("Configure"), translate_noop("Configure audio and game options."), imgConfig).submenu(configmain));
+	m_menu.add(MenuOption(translate_noop("Quit"), translate_noop("Leave the game."), imgQuit).screen(""));
```
Baklap4, line 192: Same in this file. Please do not use `translate_noop`. Use `_("dsdfdf")` instead

ooshlablu (approved):

Lord-Kamina:
Looks nice. I don't have much to add to @Baklap4's comments.

ooshlablu:
Tested it earlier with the AppImage and Packages on Ubuntu 18.04 & 20.04, seems to be consistent between all 4 of those combinations. Menus translate properly when the translation is available. Game play appears to be unaffected, tested with vocals and guitar/bass.

game/i18n.cc
OznOg: The problem is that I don't want the string to be translated here, I want it to have a translation that will be used later (when creating the menu). translate_noop seems to be the way yo do this as _() always produces a translated string. That said, I did not understood the problem with usage of this macro...

game/screen_intro.cc
OznOg: same, I don't want the menu to be created with the translated string. I need to have all internal code use english and just render it with translations (when calling opt.getValue())

game/screen_intro.cc
```diff
@@ -121,11 +121,11 @@ void ScreenIntro::draw_menu_options() {
 
 		// Regular option (not selected)
 		} else {
-			std::string title = opt.getName();
+			std::string title = _(opt.getName());
 			SvgTxtTheme& txt = getTextObject(title);
 			ColorTrans c(Color::alpha(opt.isActive() ? 1.0 : 0.5));
 			txt.dimensions.left(x).center(start_y + ii*0.065);
-			txt.draw(title);
+			txt.draw(_(title));
```
OznOg, line 128: erroneous double translation

Baklap4: The erroneous translation is still there :P

PR #748, opened 2022-04-30
Fixes a problem where localization wasn't functioning without make install

Baklap4:
### What does this PR do?

In current master when running performous from the build folder without running `make install` localization isn't working _on windows._
This is due to the fact that the localization has to be a specific format for `gettext` to understand namely:
```
.root-locale-dir
|----> en
|---------> LC_MESSAGES
|----------------> Performous.mo
|----> de
|---------> LC_MESSAGES
|----------------> Performous.mo
```

This PR fixes that issue.

### Closes Issue(s)

None

### Motivation

Wanted to try out #747 but ran into this issue.

### DIFF
--- a/lang/CMakeLists.txt
+++ b/lang/CMakeLists.txt
@@ -9,7 +9,13 @@ file(GLOB _languages *.po)
 set(_mofiles)
 foreach(_pofile ${_languages})
 	get_filename_component(_language ${_pofile} NAME_WE)
-	set(_mofile ${CMAKE_CURRENT_BINARY_DIR}/${_language}.mo)
+	if(WIN32)
+		# For some reason we need this structure on Windows to make localization work for a local build folder without install.
+		set(_mofile ${CMAKE_CURRENT_BINARY_DIR}/${_language}/LC_MESSAGES/${CMAKE_PROJECT_NAME}.mo)
+	else()
+		set(_mofile ${CMAKE_CURRENT_BINARY_DIR}/${_language}.mo)
+	endif()
+
 	add_custom_command(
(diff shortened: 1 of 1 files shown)


ooshlablu:
I test the AppImage and packages on Ubuntu 18.04 && 20.04, and everything seems to be the same on Linux.

lang/CMakeLists.txt
```diff
@@ -9,7 +9,13 @@ file(GLOB _languages *.po)
 set(_mofiles)
 foreach(_pofile ${_languages})
 	get_filename_component(_language ${_pofile} NAME_WE)
-	set(_mofile ${CMAKE_CURRENT_BINARY_DIR}/${_language}.mo)
+	if(WIN32)
+		# For some reason we need this structure on Windows to make localization work for a local build folder without install.
+		set(_mofile ${CMAKE_CURRENT_BINARY_DIR}/${_language}/LC_MESSAGES/${CMAKE_PROJECT_NAME}.mo)
```
OznOg, line 14: then why not using the same tree on linux as well?

Baklap4: Because msgfmt can't output to directories that don't exist whilst windows can !?

And if i create the directory beforehand with an extra command, that doesn't work on windows since the first output (the created folder) is set to `_mofile` which breaks msgfmt again >.>"

This way it works with the least amount of changes 😂 gotta have fun with cross platformness...

OznOg: ok

OznOg (approved):

PR #749, opened 2022-04-30
Removed SDL2 module since SDL2 comes with its own module these days

### LINKED ISSUE #632: Windows native build runtime issues

Bug/issues

### Description

This will list all runtime bugs and issues using visual studio and native builds.
So this can be picked at a later time.

### List

- [x] Menu -> Paths -> NULL reference exception. `getenv("HOME")` returns NULL on windows (fix commited #650)
- [x] Audio Devices -> Selected audio device is not remembered (but still in use and works) (fix commited in #655)
- [x] ~~Audio -> Every sample rate above 48.000 Hz is not working with a result "sample rate error" + no sound~~ (#417)
- [x] Webcam background -> Auto detect (webcam Id = -1) doesn't work as of yet. Setting this to 0 is the current workaround. (fix committed #652 )
    - [x] Auto detect was wrongly implemented by us. Autodetect has been `0` for already 11 years. So that's just a bug.
- [x] Duet songs didn't play on Windows. (fix commited #650)

### Build issues

- [x] Webserver support currently unavailable (fix commited #712)
- [x] LibRSVG -> d suffix (debug) _LIBRARY cannot be found
- [x] LibSDL2 -> d suffix (debug) sdl2d.dll doesn't get copied over after build
- [x] LibXML2 -> Library cannot be found.
- [ ] Optional: Pkgconfig -> ideally we'd like to be able to compile without pkgconfig.

Baklap4:
### What does this PR do?

Removes the SDL2 cmake module since SDL2 comes with its own cmake module these days and no extra configuration is needed.
This also resolves some build issues for Windows users having to manually copy `SDL2.dll` for debug builds (it's wrongly targetted, it should target `SDL2d.dll`)

### Closes Issue(s)

Partly closes #632

### Notes

Needs some testing on several linux distros and MacOS


ooshlablu:
Tested with the AppImage and package on Ubuntu 20.04 with only vocals and only guitar/bass individually. Everything seems to work as expected.

I would like to test this with multiple singers/instruments and both deployments on 20.04 before I give the :+1:.

OznOg (approved):

yoda-jm (approved):

PR #751, opened 2022-05-02
Prevent copies on menu option

OznOg:
Prevent many copies by preventing copies and using std::move()

Closes #742

### DIFF
--- a/game/dancegraph.cc
+++ b/game/dancegraph.cc
@@ -114,7 +114,7 @@ void DanceGraph::setupJoinMenu() {
 		}
 		m_selectedTrack = ConfigItem(ol); // Create a ConfigItem from the option list
 		m_selectedTrack.select(cur); // Set the selection to current level
-		m_menu.add(MenuOption("", _("Select track")).changer(m_selectedTrack)); // MenuOption that cycles the options
+		m_menu.add(MenuOption("", _("Select track"))).changer(m_selectedTrack); // MenuOption that cycles the options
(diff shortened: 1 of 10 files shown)


OznOg:
This is almost the same PR than https://github.com/performous/performous/pull/742/ but without the use of smart pointers, mainly for comparison.

Baklap4:
Looking at this vs #742 this looks cleaner imho

game/dancegraph.cc
```diff
@@ -114,7 +114,7 @@ void DanceGraph::setupJoinMenu() {
 		}
 		m_selectedTrack = ConfigItem(ol); // Create a ConfigItem from the option list
 		m_selectedTrack.select(cur); // Set the selection to current level
-		m_menu.add(MenuOption("", _("Select track")).changer(m_selectedTrack)); // MenuOption that cycles the options
+		m_menu.add(MenuOption{"", _("Select track")}).changer(m_selectedTrack); // MenuOption that cycles the options
```
yoda-jm, line 117: not a big fan of initializing with {} (especially when the members are not specified  (designators are C++20: https://en.cppreference.com/w/cpp/language/aggregate_initialization).

yoda-jm (dismissed):

ooshlablu:
I tested this last night with the debs and the AppImage on Ubuntu 18.04 & 20.04. Everything seems to behave normally. I didn't really notice any in-game differences which I think is supposed to be the point.


:+1:

Baklap4 (approved):

PR #754, opened 2022-05-06
Added conversion compile option for gcc and clang

Baklap4:
### What does this PR do?

This enables warnings about conversion. Within our codebase we do a lot of possible faulty conversions from double to float and other types.

### Closes Issue(s)

None

### Motivation

- Warnings are there for mostly a good reason. In our case for conversions, we have many methods using double, while we give in floats or the other way around. 
- Wanting to go warning free so we can enable treat warnings as errors.
- Windows currently complains about a lot of warnings which of many are these conversion warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,7 @@ if (APPLE)
 	endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


yoda-jm (approved):

ooshlablu (dismissed):

ooshlablu (approved):
+100

PR #755, opened 2022-05-06
Fix crash on shader error destroy

### LINKED ISSUE #750: Crash when stoping performous after trying to use 3D
https://user-images.githubusercontent.com/820313/166152661-ebcdbd54-b374-46f5-9663-bb43c64c9e26.mp4

Just activate the 3D on my laptop and exit the game, then it crashes

OznOg:
Fixes #750 

Prevent trying to delete a FBO that was invalidated and report the shader error early when creation fails.

### DIFF
--- a/game/fbo.hh
+++ b/game/fbo.hh
@@ -16,7 +16,8 @@ class FBO {
 	}
 	/// Handle clean-up
 	~FBO() {
-		if (m_fbo) glDeleteFramebuffers(1, &m_fbo);
+		if (glIsFramebuffer(m_fbo))
+			glDeleteFramebuffers(1, &m_fbo);
 	}
 	/// Returns a reference to the attached texture
 	OpenGLTexture<GL_TEXTURE_2D>& getTexture() {
--- a/game/menu.cc
+++ b/game/menu.cc
@@ -70,9 +70,15 @@ void Menu::action(int dir) {
 					auto &value = config["game/language"];
 					Game::getSingletonPtr()->setLanguage(value.getValue());
 				} else if (current().value->getName() == "graphic/stereo3d") {
-					std::clog << "video/info: Stereo 3D configuration changed, will reset shaders." << std::endl;
-					Game::getSingletonPtr()->window().resetShaders();
-				}
+					try {
+						std::clog << "video/info: Stereo 3D configuration changed, will reset shaders.\n";
+						Game::getSingletonPtr()->window().resetShaders();
+					} catch (const std::exception &) {
+						std::cerr << "video/info: Disable Stereo 3D because shader creation failed.\n";
+						current().value->b() = false; // Disable 3D if shader fails
+						throw;
+					}
+                                }
 			}
 			break;
 		}
--- a/game/menu.cc
+++ b/game/menu.cc
@@ -61,8 +61,7 @@ void Menu::action(int dir) {
 					current().value->setOldValue(current().value->getValue());
 				}
 				else if (current().value->getName() == "graphic/stereo3d") {
-					std::string oldValue((current().value->getValue() == _("Disabled")) ? "0" : "1");
-					current().value->setOldValue(oldValue);
+					current().value->setOldValue(current().value->getValue());
 				}
 				if (dir > 0) ++(*(current().value));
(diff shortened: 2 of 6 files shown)


Baklap4:
I've canceled the pipeline since we've made a breaking change for ubuntu 1804 (pr will open in a bit)

OznOg:
> I've canceled the pipeline since we've made a breaking change for ubuntu 1804 (pr will open in a bit)

Scandal!

game/fbo.hh
```diff
@@ -16,7 +16,8 @@ class FBO {
 	}
 	/// Handle clean-up
 	~FBO() {
-		if (m_fbo) glDeleteFramebuffers(1, &m_fbo);
+            if (glIsFramebuffer(m_fbo) == GL_TRUE)
```
Baklap4, line 19: This can be shortened to `if (glIsFrambuffer(m_fbo)` right?

game/menu.cc
```diff
@@ -61,16 +61,24 @@ void Menu::action(int dir) {
...
 					Game::getSingletonPtr()->setLanguage(value.getValue());
-				}
+				} else if (current().value->getName() == "graphic/stereo3d") {
+					try {
+						std::clog << "video/info: Stereo 3D configuration changed, will reset shaders.\n";
+						Game::getSingletonPtr()->window().resetShaders();
+					} catch (const std::exception &) {
+						std::cerr << "video/info: Disable Stereo 3D because shader creation failed.\n";
+						current().value->b() = false; // Disable 3D if shader fails
+						throw;
```
Baklap4, line 79: By rethrowing don't we get a runtime exception?

game/fbo.hh
Lord-Kamina: I don't think `GLBooleans` are implemented as actual bool so no. Unless you want to get ants? Because that's how you get ants.

OznOg: well Yes and no: hopefully, GL_TRUE is !=0 but who can really say it will always be the case?...

Baklap4: Fair enough

game/menu.cc
OznOg: nop, this is caught from the above exception handler and display an error message on the window

game/fbo.hh
OznOg: that said comparing to TRUE is usually a bad idea

OznOg: done

Baklap4:
Hmm while on master: 

- enabling stereoscopic 3d
- Save config
- Something went wrong with the linkin the shader
- Neglecting the error and moving to the songbrowser i do however see the image below clearly indicating that its working

image

With this PR i cannot get stereoscopic 3d to be enabled resulting in never be able to see the songbrowser as above

OznOg:
from discussion on Discord we agreed to integrate @Lord-Kamina 's patch that fixes the shaders' compilation to this PR

Baklap4 (dismissed):
Fixes linked bug. Is there also a bug for the compilation of the shaders? If yes, please link these here aswell so it gets closed

OznOg:
> Hmm while on master:
> 
>     * enabling stereoscopic 3d
> 
>     * Save config
> 
>     * Something went wrong with the linkin the shader
> 
>     * Neglecting the error and moving to the songbrowser i do however see the image below clearly indicating that its working
> 
>     * Quiting the game results into mentioned error in the linked issue
> 
> 
> image
> 
> With this PR i cannot get stereoscopic 3d to be enabled resulting in never be able to see the songbrowser as above

Even if code could "almost work" in case of shader error, I think it is better not to keep half working things going on. Disabling the 3D in case of error is a bit radical, but I think it is the best way to prevent unexpected behaviours and funky bugs. I think we should fix errors in the first place.

Baklap4:
> Even if code could "almost work" in case of shader error, I think it is better not to keep half working things going on. Disabling the 3D in case of error is a bit radical, but I think it is the best way to prevent unexpected behaviours and funky bugs. I think we should fix errors in the first place.

Agreed, but for this there was already a fix at hand, so there was no need to 'break' functionality. But in general yes for sure!

PR #756, opened 2022-05-06
Fix a breaking change in ubuntu 1804 docker image

Baklap4:
### What does this PR do?

Fixes the Ubuntu1804 build since we made a breaking change in https://github.com/performous/performous-docker/pull/14
We need to add `CC` and `CXX` variables and set them correctly for ubuntu1804 since ubuntu1804 ships with GCC 7, while we need 8 for `std::filesystem`

### Closes Issue(s)

None


ooshlablu (approved):
+100

PR #760, opened 2022-05-11
Run PR Commenter when workflow is completed.

Baklap4:
### What does this PR do?

This allows us to download artifacts if a run is completed (but with failures).
It should give us valid links for every artifact that has been made


ooshlablu (approved):
:+1:

Issue #671, opened 2021-09-07
Rework audio devices SampleRate


Baklap4:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

I didn't expect performous to crash when i didn't supply a correct sample rate.

### Query PortAudio

You can query portaudio and check if the supplied samplerate works or not on a specific device.
[Some example code](http://www.portaudio.com/docs/v19-doxydocs/querying_devices.html):
```
const PaStreamParameters *inputParameters;
const PaStreamParameters *outputParameters;
double desiredSampleRate;
...
PaError err;
err = Pa_IsFormatSupported( inputParameters, outputParameters, desiredSampleRate );
if( err == paFormatIsSupported )
{
   printf( "Hooray!\n");
}
else
{
   printf("Too Bad.\n");
}
```

We should use this method before opening an audio device.

### Notes

Marking this as Bug and as Enhancement

OznOg:
what is the expected behaviour? do we wan to reject the device or try some down-sampling?

Baklap4:
Well a device supports a couple of samplerates
The problem we're having right now is that Performous always tries to open devices with a samplerate of 48000 Hz.

If the device we're trying to open doesn't support it (and/or is currently configured by the OS to use a different samplerate let's say 16000 Hz). Then we're getting the error "invalid samplerate". Whilst the samplerate is in fact supported by the device in question

![image](https://user-images.githubusercontent.com/3918517/168151166-3dd1fedf-9965-4044-8f2a-dcd019d16300.png)

As seen in the image above the samplerate for my microphone supports 3 values: 48000Hz (this one works), 32000Hz (doesn't work), 16000Hz (doesn't work)

I'd expect that the supported rates could be used by performous, with the default being the value the OS supplies. With some testing on windows i did see PortAudio _can_ detect that i configured my mic to be at 16000Hz for the sake of testing

Issue #417, opened 2018-11-01
Invalid sample rate when trying to use Webcam's microphone


jose1711:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

- `pasuspender performous`
- in GUI select USB webcam as a microphone input
- get error: Some device failed to open
- on CLI read: `audio/error: Audio device 'dev="USB Device 0x46d:0x9a4: Audio" mics=green': Pa_OpenStream failed: Invalid sample rate`

### Output of `performous --version`: (_What version of Performous are you using?_)

```
Performous 1.1+
```

### What is your environment & configuration (arguments, platform, ...)?

- Ubuntu 18.04.1 LTS
- Webcam Logitech QuickCam E 3500
```
cat /proc/asound/card0/usbmixer 
USB Mixer: usb_id=0x046d09a4, ctrlif=2, ctlerr=1
Card: USB Device 0x46d:0x9a4 at usb-0000:00:1a.0-1.4, high speed
  Unit: 5
    Control: name="Mic Capture Volume", index=0
    Info: id=5, control=2, cmask=0x0, channels=1, type="S16"
    Volume: min=6080, max=8768, dBmin=2375, dBmax=3425
  Unit: 5
    Control: name="Mic Capture Switch", index=0
    Info: id=5, control=1, cmask=0x0, channels=1, type="INV_BOOLEAN"
    Volume: min=0, max=1, dBmin=0, dBmax=0

cat /proc/asound/card0/stream0 
USB Device 0x46d:0x9a4 at usb-0000:00:1a.0-1.4, high speed : USB Audio

Capture:
  Status: Stop
  Interface 3
    Altset 1
    Format: S16_LE
    Channels: 1
    Endpoint: 6 IN (ASYNC)
    Rates: 16000
    Data packet interval: 1000 us

```

Lord-Kamina:
Can you try and check whether it's actually our problem and not portaudio's?

jose1711:
according to http://portaudio.com/docs/v19-doxydocs/api_overview.html this can very well be the case:

> PortAudio doesn't provide sample rate conversion if you request a sample rate that is not supported by the native audio API

i wonder if it's possible to workaround this by editing `asoundrc`.

nieknooijens:
libswresample provides resampling...

nieknooijens:
[...]

Issue #160, opened 2015-09-10
Audio device configuration layout...


TobiX:
When PortAudio has many backends enabled (on MXE currently wmme,directx,wdmks,wasapi) or when there are many audio devices, the device configuration screen is almost unusable.

This is on my Windows desktop with just one sound device (& USB microphones):
![performous_20150910t193016](https://cloud.githubusercontent.com/assets/78534/9795929/a70fef86-57f3-11e5-953b-7fe271f50fc0.png)

I'm not sure how to make this better:
 - We could just improve the presentation (make the screen scrollable)
 - We could ignore specific backends (If one is better then the other) - Which ones?
 - We could add another setting before the audio device selection: audio backend
 - Something else?

We at least need to resolve the ambiguity in device naming, currently my list of devices looks like this:
```
audio/info: PortAudio devices:
  #0 Microsoft Soundmapper - Input (2 in)
  #1 FrontMic (Realtek High Definiti (2 in)
  #2 Mikrofon (2- Wireless Mic #1220 (2 in)
  #3 Microsoft Soundmapper - Output (2 out)
  #4 Lautsprecher (Realtek High Defi (2 out)
  #5 Realtek Digital Output (Realtek (2 out)
  #6 S22C350-4 (NVIDIA High Definiti (2 out)
  #7 Realtek HD Audio 2nd output (Re (2 out)
  #8 PrimÃ¤rer Soundaufnahmetreiber (2 in)
  #9 FrontMic (Realtek High Definition Audio) (2 in)
  #10 Mikrofon (2- Wireless Mic #122038319  ) (2 in)
  #11 PrimÃ¤rer Soundtreiber (2 out)
  #12 Lautsprecher (Realtek High Definition Audio) (2 out)
  #13 Realtek Digital Output (Realtek High Definition Audio) (2 out)
  #14 S22C350-4 (NVIDIA High Definition Audio) (2 out)
  #15 Realtek HD Audio 2nd output (Realtek High Definition Audio) (2 out)
  #16 Realtek Digital Output (Realtek High Definition Audio) (2 out)
  #17 Lautsprecher (Realtek High Definition Audio) (2 out)
  #18 S22C350-4 (NVIDIA High Definition Audio) (2 out)
  #19 Realtek HD Audio 2nd output (Realtek High Definition Audio) (2 out)
  #20 Mikrofon (2- Wireless Mic #122038319  ) (2 in)
  #21 FrontMic (Realtek High Definition Audio) (2 in)
[...]

PR #761, opened 2022-05-11
Rework audio backend

OznOg:
Handling of the audio backend selection is quite complicated and does not allow to test the backend withour restart. This PR should allow it.


game/libda/portaudio.cc
```diff
@@ -0,0 +1,120 @@
...
+
+	struct AudioBackendFactory::Init {
+		Init() { PORTAUDIO_CHECKED(Pa_Initialize, ());
+			for (int i = 0; i < Pa_GetHostApiCount(); ++i) {
+				backends.emplace_back(Pa_GetHostApiInfo(i));
+			}
+			if (backends.empty()) throw std::runtime_error("No suitable audio backends found.");
+		}
+		~Init() { Pa_Terminate(); }
+		std::vector<const PaHostApiInfo *> backends;
```
Lord-Kamina, line 79: Are we certain PortAudio cleans this appropriately? If not, maybe use a smart pointer?

game/libda/portaudio.cc
```diff
@@ -0,0 +1,120 @@
+#include "portaudio.hpp"
+
+#include "../platform.hh"
+
+namespace portaudio {
+
+	struct AudioBackend::AudioDevices {
+		/// Constructor gets the PA devices into a vector
+		AudioDevices(PaHostApiTypeId backend) {
+			auto backendIndex = Pa_HostApiTypeIdToHostApiIndex(backend);
```
Lord-Kamina, line 10: Please try not to use `auto` for these things, as it turns out it's not as intelligent as it should be sometimes.
Prefer portaudio's type aliases in this case.

Baklap4:
@OznOg the build contains a couple failures on Windows and MacOS

Converting this to Draft + in progress for now

game/libda/portaudio.cc
OznOg: The raw pointers are her on purpose, this is a non ownership case: PaHostApiInfo are provided and owned by portaudio and *must* not be deleted by user. For now, there is no smart pointer to indicate non ownership (discussions are still going on in the standard).

Lord-Kamina: Fair enough

OznOg:
Gents, this is still draft, I'm trying to address  #671, and  #417 and maybe #160 at some point... review those at your own risks, I may completely rewrite any part of this MR.

Lord-Kamina:
The build failure on Mac is probably cause you forgot to `#include <stringstream>`

Lord-Kamina:
> Gents, this is still draft, I'm trying to address #671, and #417 and maybe #160 at some point... review those at your own risks, I may completely rewrite any part of this MR.

I don't think there's anything to be done for #160.

game/libda/portaudio.cc
twollgam: IMO not true. Left side is always the same type as on the right side. `auto` is the right decision even for external libs.

PR #763, opened 2022-05-13
Changed the Hash struct so it doesn't collide with std::hash

Baklap4:
### What does this PR do?

Fixes the hash struct to not collide with `std::hash` causing a build failure on windows.
This became a problem since visual studio 17.2.0 update: https://devblogs.microsoft.com/visualstudio/visual-studio-2022-17-2-is-now-available/

### Closes Issue(s)

Fixes these errors:
image


ooshlablu (approved):
github actions related files lgtm

ooshlablu:
I did some quick testing with this; it loads, plays songs, and doesn't have any unusual errors in the info log (I always have some files that are broken and those appear to be the same, which is good for this)

I'm hoping to get some more testing in tomorrow, but everything looks as it should for now.

game/fs.hh
```diff
@@ -99,8 +99,8 @@ Paths const& getPaths();  ///< Get the data file search path
 Paths getThemePaths();  ///< Get the data/theme file search path (includes current and default themes in addition to data folders)
 Paths getPathsConfig(std::string const& confOption);  ///< Return expanded list of paths specified by a path config option
 
-template <>
-struct std::hash<fs::path>
+//template <>
```
OznOg, line 102: remove dead template

OznOg (changes requested):

game/fs.hh
OznOg: please

ooshlablu:
Tested the debs and the AppImage on Ubuntu 18.04 and 20.04 with guitar and vocals, everything works as expected.

game/fs.hh
Baklap4: I have no idea where this dead code is 😆 (fixed)

OznOg (approved):

Issue #603, opened 2021-07-11
cpprest and libicu incompatibility


deuteragenie:
clang static analyzer detects some incompatibility between cpprest (v2.10.18-1build) and libicu (v67.1), in 10 files.

Example:

```
In file included from ../../../game/webserver.cc:1:
In file included from ../../../game/webserver.hh:5:
In file included from ../../../game/requesthandler.hh:7:
In file included from ../../../game/screen_playlist.hh:4:
In file included from ../../../game/screen.hh:6:
In file included from ../../../game/opengl_text.hh:5:
In file included from ../../../game/unicode.hh:6:
In file included from /usr/include/unicode/tblcoll.h:65:
In file included from /usr/include/unicode/utypes.h:38:
In file included from /usr/include/unicode/umachine.h:46:
In file included from /usr/include/unicode/ptypes.h:52:
[1m/usr/include/unicode/platform.h:818:29: [0m[0;1;31merror: [0m[1mtoo few arguments provided to function-like macro invocation[0m
                            UPRV_HAS_DECLSPEC_ATTRIBUTE(dllimport))
[0;1;32m                            ^
[0m[1m/usr/include/unicode/platform.h:438:69: [0m[0;1;30mnote: [0mexpanded from macro 'UPRV_HAS_DECLSPEC_ATTRIBUTE'[0m
#   define UPRV_HAS_DECLSPEC_ATTRIBUTE(x) __has_declspec_attribute(x)
[0;1;32m                                                                    ^
[0m[1m/usr/include/unicode/platform.h:843:29: [0m[0;1;31merror: [0m[1mtoo few arguments provided to function-like macro invocation[0m
                            UPRV_HAS_DECLSPEC_ATTRIBUTE(dllimport))
[0;1;32m                            ^
[0m[1m/usr/include/unicode/platform.h:438:69: [0m[0;1;30mnote: [0mexpanded from macro 'UPRV_HAS_DECLSPEC_ATTRIBUTE'[0m
#   define UPRV_HAS_DECLSPEC_ATTRIBUTE(x) __has_declspec_attribute(x)
[0;1;32m                                                                    ^
[0m2 errors generated.

Building without webserver support eliminates all above errors.

Lord-Kamina:
I reported this ages ago at ICU. It was supposedly fixed but maybe in a more recent version (or maybe the fix doesn't work).
The issue is caused by horrible macros.

Baklap4:
[...]

PR #766, opened 2022-05-17
Rework webserver

yoda-jm:
### What does this PR do?

This enable webserver for all people using cpp-httplib that is easier to install (no dependency)

### Additional Notes

There is still a fix to do in the CMake file in order to fetch it from a clone of cpp-httplib that we'll have on performous project.


Lord-Kamina:
I can't test until I get my laptop back. That being said, there are some things from my PR that I would very much like to see here, such as:

* Refactoring in mime-type handling
* Decoupling of the search between the webserver UI and the game's actual interface.
* Ability to connect to webserver from local network only (i.e. define a subnet mask that allows more than localhost, but doesn't leave the computer open to the world)

There might be others, but that's what I remember now.

Also, I'm still not fully convinced that using a synchronous library will really work as expected.

Baklap4:
> That being said, there are some things from my PR that I would very much like to see here, such as:
> * Refactoring in mime-type handling
> * Decoupling of the search between the webserver UI and the game's actual interface.
> * Ability to connect to webserver from local network only (i.e. define a subnet mask that allows more than localhost, but doesn't leave the computer open to the world)

I think it'd be wise to add this into separate pr's as this pr is currently only porting to another library

cmake/Modules/FindWebserver.cmake
```diff
@@ -0,0 +1,47 @@
...
+include(LibFindMacros)
+include(LibFetchMacros)
+
+set(Webserver_GIT_VERSION "v0.10.7")
+
+if(SELF_BUILT_WEBSERVER STREQUAL "ALWAYS")
+	message(STATUS "Webserver forced to build from source")
+	libfetch_git_pkg(Webserver
+		#REPOSITORY https://github.com/yhirose/cpp-httplib.git
+		REPOSITORY ${SELF_BUILT_GIT_BASE}/cpp-httplib.git
```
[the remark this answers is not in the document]
Baklap4: Performous organization has that repo now forked 👍

Lord-Kamina:
> > That being said, there are some things from my PR that I would very much like to see here, such as:
> 
> > * Refactoring in mime-type handling
> 
> > * Decoupling of the search between the webserver UI and the game's actual interface.
> 
> > * Ability to connect to webserver from local network only (i.e. define a subnet mask that allows more than localhost, but doesn't leave the computer open to the world)
> 
> 
> 
> I think it'd be wise to add this into separate pr's as this pr is currently only porting to another library

What about the subnet thing? I'm looking into whether httplib supports it at all but I can't find any info.

game/webserver.cc
```diff
@@ -1,22 +1,21 @@
 #include "webserver.hh"
 #include "game.hh"
 
-#ifdef USE_WEBSERVER
 #include <boost/asio.hpp>
```
Baklap4, line 4: Hmm this should be installed and be listed within the vcpkg.json
-> `boost-asio`

if cppnetlib has a method to get the IP address we can also get rid of this include / dependency

game/requesthandler.cc
```diff
@@ -1,376 +1,235 @@
 #include "requesthandler.hh"
 #include "unicode.hh"
 #include "game.hh"
-
-#ifdef USE_WEBSERVER
```
Baklap4, line 5: do we want to include the webserver within the core functionality? If yes, then this is the way to go. If no then we still have to support it as a 'module' being captured within `USE_WEBSERVER`

game/requesthandler.cc
```diff
@@ -1,376 +1,235 @@
...
+				{"Title", song->title},
+				{"Artist", song->artist},
+				{"Edition", song->edition},
+				{"Language", song->language},
+				{"Creator", song->creator},
+				{"name", song->artist + " " + song->title},
+				{"Duration", song->getDurationSeconds()},
+			});
+		}
+		res.set_content(jsonRoot.dump(), "application/json");
```
Baklap4, line 73: If no status is set it defaults to `200 OK`? I think it'd be nice to always set it explicitly (for each api endpoint)

game/requesthandler.cc
```diff
@@ -1,376 +1,235 @@
...
+	});
+
+	m_server.Post("/api/add", [&](const httplib::Request &, httplib::Response& res, const httplib::ContentReader &contentReader) {
+		std::string buffer;
+		contentReader([&](const char *data, size_t dataLength) {
+			buffer.append(data, dataLength);
+			return true;
+		});
+		const auto jsonRoot = nlohmann::json::parse(buffer);
+		for(int i = 0; i< m_songs.size(); i++) {
```
Baklap4, line 88: Can't we use a range-based loop here?

game/requesthandler.cc
```diff
@@ -1,376 +1,235 @@
...
+		std::string buffer;
+		contentReader([&](const char *data, size_t dataLength) {
+			buffer.append(data, dataLength);
+			return true;
+		});
+		const auto jsonRootReq = nlohmann::json::parse(buffer);
+		const std::string query = jsonRootReq["query"];
+		m_songs.setFilter(query);
+		auto jsonRoot = nlohmann::json::array();
+		for (size_t i=0; i< m_songs.size(); i++) {
```
Baklap4, line 176: can't we use range-based loop here?

.github/workflows/build_and_release.yml
```diff
@@ -157,7 +157,7 @@ jobs:
           cd build_cmake
           cmake ${EXTRA_CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release \
             -DPERFORMOUS_VERSION=$PACKAGE_VERSION \
-            -DENABLE_WEBSERVER=ON \
+            -DSELF_BUILT_WEBSERVER=ALWAYS \
```
Baklap4, line 160: We're missing the change within the OSX-Utils/performous-app-build.sh

data/themes/default/www/js/documentloader.js
```diff
@@ -16,7 +16,7 @@ $.loadScript = function (url, callback) {
     The website is now fully useable.
 */
 function loadSearchJs() {
-    $("#custom-scripts-search").load("search.js", function () {
+    $("#custom-scripts-search").load("js/search.js", function () {
```
Baklap4, line 19: Not quite sure why we would need this `js/` prefix here? The other `*.js` files live in the same folder as `documentloader.js`

Baklap4 (changes requested):
A couple of questions + some changes are needed

game/songs.cc
```diff
@@ -51,7 +51,8 @@ void Songs::reload_internal() {
 		m_songs.clear();
 		m_dirty = true;
 	}
-#ifdef USE_WEBSERVER
+#if 0
```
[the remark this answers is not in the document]
Baklap4: What is your current concern about the cache feature? And did you also look into https://github.com/performous/performous/pull/721 this pr makes the cache even better

Baklap4:
Added both #603 & #452 as linked issues as these will be closed by merging this.

game/webserver.cc
Baklap4: Confirmed: Adding `boost-asio` or an equivalent to vcpkg.json will fix the build.

However i do get weird errors regarding caching on windows:
```
songs/error: Could not save C:\Users\Arjan\AppData\Roaming\performous\cache\songs.json: [json.exception.type_error.316] invalid UTF-8 byte at index 45: 0x72    
```

game/requesthandler.cc
Lord-Kamina: We can if we implement custom iterators for the class, as I did on my PR.

Baklap4:
@yoda-jm any objections that i take over this branch so i can make it up to date again?

yoda-jm:
> @yoda-jm any objections that i take over this branch so i can make it up to date again?

Please go ahead :)

PR #769, opened 2022-05-20
Switch to using float wherever possible, fix all conversion warnings.

Lord-Kamina:
### What does this PR do?

This PR switches to float wherever we can without altering the game. i.e., keep double for things dealing with pitch/frequency/fft/etc and also for calculations involving time and animations (although this would probably not be affected much by the change either)

It also fixes essentially all warnings I could find in the build.

This is essentially my version of #744 

### Motivation

1) Warnings are ugly.
2) We don't really need to use double for most things. Initially my motivation was the fact OpenGL uses float internally but after investigating and doing some actual benchmarking, I found floats to be about as performant (in some cases, faster) than using double and, they use less memory. 

### More

### Additional Notes

Ideally we want to test quite a bit, because I managed to introduce two not-minor and not very intuitive bugs doing this (I already fixed them, of course)

### DIFF
--- a/game/audio.cc
+++ b/game/audio.cc
@@ -267,7 +267,10 @@ struct Sample {
 			// No more data to play in this sample
 			return;
 		}
-		if (end <= begin) return;
+		if (end <= begin) {
+			eof = true;
+			return;
+		}
 		std::int64_t size = end - begin;
 		std::vector<float> mixbuf(static_cast<size_t>(size));
 		if(!audioBuffer.read(mixbuf.data(), size, m_pos, 1.0)) {
@@ -295,6 +298,7 @@ struct Synth {
 		static double phase = 0.0;
 		for (float *i = begin; i < end; ++i) *i *= 0.3f; // Decrease music volume
 		std::int64_t size = end - begin;
+		if (end <= begin) return;
 		std::vector<float> mixbuf(static_cast<size_t>(size));
 		Notes::const_iterator it = m_notes.begin();
 
--- a/game/backgrounds.cc
+++ b/game/backgrounds.cc
@@ -61,7 +61,7 @@ void Backgrounds::reload_internal(fs::path const& parent) {
 				m_dirty = true;
 			}
 		}
-	} catch (std::exception const& e) {
+	} catch (std::exception const&) {
 		std::cout << "Error accessing " << parent << std::endl;
 	}
 }
--- a/game/backgrounds.cc
+++ b/game/backgrounds.cc
@@ -61,8 +61,8 @@ void Backgrounds::reload_internal(fs::path const& parent) {
 				m_dirty = true;
 			}
 		}
-	} catch (std::exception const&) {
-		std::cout << "Error accessing " << parent << std::endl;
+	} catch (std::exception const& e) {
+		std::cout << "Exception thrown: " + std::string(e.what()) + ", error accessing " + parent.string() << std::endl;
 	}
 }
 
--- a/game/configuration.cc
+++ b/game/configuration.cc
@@ -39,6 +39,7 @@ ConfigItem& ConfigItem::incdec(int dir) {
 		int& val = std::get<int>(m_value);
 		int step = std::get<int>(m_step);
 		val = clamp(((val + dir * step)/ step) * step, std::get<int>(m_min), std::get<int>(m_max));
+	} else if (m_type == "uint") {
 		unsigned short& val = std::get<unsigned short>(m_value); 
 		int value = static_cast<int>(val);
 		int step = std::get<unsigned short>(m_step);
@@ -97,7 +98,8 @@ void ConfigItem::select(unsigned short i) { verifyType("option_list"); m_sel = c
 namespace {
 	template <typename T, typename VariantAll, typename VariantNum> std::string numericFormat(VariantAll const& value, VariantNum const& multiplier, VariantNum const& step) {
 		// Find suitable precision (not very useful for integers, but this code is generic...)
-		T s = std::abs<T>(std::get<T>(multiplier) * std::get<T>(step));
+		T m = std::get<T>(multiplier);
+		T s = std::abs<T>(m * std::get<T>(step));
 		int precision = 0;
 		while (s > static_cast<T>(0) && (s *= static_cast<T>(10)) < static_cast<T>(10)) ++precision;
 		// Not quite sure how to format this with FMT
@@ -147,8 +149,11 @@ std::string const ConfigItem::getValue() const {
 		return languageName;
 	}
 	if (m_type == "int") {
-		int val = std::get<int>(m_value);
 		return numericFormat<int>(m_value, m_multiplier, m_step) + _(m_unit);
+	}
+	if (m_type == "uint") {
+		unsigned val = ui();
+		if (val < m_enums.size()) return m_enums[val];
 		return numericFormat<unsigned short>(m_value, m_multiplier, m_step) + _(m_unit);
 	}
 	if (m_type == "float") return numericFormat<float>(m_value, m_multiplier, m_step) + _(m_unit);
@@ -159,8 +164,7 @@ std::string const ConfigItem::getValue() const {
 		return sl.size() == 1 ? "{" + sl[0] + "}" : fmt::format(_("{:d} items"), sl.size());
 	}
 	if (m_type == "option_list") {
-		if (m_sel <= -1) throw std::logic_error("Invalid option selected at index: " + std::to_string(m_sel));
-		return std::get<OptionList>(m_value).at(static_cast<size_t>(m_sel));
+		return std::get<OptionList>(m_value).at(m_sel);
 	}
 	throw std::logic_error("ConfigItem::getValue doesn't know type '" + m_type + "'");
 }
@@ -187,7 +191,7 @@ namespace {
 }
 
 void ConfigItem::addEnum(std::string name) {
-	verifyType("int");
+	verifyType("uint");
 	if (find(m_enums.begin(),m_enums.end(),name) == m_enums.end()) {
 		m_enums.push_back(name);
 	}
@@ -205,7 +209,7 @@ void ConfigItem::selectEnum(std::string const& name) {
 
 
 std::string const ConfigItem::getEnumName() const {
-	int val = i();
+	unsigned val = ui();
 	if (val < m_enums.size()) { return m_enums[val]; }
 	else { return std::string(); }
 }
@@ -245,6 +249,8 @@ void ConfigItem::update(xmlpp::Element& elem, unsigned mode) try {
 		m_longDesc = getText(elem, "long");
 	} else {
 		std::string type = getAttribute(elem, "type");
+		if (type == "int" && m_type == "uint") type = "uint"; // Convert old config values.
+		std::clog << "configuration/info: Converting config value: " + getAttribute(elem, "name") + ", to type uint." << std::endl;
 		if (!type.empty() && type != m_type) throw std::runtime_error("Entry type mismatch: " + getAttribute(elem, "name") + ": schema type = " + m_type + ", config type = " + type);
 	}
 	if (m_type == "bool") {
@@ -258,6 +264,8 @@ void ConfigItem::update(xmlpp::Element& elem, unsigned mode) try {
 		std::string value_string = getAttribute(elem, "value");
 		if (!value_string.empty()) m_value = std::stoi(value_string);
 		updateNumeric<int>(elem, mode);
+	} else if (m_type == "uint") {
+		std::string value_string = getAttribute(elem, "value");
 		if (!value_string.empty()) m_value = static_cast<unsigned short>(std::stoi(value_string));
 			// Enum handling
 			if (mode == 0) {
@@ -341,6 +349,7 @@ void writeConfig(bool system) {
 			}
 		}
 		else if (type == "int") entryNode->set_attribute("value",std::to_string(item.i()));
+		else if (type == "uint") entryNode->set_attribute("value",std::to_string(item.ui()));
 		else if (type == "bool") entryNode->set_attribute("value", item.b() ? "true" : "false");
 		else if (type == "float") entryNode->set_attribute("value",std::to_string(item.f()));
 		else if (item.get_type() == "string") xmlpp::add_child_element(entryNode, "stringvalue")->add_child_text(item.s());
@@ -482,7 +491,7 @@ void readConfig() {
 	// Populate themes
 	ConfigItem& ci = config["game/theme"];
 	for (std::string const& theme: getThemes()) ci.addEnum(theme);
-	if (ci.i() == -1) ci.selectEnum("default");  // Select the default theme if nothing is selected
+	if (ci.ui() == 1337) ci.selectEnum("default");  // Select the default theme if nothing is selected
 }
 
 void populateBackends (const std::list<std::string>& backendList) {
--- a/game/configuration.hh
+++ b/game/configuration.hh
@@ -17,6 +17,7 @@ class ConfigItem {
 	ConfigItem() : m_sel() {};
 	explicit ConfigItem(bool bval);
 	explicit ConfigItem(int ival);
+	explicit ConfigItem(unsigned short uival);
 	explicit ConfigItem(float fval);
 	ConfigItem(std::string sval);
 	ConfigItem(OptionList opts);
@@ -28,6 +29,8 @@ class ConfigItem {
 	std::string get_type() const { return m_type; } ///< get the field type
 	int& i(); ///< Access integer item
 	int const& i() const; ///< Access integer item
+	unsigned short& ui(); ///< Access unsigned integer item
+	unsigned short const& ui() const; ///< Access unsigned integer item
 	bool& b(); ///< Access boolean item
 	float& f(); ///< Access floating-point item
 	std::string& s(); ///< Access string item
@@ -56,14 +59,15 @@ class ConfigItem {
 	std::string m_shortDesc;
 	std::string m_longDesc;
 
-	typedef std::variant<bool, int, float, std::string, StringList> Value;
+	typedef std::variant<bool, unsigned short, int, float, std::string, StringList> Value;
+// I wanted to allow actual conversion (where possible without changing values erroneously) between i() and ui() but it's not as simple to do as I thought. verifyType() is going bye-bye anyway, so maybe there's a better way to implement it properly with the eventual std::visit implementations.
 	bool isDefaultImpl(Value const& defaultValue) const;
 	Value m_value; ///< The current value
 	Value m_factoryDefaultValue; ///< The value from config schema
 	Value m_defaultValue; ///< The value from config schema or system config
 	std::string m_oldValue; ///< A previous value, as output by getValue().
 	std::vector<std::string> m_enums; ///< Enum value titles
-	std::variant<int, float> m_step, m_min, m_max, m_multiplier;
+	std::variant<int, unsigned short, float> m_step, m_min, m_max, m_multiplier;
 	std::string m_unit;
 	unsigned short m_sel;
 };
--- a/game/controllers-joystick.cc
+++ b/game/controllers-joystick.cc
@@ -40,7 +40,7 @@ namespace input {
 			else return false;
 			try {
 			event.source = SourceId(SourceType::JOYSTICK, static_cast<unsigned>(sdlEv.jbutton.which));  // All j* structures have .which at the same position as jbutton
-			} catch (std::exception const& e) {
+			} catch (std::exception const&) {
 				std::clog << "joystick/error: " + std::to_string(sdlEv.jbutton.which) + " is an invalid SDL_JoystickID." << std::endl;
 			}
 			return true;
--- a/game/controllers-joystick.cc
+++ b/game/controllers-joystick.cc
@@ -76,10 +76,10 @@ void init() {
 		ci2.addEnum(title);
 		ci3.addEnum(title);
 		// Check for active items
-		if (i == ci0.i()-1) instruments.push_back("0:" + it->first);
-		if (i == ci1.i()-1) instruments.push_back("1:" + it->first);
-		if (i == ci2.i()-1) instruments.push_back("2:" + it->first);
-		if (i == ci3.i()-1) instruments.push_back("3:" + it->first);
+		if (i == ci0.ui()-1) instruments.push_back("0:" + it->first);
+		if (i == ci1.ui()-1) instruments.push_back("1:" + it->first);
+		if (i == ci2.ui()-1) instruments.push_back("2:" + it->first);
+		if (i == ci3.ui()-1) instruments.push_back("3:" + it->first);
 	}
 
 	// Check all forced instruments
--- a/game/controllers-midi.cc
+++ b/game/controllers-midi.cc
@@ -39,9 +39,9 @@ namespace input {
 			PmEvent ev;
 			for (auto it = m_streams.begin(); it != m_streams.end();++it) {
 				if (Pm_Read(*it->second, &ev, 1) != 1) continue;
-				unsigned char evnt = static_cast<unsigned char>(ev.message & 0xF0);
-				unsigned char note = static_cast<unsigned char>(Pm_MessageData1(ev.message));
-				unsigned char vel  = static_cast<unsigned char>(Pm_MessageData2(ev.message));
+				auto evnt = static_cast<unsigned char>(Pm_MessageStatus(ev.message) & 0xF0);
+				auto note = static_cast<unsigned char>(Pm_MessageData1(ev.message));
+				auto vel  = static_cast<unsigned char>(Pm_MessageData2(ev.message));
 				unsigned chan = (ev.message & 0x0F) + 1;  // It is conventional to use one-based indexing
 				if (evnt == 0x80 /* NOTE OFF */) { evnt = static_cast<unsigned char>(0x90); vel = 0; }  // Translate NOTE OFF into NOTE ON with zero-velocity
 				if (evnt != 0x90 /* NOTE ON */) continue;  // Ignore anything that isn't NOTE ON/OFF
--- a/game/controllers-midi.cc
+++ b/game/controllers-midi.cc
@@ -28,7 +28,7 @@ namespace input {
 				}
 			}
 		}
-		std::string getName(unsigned dev) const override {
+		std::string getName(int dev) const override {
 			PmDeviceInfo const* info = Pm_GetDeviceInfo(dev);
 			if (!info) throw std::logic_error("Invalid MIDI device requested in Midi::getName");
 			std::ostringstream name;
@@ -37,18 +37,18 @@ namespace input {
 		}
 		bool process(Event& event) override {
 			PmEvent ev;
-			for (auto it = m_streams.begin(); it != m_streams.end(); ++it) {
+			for (auto it = m_streams.begin(); it != m_streams.end();++it) {
 				if (Pm_Read(*it->second, &ev, 1) != 1) continue;
-				unsigned char evnt = ev.message & 0xF0;
-				unsigned char note = ev.message >> 8;
-				unsigned char vel  = ev.message >> 16;
+				unsigned char evnt = static_cast<unsigned char>(ev.message & 0xF0);
+				unsigned char note = static_cast<unsigned char>(ev.message >> 8);
+				unsigned char vel  = static_cast<unsigned char>(ev.message >> 16);
 				unsigned chan = (ev.message & 0x0F) + 1;  // It is conventional to use one-based indexing
-				if (evnt == 0x80 /* NOTE OFF */) { evnt = 0x90; vel = 0; }  // Translate NOTE OFF into NOTE ON with zero-velocity
+				if (evnt == 0x80 /* NOTE OFF */) { evnt = static_cast<unsigned char>(0x90); vel = 0; }  // Translate NOTE OFF into NOTE ON with zero-velocity
 				if (evnt != 0x90 /* NOTE ON */) continue;  // Ignore anything that isn't NOTE ON/OFF
 				std::clog << "controller-midi/info: MIDI NOTE ON/OFF event: ch=" << unsigned(chan) << " note=" << unsigned(note) << " vel=" << unsigned(vel) << std::endl;
-				event.source = SourceId(SourceType::MIDI, it->first, chan);
-				event.hw = note;
 				event.value = vel / 127.0;
+				event.source = SourceId(SourceType::MIDI, it->first, chan);
+				event.hw = static_cast<unsigned>(note);
 				return true;
 			}
 			return false;
--- a/game/controllers-midi.cc
+++ b/game/controllers-midi.cc
@@ -40,8 +40,8 @@ namespace input {
 			for (auto it = m_streams.begin(); it != m_streams.end();++it) {
 				if (Pm_Read(*it->second, &ev, 1) != 1) continue;
 				unsigned char evnt = static_cast<unsigned char>(ev.message & 0xF0);
-				unsigned char note = static_cast<unsigned char>(ev.message >> 8);
-				unsigned char vel  = static_cast<unsigned char>(ev.message >> 16);
+				unsigned char note = static_cast<unsigned char>(Pm_MessageData1(ev.message));
+				unsigned char vel  = static_cast<unsigned char>(Pm_MessageData2(ev.message));
 				unsigned chan = (ev.message & 0x0F) + 1;  // It is conventional to use one-based indexing
 				if (evnt == 0x80 /* NOTE OFF */) { evnt = static_cast<unsigned char>(0x90); vel = 0; }  // Translate NOTE OFF into NOTE ON with zero-velocity
 				if (evnt != 0x90 /* NOTE ON */) continue;  // Ignore anything that isn't NOTE ON/OFF
--- a/game/game.cc
+++ b/game/game.cc
@@ -68,10 +68,10 @@ void Game::drawLoading() {
 	const float x = 0.3f;
 	const float spacing = 0.01f;
 	const float sq_size = (2*x - (maxi-1)*spacing) / maxi;
-	for (int i = 0; i <= m_loadingProgress * maxi; ++i) {
+	for (float f = 0.0f; f <= m_loadingProgress * maxi; ++f) {
 		ColorTrans c(Color(0.2f, 0.7f, 0.7f, (m_loadingProgress + 1.0f)*0.5f));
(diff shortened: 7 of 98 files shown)


Lord-Kamina:
I see there are still warnings on the runners because it seems GCC displays warnings that clang doesn't. Help maybe @Baklap4 ?

Baklap4:
Just leaving it here for backtracing. On windows there's quite some progress on resolving warnings:

With the ced patch and a tweak in compile_options (`/w4` instead of `/wAll`) i'm down to just 334 warnings left on windows. However a couple of `/wAll` warnings we do need to look into i think:

- [ ] C4146 - Unary minus operator applied to unsigned type, result still unsigned (2)
- [ ] C4242, C4244, C4267, C4365, C4800, C5219 - (implicit) Conversion from to possible loss of data (789) << conversion stuff that still has to be addressed
- [ ] C4266 - No override available for virtual member; function is hidden (6) << not sure if intended
- [ ] C4296 - Expression is always true (2)
- [ ] C4305 - truncation from int to da::sample (68) << Conversion stuff that still has to be addressed
- [ ] C4355 - this used in base member initializer list (74) << Incorrect usage of `this`. Needs to be addressed
- [ ] C4514 - unreferenced inline function has been removed (8371) << compiler optimizations. No need to tackle this
- [ ] C4625, 4626, 5026,5027 - Defined as deleted (2595) << compiler optimizations. No need to tackle this
- [ ] C4820 - 'X' bytes padding added after data memember '' (1763) << compiler optimizations. Not sure if we need to tackle this
- [ ] C5045 - Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified (284) << compiler optimizations. Do we want spectre mitigation?

Lord-Kamina:
> Just leaving it here for backtracing. On windows there's quite some progress on resolving warnings:
> 
> 
> 
> With the ced patch and a tweak in compile_options (`/w4` instead of `/wAll`) i'm down to just 334 warnings left on windows. However a couple of `/wAll` warnings we do need to look into i think:
> 
> 
> 
> - [ ] C4146 - Unary minus operator applied to unsigned type, result still unsigned (2)
> 
> - [ ] C4242, C4244, C4267, C4365, C4800, C5219 - (implicit) Conversion from to possible loss of data (789) << conversion stuff that still has to be addressed
> 
> - [ ] C4266 - No override available for virtual member; function is hidden (6) << not sure if intended
> 
> - [ ] C4296 - Expression is always true (2)
> 
> - [ ] C4305 - truncation from int to da::sample (68) << Conversion stuff that still has to be addressed
> 
> - [ ] C4355 - this used in base member initializer list (74) << Incorrect usage of `this`. Needs to be addressed
> 
> - [ ] C4514 - unreferenced inline function has been removed (8371) << compiler optimizations. No need to tackle this
> 
> - [ ] C4625, 4626, 5026,5027 - Defined as deleted (2595) << compiler optimizations. No need to tackle this
> 
> - [ ] C4820 - 'X' bytes padding added after data memember '' (1763) << compiler optimizations. Not sure if we need to tackle this
> 
> - [ ] C5045 - Compiler will insert Spectre mitigation for memory load if /Qspectre switch specified (284) << compiler optimizations. Do we want spectre mitigation?

It'd be better if you could provide file/linenumbers.

Baklap4:
> It'd be better if you could provide file/linenumbers.

That'd be a lot of linenumbers. The warnings can be found within the Windows build on github actions. It was more an indication of what other warnings are there as of yet and if we want to fix them or not

Lord-Kamina:
Fair enough, I guess I can search the logs for those strings.

game/controllers-midi.cc
```diff
@@ -37,25 +37,25 @@ namespace input {
...
-				unsigned char evnt = ev.message & 0xF0;
-				unsigned char note = ev.message >> 8;
-				unsigned char vel  = ev.message >> 16;
-				unsigned chan = (ev.message & 0x0F) + 1;  // It is conventional to use one-based indexing
-				if (evnt == 0x80 /* NOTE OFF */) { evnt = 0x90; vel = 0; }  // Translate NOTE OFF into NOTE ON with zero-velocity
+			for (auto const& [key, value]: m_streams) { 
+				if (Pm_Read(value.get(), &ev, 1) != 1) continue;
+				unsigned char evnt = static_cast<unsigned char>(ev.message & 0xF0);
+				unsigned char note = static_cast<unsigned char>(ev.message >> 8);
+				unsigned char vel  = static_cast<unsigned char>(ev.message >> 16);
```
OznOg, line 44: what about using http://portmedia.sourceforge.net/portmidi/doxygen/group__grp__events__filters.html#g5e4ff2b1f4d8ae84ddeaa6465d5de5db macros?

game/controllers-keyboard.cc
```diff
@@ -44,14 +44,14 @@ namespace input {
 			// Convert SDL event into controller Event
 			event.source = SourceId(SourceType::KEYBOARD, 0);
 			event.hw = sdlEv.key.keysym.scancode;
-			if (sdlEv.type == SDL_KEYDOWN) event.value = 1.0;
+			if (sdlEv.type == SDL_KEYDOWN) event.val= 1.0;
```
Baklap4, line 47: spaces around the `=`

game/controllers.hh
```diff
@@ -61,11 +61,13 @@ namespace input {
 		HWButton hw; ///< Hardware button number (for internal use and debugging only)
 		Button button; ///< Mapped button id
 		NavButton nav; ///< Navigational button interpretation
-		std::optional<double> value; ///< false for button release, up to 1.0 for press (e.g. velocity value), or axis value (-1.0 .. 1.0)
+		std::optional<double> val; ///< false for button release, up to 1.0 for press (e.g. velocity value), or axis value (-1.0 .. 1.0)
+		double value() const { return val.value_or(0.0); }
+		uint32_t uValue() const { return static_cast<uint32_t>(value()); }
```
Baklap4, line 66: Personally i wouldn't use the fixed-width type. Since it doesn't solve the problem we're trying to tackle (conversion warnings).
As said on discord fixed-width types are a thing when you do embedded stuff and need to exactly know how many bytes a certain type is since sometimes you need to do micro optimizations like that for your specific hardware/compiler. 

For us that's not the case, so `int` should suffice here.
Also not going to repeat myself throughout this review, so please change it accordingly in other places

CMakeLists.txt
```diff
@@ -84,6 +82,10 @@ endif()
 add_subdirectory(data)
 
 add_subdirectory(game)
+
+target_compile_options(performous PRIVATE -Wall $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wextra> $<$<CXX_COMPILER_ID:gcc>:-fcx-limited-range> $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wconversion>)
```
Baklap4, line 86: As said on Discord, we shouldn't add flags like this since they can differ between compilers. Currently noted with the `NOT:COMPILER_ID:MSVC` for example. 
If possible make use of cmake compile features list: https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES

this will make sure the correct compile option is added for the correct compiler

game/3dobject.hh
```diff
@@ -17,18 +17,18 @@ class Object3d {
...
 		load(filepath, texturepath, scale);
 	}
 	/// load a new object file
-	void load(fs::path const& filepath, fs::path const& texturepath = fs::path(), float scale = 1.0);
+	void load(fs::path const& filepath, fs::path const& texturepath = fs::path(), float scale = 1.0f);
 	/// draws the object (with texture if given)
 	void draw();
 	/// draws the object with a transform
-	void draw(float x, float y, float z = 0, float s = 1.0);
+	void draw(float x, float y, float z = 0, float s = 1.0f);
```
Baklap4, line 33: float z is not a float. Use the suffix `f`

game/audio.hh
```diff
@@ -145,8 +146,8 @@ public:
 class Music {
 struct Track {
 	AudioBuffer audioBuffer;
-	float fadeLevel = 1.0f;
-	float pitchFactor = 0.0f;
+	double fadeLevel = 1.0;
+	double pitchFactor = 0.0f;
```
Baklap4, line 150: pitchfactor is defined as double yet we pass in a float.

game/configuration.cc
```diff
@@ -131,24 +131,27 @@ std::string const ConfigItem::getValue() const {
...
+		unsigned autoLanguageType = 1337;
+		unsigned val = LanguageToLanguageId(this->getEnumName()); // In the case of the language, val is the real value while m_value is the enum case for its cosmetic name.
 		std::string languageName = (val != autoLanguageType) ? this->getEnumName() : "Auto";
 		return languageName;
 	}
 	if (m_type == "int") {
-		int val = std::get<int>(m_value);
-		if (val >= 0 && val < int(m_enums.size())) return m_enums[val];
-		return numericFormat<int>(m_value, m_multiplier, m_step) + _(m_unit);
+		unsigned val = static_cast<unsigned>(std::get<int32_t>(m_value));
```
Baklap4, line 140: why not just get unsigned int 32?

game/dancegraph.cc
```diff
@@ -171,16 +172,16 @@ void DanceGraph::finalizeTrackChange() {
...
+	else if (gm == "dance-double") { m_pads = 4; std::copy(mapping8, mapping8+max_panels, m_arrow_map); }
+	else if (gm == "dance-couple") { m_pads = 4; std::copy(mapping8, mapping8+max_panels, m_arrow_map); }
+	else if (gm == "dance-solo") { m_pads = 4; std::copy(mapping6, mapping6+max_panels, m_arrow_map); }
+	else if (gm == "pump-single") { m_pads = 4; std::copy(mapping5, mapping5+max_panels, m_arrow_map); }
+	else if (gm == "pump-double") { m_pads = 4; std::copy(mapping10, mapping10+max_panels, m_arrow_map); }
+	else if (gm == "pump-couple") { m_pads = 4; std::copy(mapping10, mapping10+max_panels, m_arrow_map); }
+	else if (gm == "ez2-single") { m_pads = 4; std::copy(mapping5, mapping5+max_panels, m_arrow_map); }
+	else if (gm == "ez2-double") { m_pads = 4; std::copy(mapping10, mapping10+max_panels, m_arrow_map); }
+	else if (gm == "ez2-real") { m_pads = 4; std::copy(mapping7, mapping7+max_panels, m_arrow_map); }
+	else if (gm == "para-single") { m_pads = 4; std::copy(mapping5, mapping5+max_panels, m_arrow_map); }
```
Baklap4, line 184: Something funky going on with assigning m_pads value to always 4?

Baklap4 (dismissed):

game/controllers.hh
Lord-Kamina: It actually should solve some of them, if you are careful to (where possible) always mix unsigneds with bigger signed types, which is what I tried to do in a few places.

CMakeLists.txt
Lord-Kamina: I mean sure, but I just copy-pasted from a previous `add_compile_features` which is objectively worse. 👀😂

game/dancegraph.cc
Lord-Kamina: Rofl this might explain the bug @ooshlablu saw. It's evidently a botched regex (I changed it back and forth between float and unsigned like 10 times)

game/configuration.cc
Lord-Kamina: > why not just get unsigned int 32?

This was actually a bug, for a different reason.

Lord-Kamina:
@Baklap4 I've addressed most of your comments. The ones I haven't, I've specified why.

Lord-Kamina:
@Baklap4 this doesn't look kosher:
```c++
In function 'std::string {anonymous}::loadFile(const std::filesystem::__cxx11::path&)',
    inlined from 'Shader& Shader::compileFile(const std::filesystem::__cxx11::path&)' at /__w/performous/performous/performous/game/glshader.cc:105:41:
/__w/performous/performous/performous/game/glshader.cc:24:29: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
   24 |                 data.back() = '\0';
   ```

File: game/audio.cc [hunk 9]
```diff
@@ -267,12 +267,14 @@ struct Sample {
 			// No more data to play in this sample
 			return;
 		}
-		std::vector<float> mixbuf(end - begin);
-		if(!audioBuffer.read(mixbuf.data(), mixbuf.size(), m_pos, 1.0)) {
+		if (end <= begin) return;
```
OznOg, line 270: probably needs eof to be set @ true

game/audio.cc
```diff
@@ -291,24 +293,24 @@ struct Synth {
 	Synth(Notes const& notes, unsigned int sr) : m_notes(notes), srate(sr) {}
 	void operator()(float* begin, float* end, double position) {
 		static double phase = 0.0;
-		for (float *i = begin; i < end; ++i) *i *= 0.3; // Decrease music volume
-
-		std::vector<float> mixbuf(end - begin);
+		for (float *i = begin; i < end; ++i) *i *= 0.3f; // Decrease music volume
+		std::int64_t size = end - begin;
+		std::vector<float> mixbuf(static_cast<size_t>(size));
```
OznOg, line 298: I guess the same kind of check is necessary in order to prevent negative values to be cast in size_t

game/backgrounds.cc
```diff
@@ -61,7 +61,7 @@ void Backgrounds::reload_internal(fs::path const& parent) {
 				m_dirty = true;
 			}
 		}
-	} catch (std::exception const& e) {
+	} catch (std::exception const&) {
 		std::cout << "Error accessing " << parent << std::endl;
```
OznOg, line 65: maybe deserve to be displayed in error as well (instead of droping it)

game/controllers-joystick.cc
```diff
@@ -20,25 +20,29 @@ namespace input {
...
 			}
 		}
-		std::string getName(unsigned device) const override {
+		std::string getName(int device) const override {
 			return SDL_JoystickNameForIndex(device);
 		}
 		bool process(Event& event, SDL_Event const& sdlEv) override {
 			if (sdlEv.type == SDL_JOYBUTTONDOWN || sdlEv.type == SDL_JOYBUTTONUP) {
 				event.hw = sdlEv.jbutton.button;
-				event.value = (sdlEv.type == SDL_JOYBUTTONDOWN ? 1.0 : 0.0);
```
OznOg, line 29: careful, this code is not equivalent, the event.value is not set when type != SDL_JOYBUTTONDOWN now

game/controllers-keyboard.cc
```diff
@@ -44,14 +44,14 @@ namespace input {
 			// Convert SDL event into controller Event
 			event.source = SourceId(SourceType::KEYBOARD, 0);
 			event.hw = sdlEv.key.keysym.scancode;
-			event.value = (sdlEv.type == SDL_KEYDOWN ? 1.0 : 0.0);
+			if (sdlEv.type == SDL_KEYDOWN) event.val = 1.0;
```
OznOg, line 47: same, may be undefined now

game/unicode.hh
```diff
@@ -16,8 +17,8 @@ struct UnicodeUtil {
 	static std::string getCharset(std::string const& str);
 	static void convertToUTF8 (std::stringstream &_stream, std::string _filename);
 	static std::string convertToUTF8 (std::string const& str);
-	static std::string toLower (std::string const& str, size_t length = 0);
-	static std::string toUpper (std::string const& str, size_t length = 0);
+	static std::string toLower (std::string const& str, int length = 0);
+	static std::string toUpper (std::string const& str, int length = 0);
```
OznOg, line 21: len has conceptualy no reason to be negative, thus this would deserve to remain unsigend (moreover, if signed, the function is now buggy because we don't check for negative values)

game/unicode.cc
```diff
@@ -67,22 +67,24 @@ std::string UnicodeUtil::convertToUTF8 (std::string const& str) {
...
+	int sizeI = static_cast<int>(str.size());
+	if (length == 0 || length >= sizeI) length = sizeI;
 		icu::UnicodeString tmp = icu::UnicodeString::fromUTF8(ss.str());
 		tmp = tmp.tempSubString(0, length).toLower() + tmp.tempSubString(length, tmp.length() - 1);
 		tmp.toUTF8String(ret);
 	return ret;
 }
 
-std::string UnicodeUtil::toUpper (std::string const& str, size_t length) {
+std::string UnicodeUtil::toUpper (std::string const& str, int length) {
```
OznOg, line 82: IMHO, sizes should not be signed

game/textinput.hh
```diff
@@ -9,19 +9,19 @@ struct TextInput {
 	/// appends unicode symbol
 	TextInput& operator+=(unsigned int ucs) { //don't know if this is still relevant since SDL2 handles unicode
 		if (ucs < 0x80) {
-			text += ucs;
+			text += static_cast<std::basic_string<char>::value_type>(ucs);
```
OznOg, line 12: looks strange, I could understand std::string::value_type or char directly, but as the template param IS the value_type, seem to be complicated here...

game/songparser.hh
```diff
@@ -16,6 +17,8 @@ namespace SongParserUtil {
 	void assign(unsigned& var, std::string const& str);
 	/// Parse a double from string and assign it to a variable
 	void assign(double& var, std::string str);
```
OznOg, line 19: is the double still useful?

game/songparser.cc
```diff
@@ -17,11 +19,19 @@ namespace SongParserUtil {
...
+			var = stou(str);
 		} catch (...) {
 			throw std::runtime_error ("\"" + str + "\" is not valid unsigned integer value");
 		}
 	}
+	void assign (float& var, std::string str) {
+		std::replace (str.begin(), str.end(), ',', '.');  // Fix decimal separators
+		try {
+			var = std::stof (str);
+		} catch (...) {
```
OznOg, line 31: please catch the exception and use it.

game/songparser-sm.cc
```diff
@@ -234,8 +234,8 @@ Notes SongParser::smParseNotes(std::string line) {
 			forceMeasure = true;
 			line = line.substr(0, line.size()-1);
 		}
-		for(std::size_t i = 0; i < line.size(); i++) {
-			char notetype = line[i];
+		for(unsigned i = 0; i < static_cast<unsigned>(line.size()); i++) {
```
OznOg, line 237: why casting things here if you end up by casting i to float bellow?

game/songparser-sm.cc
```diff
@@ -247,12 +247,12 @@ Notes SongParser::smParseNotes(std::string line) {
...
 			else if(notetype >= 'A' && notetype <= 'Z') note.type = Note::Type::TAP;
 			else throw std::runtime_error(std::string("Invalid note command: ") + notetype);
-			note.note = i;
+			note.note = static_cast<float>(i);
 			chord[i] = note;
 		}
 		chords.push_back(chord);
 	}
-	m_tsEnd = std::max(m_tsEnd, measure * 16);
+	m_tsEnd = std::max(m_tsEnd, static_cast<unsigned>(std::round(measure * 16.0)));
```
OznOg, line 255: not sure round is useful as casting a float to an int is already some kind of round (ok, actually it truncates...)

game/songitems.cc
```diff
@@ -32,9 +32,10 @@ void SongItems::save(xmlpp::Element* songs) {
 	}
 }
 
-int SongItems::addSongItem(std::string const& artist, std::string const& title, int id) {
+unsigned SongItems::addSongItem(std::string const& artist, std::string const& title, SongId _id) {
 	SongItem si;
-	if (id==-1) id = assign_id_internal();
+	unsigned id;
+	id = _id ? _id.value() : assign_id_internal();
```
OznOg, line 38: id = _id.value_or(assign_id_internal());

game/songitems.cc
```diff
@@ -50,14 +51,10 @@ int SongItems::addSongItem(std::string const& artist, std::string const& title,
...
-	if (id == -1)
-	{
-		id = addSongItem(song->artist, song->title);
-	}
-
+	SongId id = lookup(song);
+	if (!id) id = addSongItem(song->artist, song->title);
 	SongItem si;
-	si.id = id;
+	si.id = id.value();
```
OznOg, line 57: si.id = lookup(song).value_or(addSongItem(song->artist, song->title));

game/songitems.cc
```diff
@@ -71,32 +68,27 @@ void SongItems::addSong(std::shared_ptr<Song> song) {
 	m_songs.insert(si);
 }
 
-int SongItems::lookup(std::shared_ptr<Song> song) const {
-	for (auto const& s: m_songs) {
-		if (song->collateByArtistOnly == s.artist && song->collateByTitleOnly == s.title) return s.id;
-	}
-	return -1;
+SongId SongItems::lookup(Song const& song) const {
+	auto const& si = std::find_if(m_songs.begin(), m_songs.end(), [song](SongItem const& si) {
```
OznOg, line 72: use std::any_of

game/songitems.cc
```diff
@@ -71,32 +68,27 @@ void SongItems::addSong(std::shared_ptr<Song> song) {
...
 	else if (!it->song) return it->artist + " - " + it->title;
 	else return it->song->artist + " - " + it->song->title;
 }
 
-int SongItems::assign_id_internal() const {
+unsigned SongItems::assign_id_internal() const {
 	// use the last one with highest id
 	auto it = m_songs.rbegin();
 	if (it != m_songs.rend()) return it->id+1;
-	else return 1; // empty set
```
OznOg, line 101: you change 1 to 0 on purpose?

game/midifile.cc
```diff
@@ -34,40 +36,40 @@ class MidiStream {
...
+		void consume(std::streamoff bytes);
 	};
 
 private:
 
 	std::stringstream f;
-	uint16_t read_uint16() { return f.get() << 8 | f.get(); }
-	uint32_t read_uint32() { return f.get() << 24 | f.get() << 16 | f.get() << 8 | f.get(); }
+	std::uint16_t read_uint16() { return static_cast<std::uint16_t>(f.get() << 8 | f.get()); }
+	std::uint32_t read_uint32() { return static_cast<std::uint32_t>(f.get() << 24 | f.get() << 16 | f.get() << 8 | f.get()); }
```
OznOg, line 72: I'd suggest 
uint32_t v = f.get(); and doing bitshift on 32bits. keeping the cast after bit shifts is making it more strange.

game/platform/execname.mac.inc
```diff
@@ -1,8 +1,9 @@
 #include <mach-o/dyld.h>
+#include <cstdint>
 
 fs::path execname() {
 	char buf[1024];
-	uint32_t size = sizeof(buf);
+	std::uint32_t size = sizeof(buf);
```
OznOg, line 6: maybe time to use std::array<1024>

game/player.cc
```diff
@@ -79,13 +77,10 @@ void Player::calcRowRank() {
...
 		}
 		m_lineScore = 0;
 	}
 }
 
-PlayerItem::PlayerItem(PlayerId id) 
-: id(id)
-{
-}
+PlayerItem::PlayerItem(PlayerId _id) { if (_id) id = _id.value(); }
```
OznOg, line 86:
```suggestion
PlayerItem::PlayerItem(PlayerId _id) : id{_id} {}
```

game/player.hh
```diff
@@ -89,6 +88,7 @@ struct PlayerItem {
 	 Provides ordering and ensures id is unique.*/
 	bool operator< (PlayerItem const& pi) const
```
OznOg, line 89: not sure this is correct now: invalid id will not be stable in their ordering, this may mess up the user.

game/singleton.hh
```diff
@@ -4,7 +4,7 @@
 template <class T> class Singleton {
   protected:
 	/// pointer to object
-	static T* ms_Singleton;
+	inline static T* ms_Singleton{};
```
OznOg, line 7: useless

game/controllers-joystick.cc
Lord-Kamina: I want to look well into this because there is a joystick bug I haven't been able to get yet.

However, I don't think I'm ever actually checking the optional value directly, but rather using a function that returns `.value_or(0.0)`

game/unicode.hh
Lord-Kamina: I changed that because the underlying ICU function takes int.

game/unicode.cc
Lord-Kamina: I agree butnC libraries do not.

game/textinput.hh
Lord-Kamina: I just took what the compiler suggested there. I want to revise this code in a later PR because I'm not completely sure most of it is actually necessary.

game/songitems.cc
Lord-Kamina: I did, I figured breakage potential for existing users is very small and starting at 1 is weird anyway.

game/controllers-midi.cc
Lord-Kamina: I'm not really sure which ones, tbh.

game/game.cc
```diff
@@ -10,16 +10,14 @@
 #include <stdexcept>
 #include <cstdlib>
 
-template<> Game* Singleton<Game>::ms_Singleton = nullptr;
```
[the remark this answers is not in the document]
Lord-Kamina: I think it's not needed if we use the notation below that you said was useless.

game/hiscore.cc
```diff
@@ -23,28 +24,29 @@ bool Hiscore::reachedHiscore(unsigned score, unsigned songid, unsigned level, st
...
 	HiscoreVector hv;
 	for (auto const& h: m_hiscore) {
-		if (playerid != unsigned(-1) && playerid != h.playerid) continue;
-		if (songid != unsigned(-1) && songid != h.songid) continue;
+		if (playerid && playerid.value() != h.playerid) continue;
+		if (songid && songid.value() != h.songid) continue;
 		if (currentLevel() != h.level) continue;
 		if (!track.empty() && track != h.track) continue;
-		if (--max == 0) break;
+		if (max && --max.value() == 0) break;
```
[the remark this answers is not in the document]
Lord-Kamina: :( I love that syntax, it's so Swift.

game/hiscore.hh
```diff
@@ -44,17 +48,17 @@ public:
...
-	void addHiscore(unsigned score, unsigned playerid, unsigned songid, unsigned level, std::string const& track);
+	void addHiscore(unsigned score, PlayerId playerid, SongId songid, unsigned short level, std::string const& track);
 
 	typedef std::vector<HiscoreItem> HiscoreVector;
 
 	/// This queries the database for a sorted vector of highscores. The defaults mean to query everything.
 	/// @param max limits the number of elements returned.
-	HiscoreVector queryHiscore(unsigned max, unsigned playerid, unsigned songid, std::string const& track) const;
-	bool hasHiscore(unsigned songid) const;
+	HiscoreVector queryHiscore(PlayerId playerid, SongId songid, std::string const& track, std::optional<unsigned> max = std::optional<unsigned>()) const;
```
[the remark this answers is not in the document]
Lord-Kamina: Yeah, I had this really dumb moment I began doing this because I couldn't understand why `nullopt`wasn't working... then I realized I was typing `nullopt_t`. Clearly missed this one when I went around fixing it.

game/menu.hh
```diff
@@ -93,14 +94,14 @@ public:
 
 	bool empty() const { return (menu_stack.empty() || (menu_stack.size() == 1 && menu_stack.back()->empty())); }
 	bool isOpen() const { return m_open; }
-	size_t getSubmenuLevel() const { return menu_stack.size() - 1; }
```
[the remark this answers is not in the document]
Lord-Kamina: If I then try to assign these size_t to an unsigned value I get a loss-of-precision warning, so I'd have to use size_t everywhere.

game/unicode.cc
OznOg: well ssize_t exists, but size_t is unsigned by default, not sure I undrstnd your reply

Lord-Kamina: Same as above, I'm using int because ICU wants int. I we didn't use int here, we'd have to cast it later anyway.

game/player.cc
Lord-Kamina: This one won't work because PlayerItem actually wants an unsigned and not an optional.

game/player.hh
Lord-Kamina: I'm not sure this is relevant. In practice it means invalid IDs would be bunched together (which would already happen with the current code) and if we're trying to sort anything more than at most one invalid ID, we're doing something wrong IMO.

Lord-Kamina: So yeah, turns out you were right and this was the source of the bug. I reverted this part.

game/controllers-midi.cc
OznOg: done

game/singleton.hh
OznOg: I added some comments in the code: this singleton is completely broken, and as singleton is an antipattern, better remove it than doing half baked fixes

OznOg (approved):

game/controllers.hh
OznOg: the code was removed;no more int32_t on last version

game/animvalue.hh
```diff
@@ -69,8 +71,8 @@ class AnimAcceleration {
 			m_velocity = 0.0;
 			return m_position = m_target;
 		}
-		std::size_t rounds = 1.0 + 1000.0 * duration; // 1 ms or shorter timesteps
-		double t = duration / rounds;
+		std::size_t rounds = static_cast<size_t>(1.0 + 1000.0 * duration); // 1 ms or shorter timesteps
+		double t = duration / static_cast<double>(rounds);
```
nieknooijens, line 75: Wait... Double? I thought we were converting to float?

game/audio.cc
```diff
@@ -407,12 +413,12 @@ struct Output {
 };
 
 
-Device::Device(unsigned int in, unsigned int out, double rate, unsigned int dev):
+Device::Device(int in, int out, double rate, PaDeviceIndex dev):
```
nieknooijens, line 416: I know pipewire supports pulseaudio protocol, but having specific mentions of pulseaudio in the code does kind of get on my nerves...

game/configuration.hh
```diff
@@ -47,25 +51,25 @@ class ConfigItem {
...
 	Value m_factoryDefaultValue; ///< The value from config schema
 	Value m_defaultValue; ///< The value from config schema or system config
 	std::string m_oldValue; ///< A previous value, as output by getValue().
 	std::vector<std::string> m_enums; ///< Enum value titles
-	std::variant<int, double> m_step, m_min, m_max;
-	std::variant<int, double> m_multiplier;
+	std::variant<int, unsigned short, float> m_step, m_min, m_max, m_multiplier;
 	std::string m_unit;
-	int m_sel;
+	unsigned short m_sel;
```
nieknooijens, line 72: Everywhere else you explicitly use std:int64_t or other types with a guaranteed bitlength, yet here you use short. Looks a bit inconsistent to me.

game/guitargraph.hh
```diff
@@ -121,14 +123,14 @@ class GuitarGraph: public InstrumentGraph {
 	void drawNeckStuff(double time);  ///< Anything in neck coordinates
 	void drawNotes(double time);  ///< Frets etc.
 	void drawBar(double time, float h);
-	void drawNote(unsigned fret, Color, float tBeg, float tEnd, float whammy = 0, bool tappable = false, bool hit = false, double hitAnim = 0.0, double releaseTime = 0.0);
-	void drawDrumfill(float tBeg, float tEnd);
+	void drawNote(unsigned fret, Color, double tBeg, double tEnd, float whammy = 0.0f, bool tappable = false, bool hit = false, double hitAnim = 0.0, double releaseTime = 0.0);
+	void drawDrumfill(double tBeg, double tEnd);
```
nieknooijens, line 127: Again, don't understand why this became double

game/midifile.cc
```diff
@@ -180,8 +182,8 @@ MidiFileParser::Track MidiFileParser::read_track(MidiStream& stream) {
 					std::string sect_name = data.substr(sect_pfx.length(), data.length()-sect_pfx.length()-1);
 					if (sect_name != "big_rock_ending") {
 						bool space = true;
-						for (auto& ch: sect_name) {
```
nieknooijens, line 183: Good one. I hate the auto keyword.better to be explicit

game/notes.cc
```diff
@@ -43,35 +43,35 @@ double Note::scoreMultiplier() const {
 }
 
 double thresholdForFullScore() {
-	switch(GameDifficulty(config["game/difficulty"].i())){
+	switch(GameDifficulty(config["game/difficulty"].ui())){
 		case GameDifficulty::PERFECT:
```
nieknooijens, line 47: Is the fall through here desired? Maybe add //fallthrough or //falls through

game/animvalue.hh
Lord-Kamina: The idea is to use double for frequencies and also stuff involving time (such as animations). I've also in some places preserved the double during calculation and then cast to float to actually use the results.

game/audio.cc
Lord-Kamina: What?

game/configuration.hh
Lord-Kamina: If I do it's a merging mistake. I wanted to use exclusively fixed-length types but @Baklap4 wont't have it, so I changed to all non-fixed-width again, except for a few exceptions.

game/guitargraph.hh
Lord-Kamina: It's times. Float is not good enough for time.

game/audio.cc
OznOg: I'm currently reworking the libportaudio part, and this modification is just about making clear the dev was the PA index not a random int (dependency on PA was hidden behind a bad cast). So this may look like an issue, but that's just because it reveals that the former code was not good anyway.

nieknooijens (dismissed):

CMakeLists.txt
OznOg: ok, let's move on an fix this in a separate MR

game/notes.cc
OznOg: can't see any fall through here, what are you referring to?

ooshlablu (approved):
Can't comment on the code, but I have tested the latest build with singing and guitar/bass and everything appears to work as expected.

PR #772, opened 2022-05-27
Fix joystick test

Lord-Kamina:
### What does this PR do?

Fix joystick test mode.

### Motivation

Debugging whatever I apparently broke in #769 

### More

### Additional Notes

~Haven't actually tested this properly, made this branch remotely via VNC but it seems to be doing as intended.~
Just got home and confirmed it's working as intended.


Baklap4 (commented):
Code looks good to me, but i cannot test this as i don't own a controller.
Anyone with a controller who can double check this? @ooshlablu @OznOg 

If it works feel free to aprove + merge 👍

game/main.cc
```diff
@@ -209,6 +209,8 @@ void jstestLoop() {
 		config["graphic/window_width"].i() = 640;
 		config["graphic/window_height"].i() = 360;
 		Window window;
+		input::Controllers controllers;
+		controllers.enableEvents(true);
```
Baklap4, line 213: As a side note i don't like this method naming/argument combi. But that's another issue on itself 😂 
`enableEvents` is already a certain action which doesn't need a bool input. A better name if you want to keep continueing the bool argument: `toggleEvents(true|false)`. Or have 2 methods: `enableEvents()` and `disableEvents()`

Lord-Kamina: Yeah we can change that later. Incidentally, I also think we should a) provide an option to configure the dead-zone for joystick axes and b) allow overriding the default in controllers.xml.

ooshlablu: Yeah, in looking into the issues in 769, I'm realizing that the X-plorer is pretty mis-configured. Currently the whammy bar is set to "god mode", but that should be triggered by tilting the guitar neck up. Since the gyro for that is constantly cranking out data, it would trigger no matter what. Having some kind of threshold setting would allow it to be configured properly.

ooshlablu (approved):
LGTM

PR #780, opened 2022-06-17
Introduce optional testing with GoogleTest

twollgam:
### What does this PR do?

This PR introduces automatic testing of functions and classes with the GoogleTest framework. The tests are only compiled if cmake detects the framework.

The PR includes also 2 workflows for ubuntu. They will show how to build and run tests.

### Closes Issue(s)

None.

### Motivation

Tests will help developers changing and refactoring existing code. For example I would like to add a new game mode where guitar chords can be played. To reuse the existing fft stuff I would create tests for the current implementation. Later the tests will show if eventhing works as expected even if there are changes or refactorings made.

### More

- [ ] Add/update documentation if the PR will be accepted and merged


.github/workflows/build_ubuntu_on_push.yml
```diff
@@ -0,0 +1,93 @@
...
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Prepare Submodules
+      run : git submodule update --init --recursive
+
+    - name: Install libraries
+      run: |
+        #sudo apt-get -y install software-properties-common
```
ooshlablu, line 49: I don't really understand why you are creating a build we already have. Would it make more sense to pull down one of the existing Ubuntu builds? All of the deps installed here are already installed in the Ubuntu containers so we don't waste time re-installing everything. If this does need to be built separately, then this should probably be run inside one of the Containers.

testing/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.15)
+
+file(GLOB SOURCE_FILES "*.cc")
```
OznOg, line 3: please don't use glob, this is known as buggy. prefer list files

testing/testhost.cc
```diff
@@ -0,0 +1,6 @@
+#include <gtest/gtest.h>
+
```
OznOg, line 2: If you link iwth gtest::ain, this should be useless

OznOg:
sadly I am rewriting configuration.cc, I guess I'll have some hard time for rebasing.

testing/CMakeLists.txt
twollgam: I copied it from game folder. Should we change it there too?

testing/testhost.cc
twollgam: Removed testhost.cc

testing/CMakeLists.txt
OznOg: well actually, I would agree with it, but as there are quite a lot of file in game/*c, I'm not sure everybody will agree... That said, in your case, you should have quite a few files, so should be ok.

OznOg:
> sadly I am rewriting configuration.cc, I guess I'll have some hard time for rebasing.

Just to stress that, even if my rebase will be painful, I believe that is a good idea to initiate unit tests in our code base, good idea.

testing/CMakeLists.txt
twollgam: Done

twollgam:
@OznOg currently I let compile files from game folder needed for the test host again cause I cannot link against the performous binary. At work we make a static library that contains all but the main function. So one can link both the performous binary and the test host against it. I would like to know what you think about it.

OznOg:
> @OznOg currently I let compile files from game folder needed for the test host again cause I cannot link against the performous binary. At work we make a static library that contains all but the main function. So one can link both the performous binary and the test host against it. I would like to know what you think about it.

The main question is "what are you aiming to test". You rework seems to aim isolating ConfigItem from the main config stuff, so I would suggest creating a static lib from configItem.cc and link it in the main game on one side and with the test on another. 

You may also test serialization/deserialization, but I think a little more work is needed: the config should not know about the format of the serialized file , so read/write Config should be put aside, and, rely on visit at some point; that may probably be done later as this MR is already adding quite a lot of things.

twollgam:
I decided not to make the split into executable and static lib now. Will be much easier if the pr is merged. Otherwise rebasing will be painful to not solvable.

.github/workflows/build_ubuntu_on_push.yml
twollgam: @ooshlablu can you add a workflow that installes google test so that this one willl run the tests? The demo workflows aren't run here, cause they are startet if commits are pushed. I will remove them afterwards.

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+
+set(SOURCE_FILES 
+    "configitemtest.cc"
+    "fixednotegraphscalertest.cc"
+    "notegraphscalerfactorytest.cc"
+    "utiltest.cc"
+)
+set(HEADER_FILES
+)
+set(GAME_SOURCES
```
OznOg, line 11: here you should probaly not compile again thos sources but create a static lib with them pulling the correct dependencies (boost::local and fmt)

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+cmake_minimum_required(VERSION 3.15)
+
+set(SOURCE_FILES 
+    "configitemtest.cc"
+    "fixednotegraphscalertest.cc"
+    "notegraphscalerfactorytest.cc"
+    "utiltest.cc"
+)
+set(HEADER_FILES
```
OznOg, line 9: useless

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
+cmake_minimum_required(VERSION 3.15)
+
+set(SOURCE_FILES
```
OznOg, line 3: don't use variable just put files in the add_executable

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    "../game/notes.cc"
+    "../game/notegraphscalerfactory.cc"
+)
+
+set(SOURCES ${SOURCE_FILES} ${HEADER_FILES} ${GAME_SOURCES})
+
+add_executable(performous_test ${SOURCES})
+
+find_package(Boost 1.55 REQUIRED COMPONENTS program_options iostreams system locale)
+target_include_directories(performous_test PRIVATE ${Boost_INCLUDE_DIRS})
```
OznOg, line 25: shoudl not be necessary

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+target_link_libraries(performous_test PRIVATE ${Boost_LIBRARIES})
+
+find_package(fmt REQUIRED CONFIG)
+target_link_libraries(performous_test PRIVATE fmt::fmt)
+
+target_include_directories(performous_test PRIVATE ${GTEST_INCLUDE_DIRS})
```
OznOg, line 31:
```suggestion

```
same

testing/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+target_link_libraries(performous_test PRIVATE GTest::GTest GTest::Main)
+
+target_include_directories(performous_test PRIVATE "..")
```
OznOg, line 34: same should be pulled by using a static lib

testing/configitemtest.cc
```diff
@@ -0,0 +1,285 @@
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include "game/configitem.hh"
```
OznOg, line 4: this should appear first, as this would also test the header is self contained (does not need extra headers)

testing/configitemtest.cc
```diff
@@ -0,0 +1,285 @@
...
+
+using ::testing::ElementsAre;
+using ::testing::IsEmpty;
+
+namespace {
```
OznOg, line 9: not sure this namespace is really useful

twollgam: It was suggested by google. Don't know when it is a must.

testing/CMakeLists.txt
twollgam: Will remove it.

testing/CMakeLists.txt
twollgam: I would like to do that in a following PR just to get a start for the testing stuff.

testing/CMakeLists.txt
twollgam: Why? Until there is no lib it is easier to divide the sources by variables. Or do I miss something?

testing/CMakeLists.txt
twollgam: :+1:

testing/configitemtest.cc
twollgam: I changed include and sequence.

testing/CMakeLists.txt
twollgam: Later ... (Another PR)

.github/workflows/build_ubuntu_on_push.yml
twollgam: I have done that already.

ooshlablu (dismissed):
Workflow and cmake changes look good. Tested with vocals and guitar on Ubuntu 18.04 and 20.04, everything works as expected

Lord-Kamina (approved):
LGTM

PR #782, opened 2022-06-21
performous-app-build.sh: Add option to create debuggable xcode project file

Lord-Kamina:
### What does this PR do?

Adds a couple modes to the macOS bundling script that allow it to create an xcode project that can then be used to run  performous under the debugger.

It has some caveats to be aware of, however.

1) If you have both Homebrew AND Macports installed, it will be a problem when running the "install" target in xcode (because that situation presents the user with an interactive prompt).

2) It's apparently not possible to edit the meta-targets CMake creates, so I couldn't add scheme options to those. In practice, this is mostly just a bit of a bother and means to make it work, you need to follow the following steps:

    a) Create the xcode project (passing the `-x` or `--xcode-project` options to this script).
    b) Open the Xcode project.
    c) Select the install scheme and build under `Product -> Build For -> Runnng.`
    d) Select the "performous" scheme and use `Product -> Run.`

### More

### Additional Notes

Originally, this script was meant mostly to be a useful timesaver, but with time it's become more and more complex.
I've been thinking of giving it a slight rewrite, maybe using something such as https://github.com/docopt/docopt. Any thoughts here?


Lord-Kamina:
Good thing nobody reviewed this yet because some code had slipped throgh the cracks.  😅

ooshlablu (approved):

Issue #783, opened 2022-07-01
The app doesn't exit while songs are being loaded


Tronic:
### Performous & operating system versions

Git master

### What is the problem?

Quiting the application while song loading is in progress closes the window but keeps playing music. The application only exits once the song loading is complete, which can take considerable time.

### Steps to reproduce

1. Add a large collection of songs
2. Ctrl-Q, close the window or hit Quit on menu
3. The window closes but menu music or other sounds keep playing, the process won't exit unless kill -9'd, app stays visible in MacOS task switcher
4. After a while the app closes

### Additional context

The song loader thread should be interrupted immediately (I believe there is a quit flag for this and I think it has worked before but for some reason this is now broken).

ooshlablu:
I have noticed this issue as well, and it has been mentioned in Discord a few times by others.

OznOg:
fix in #784

PR #784, opened 2022-07-01
Allow early exit when scanning songs is really long.

OznOg:
fix for bug #783


OznOg:
I actually tested adding /proc in the songs directory

PR #785, opened 2022-07-07
Speed up the sorting of songs by score

twollgam:
### What does this PR do?

A bugfix for SongItems::lookup slows down the sorting of songs by score. This PR speed up the sorting.

### Closes Issue(s)

None

### Motivation

Better user experience

PR #786, opened 2022-07-08
Fixes to UnicodeUtil, and a small refactor to the database.

Lord-Kamina:
### What does this PR do?

First, sorry because it does two things; the changes in Database are because this PR was born from the same issue #785 is fixing.

Anyway, there's a small refactor, it fixes a bug introduced with #769 that produced duplicate song entries in the database and, the most important part, it includes a pretty big rewrite to `UnicodeUtil`:

- We're no longer using std::stringstream.
- Try to optimize the conversion functions a bit by adding early returns.
- Combined the case convertion functions with convertToUTF8, so we don't need to instantiate UnicodeString twice.
- Added a case-insensitive compare function.
- Fixed some other things like using some C types instead of the c++ interface.

### More


game/hiscore.cc
```diff
@@ -42,15 +42,16 @@ Hiscore::HiscoreVector Hiscore::queryHiscore(std::optional<PlayerId> playerid, s
 	return hv;
 }
 
-bool Hiscore::hasHiscore(const SongId &songid) const {
-	return std::any_of(m_hiscore.begin(),m_hiscore.end(), [&songid, level = currentLevel()](auto const& h) { return songid == h.songid && level == h.level; });
+bool Hiscore::hasHiscore(const SongId songid) const {
+	return std::any_of(m_hiscore.begin(),m_hiscore.end(), [songid, level = currentLevel()](auto const& h) {
+		return songid == h.songid && level == h.level;
+	});
 }
```
twollgam, line 49: Thanks for the line breaks 👍

game/unicode.cc
```diff
@@ -37,62 +39,123 @@ std::string UnicodeUtil::getCharset (std::string const& str) {
...
+	else {
+		if (!ret.empty()) {
+			std::clog << "unicode/error: tried to convert text in an unknown encoding: " << charset << std::endl;
 		}
 	}
+	return ret;
+}
+
+bool UnicodeUtil::caseEqual (std::string const& lhs, std::string const& rhs, bool assumeUTF8) {
+	if (lhs == rhs) return true; // Early return
```
twollgam, line 88: 👍

game/unicode.cc
```diff
@@ -37,62 +39,123 @@ std::string UnicodeUtil::getCharset (std::string const& str) {
...
+	
+	std::string a = lhs;
+	std::string b = rhs;
```
twollgam, line 91: I think using string_view would be good and possible cause only the start of a string will be erased. So the terminating zero is always valid.

game/unicode.cc
```diff
@@ -37,62 +39,123 @@ std::string UnicodeUtil::getCharset (std::string const& str) {
...
+	std::string aCharset;
+	std::string bCharset;
```
twollgam, line 93: *Charset is used as flag here. So we should use bool and a const std::string for "UTF-8".

Lord-Kamina: No, because there are other valid values for the charset (essentially all old-timey code-pages, as well as Shift-JIS and such) which, if obtained, are then passed to the `UnicodeString` constructor.

game/unicode.cc
Lord-Kamina: The thing with string_view is we can't guarantee it will be null-terminated, which is kind of a big deal because ICU consumes C-strings. If we decided to do it anyway and explicitly cast the string_views to C strings, it could work for sure, but it's got potential to introduce bugs in the future.

game/unicode.cc
```diff
@@ -1,32 +1,34 @@
 #include "unicode.hh"
 
 #include "configuration.hh"
+#include "game.hh"
+
 #include <regex>
 #include <sstream>
```
OznOg, line 7: useless  now

game/unicode.cc
```diff
@@ -1,32 +1,34 @@
...
 #include <stdexcept>
 #include <unicode/unistr.h>
 #include <unicode/ustream.h>
+#include <unicode/ubidi.h>
 #include "compact_enc_det/compact_enc_det.h"
 
-UErrorCode UnicodeUtil::m_staticIcuError = U_ZERO_ERROR;
+icu::ErrorCode UnicodeUtil::m_staticIcuError = icu::ErrorCode();
 icu::RuleBasedCollator UnicodeUtil::m_dummyCollator (icu::UnicodeString (""), icu::Collator::PRIMARY, m_staticIcuError);
 icu::RuleBasedCollator UnicodeUtil::m_sortCollator  (nullptr, icu::Collator::SECONDARY, m_staticIcuError);
```
OznOg, line 16: is that still usefull?

game/unicode.cc
```diff
@@ -37,62 +39,123 @@ std::string UnicodeUtil::getCharset (std::string const& str) {
 	return MimeEncodingName(encoding);
 }
 
-void UnicodeUtil::convertToUTF8 (std::stringstream &_stream, std::string _filename) {
-	std::string data = _stream.str();
-	std::string charset = UnicodeUtil::getCharset(data);	
+std::string UnicodeUtil::convertToUTF8 (std::string const& str, std::string _filename, CaseMapping toCase, bool assumeUTF8) {
```
OznOg, line 42:
```suggestion
std::string UnicodeUtil::convertToUTF8 (std::string const& str, const std::string &_filename, CaseMapping toCase, bool assumeUTF8) {
```

OznOg: actually filename seems really useless here... (ok code was already like this)

game/hiscore.hh
```diff
@@ -45,7 +45,7 @@ public:
 	/// @param max limits the number of elements returned.
 	unsigned getHiscore(unsigned songid) const;
 	HiscoreVector queryHiscore(std::optional<PlayerId> playerid, std::optional<SongId> songid, std::string const& track, std::optional<unsigned> max = std::nullopt) const;
-	bool hasHiscore(const SongId &songid) const;
+	bool hasHiscore(const SongId songid) const;
```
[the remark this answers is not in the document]
Lord-Kamina: Because it's just an `unsigned int`, I mean sure we can use a ref but... why?

game/songitems.cc
```diff
@@ -50,7 +50,8 @@ SongId SongItems::addSongItem(std::string const& artist, std::string const& titl
 
 void SongItems::addSong(std::shared_ptr<Song> song) {
 	SongItem si;
-	si.id = lookup(song).value_or(addSongItem(song->artist, song->title));;
+	std::optional<unsigned> val = lookup(song);
+	si.id = val ? val.value() : (addSongItem(song->artist, song->title));;
```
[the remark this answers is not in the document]
Lord-Kamina: No. I don't understand why (seems like a C++ library bug) but that caused a bug. Apparently the "or" value was being evaluated regardless of whether there was a value or not, and that led to songs being duplicated in the database.

game/unicode.cc
```diff
@@ -37,62 +39,123 @@ std::string UnicodeUtil::getCharset (std::string const& str) {
...
+		_uBiDiObj.get(),
+		ustring.getBuffer(),
+		-1,
+		UBIDI_DEFAULT_LTR,
+		nullptr,
+		_unicodeError
+	);
+	if (_unicodeError.isSuccess()) _return = (ubidi_getDirection(_uBiDiObj.get()) != UBIDI_LTR);
+	else {
+		std::clog << "unicode/warning: Error (" << std::to_string(_unicodeError.get()) << ": " << _unicodeError.errorName() << "), determining text direction for: " << str << ", will assume LTR." << std::endl;
```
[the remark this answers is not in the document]
Lord-Kamina: It's an ICU error but warning as far as we're concerned. Anyway, this function is just future-proofing, we're not currently using it anywhere.

game/unicode.hh
```diff
@@ -3,22 +3,28 @@
...
 typedef std::map<std::string, std::string> songMetadata;
 
-struct UnicodeUtil {
+class UnicodeUtil {
+	static icu::ErrorCode m_staticIcuError;
+	enum class CaseMapping { LOWER, UPPER, TITLE, NONE };
+
+	public:
 	UnicodeUtil() {}
 	~UnicodeUtil() {};
```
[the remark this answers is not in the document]
Lord-Kamina: Remove what, the constructor an destructor?

game/unicode.cc
Lord-Kamina: What, the error? Yes, several ICU functions take a reference to an `UErrorCode` (or `icu::ErrorCode`)

game/unicode.cc
Lord-Kamina: Filename is only used for providing info to the user, in case they want to take the time of fixing the encodings.

game/hiscore.hh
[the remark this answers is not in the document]
OznOg: then be consistent and remove the const

game/songitems.cc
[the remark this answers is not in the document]
OznOg: yes you are right, this is not a but, this is actual normal behavior in C/C++ all parameters are evaluated prior to the call, so you are right, this cannot be done the way I suggested it. Maybe a comment will prevent future refactor to fall in the trap.

game/unicode.cc
OznOg: :m_sortCollator and friends ?

game/unicode.cc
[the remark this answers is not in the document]
OznOg: then I suggest just dropping it. If no one uses it, it probably does not work

game/unicode.hh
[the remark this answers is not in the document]
OznOg: yes

game/hiscore.hh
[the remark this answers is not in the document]
Lord-Kamina: Wouldn't it be more correct leave the const as a signal that we don't intend to modify that value, though?

game/unicode.cc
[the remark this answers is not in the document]
Lord-Kamina: Nobody uses it because we don't currently have any translations to RTL languages, but we might in the future.

game/unicode.cc
Lord-Kamina: Ah, yes. They are used when filtering songs.

game/hiscore.hh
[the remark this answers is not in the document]
OznOg: Well C/C++ is done on the wrong way, all should be const and mutability shoudl be explicit.... We have to cope with it, and adding const to every place where passed by copy is done seem awkward and overkill. IMHO adding a const a copy is just WTF for the reader.

game/unicode.cc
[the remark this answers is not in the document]
OznOg: You sculd create a branch with it if you wish, but keeping it here is just useless dead code; YAGNI,  IMHO

game/unicode.cc
OznOg: Ok, that was already like this, but this would deserve rework as well IMHO.

OznOg (dismissed):

game/unicode.cc
twollgam: But where are the other values? There may be others but we use only this one. So why not using a static const? Even we have to use other values in the future they could also be static const. Actually I don't see that this must be dynamic.

Lord-Kamina: The values are returned by `getCharset`and passed directly to the `UnicodeString` constructor.
We check specifically for UTF-8 because there is a different factory method when you know for sure that you're dealing with UTF-8.

game/database.cc
```diff
@@ -24,7 +24,7 @@ void Database::load() {
 		m_songs.load(nodeRoot->find("/performous/songs/song"));
 		m_hiscores.load(nodeRoot->find("/performous/hiscores/hiscore"));
 		std::clog << "database/info: Loaded " << m_players.count() << " players, " << m_songs.size() << " songs and " << m_hiscores.size() << " hiscores from " << m_filename.string() << std::endl;
-	} catch (std::exception& e) {
+	} catch (std::exception const& e) {
```
twollgam, line 27: 👍

game/hiscore.cc
```diff
@@ -42,15 +42,16 @@ Hiscore::HiscoreVector Hiscore::queryHiscore(std::optional<PlayerId> playerid, s
 	return hv;
 }
 
-bool Hiscore::hasHiscore(const SongId &songid) const {
-	return std::any_of(m_hiscore.begin(),m_hiscore.end(), [&songid, level = currentLevel()](auto const& h) { return songid == h.songid && level == h.level; });
+bool Hiscore::hasHiscore(const SongId& songid) const {
+	return std::any_of(m_hiscore.begin(),m_hiscore.end(), [&songid, level = currentLevel()](auto const& h) {
+		return songid == h.songid && level == h.level;
+	});
```
twollgam, line 48: 👍

twollgam (approved):

twollgam (dismissed):

twollgam (approved):

PR #790, opened 2022-07-15
Some more fixes for UnicodeUtil

Lord-Kamina:
### What does this PR do?

Chiefly two things:

1) Manually instantiate UConverters. Why? The UnicodeString constructor we were using was essentially creating and destroying one of these on each invocation. Now we create them when we need them, and they stay around until the app quits.

2) Implement locale-aware sorting and searching. This means the sorting and searching should now behave appropriately according to the language selected in the Performous configuration.

And, there's also some minor fixes and clean-up.

### More

### DIFF
--- a/game/unicode.cc
+++ b/game/unicode.cc
@@ -10,9 +10,8 @@
 #include <unicode/ubidi.h>
 #include "compact_enc_det/compact_enc_det.h"
 
-icu::ErrorCode UnicodeUtil::m_staticIcuError = icu::ErrorCode();
-icu::RuleBasedCollator UnicodeUtil::m_searchCollator (icu::UnicodeString (""), icu::Collator::PRIMARY, m_staticIcuError);
-icu::RuleBasedCollator UnicodeUtil::m_sortCollator  (nullptr, icu::Collator::SECONDARY, m_staticIcuError);
+std::unique_ptr<icu::RuleBasedCollator> UnicodeUtil::m_searchCollator = nullptr;
+std::unique_ptr<icu::RuleBasedCollator> UnicodeUtil::m_sortCollator = nullptr;
 
 std::map<std::string, Converter> UnicodeUtil::m_converters{};
 
--- a/game/unicode.cc
+++ b/game/unicode.cc
@@ -14,12 +14,37 @@ icu::ErrorCode UnicodeUtil::m_staticIcuError = icu::ErrorCode();
 icu::RuleBasedCollator UnicodeUtil::m_searchCollator (icu::UnicodeString (""), icu::Collator::PRIMARY, m_staticIcuError);
 icu::RuleBasedCollator UnicodeUtil::m_sortCollator  (nullptr, icu::Collator::SECONDARY, m_staticIcuError);
 
+std::map<std::string, Converter> UnicodeUtil::m_converters{};
+
+Converter::Converter(std::string const& codepage): m_codepage(codepage), m_converter(nullptr, &ucnv_close) {
+	m_converter = std::unique_ptr<UConverter, decltype(&ucnv_close)>(ucnv_open(m_codepage.c_str(), m_error), &ucnv_close);
+	if (m_error.isFailure()) throw std::runtime_error("unicode/error: " + std::to_string(m_error.get()) + ": " + std::string(m_error.errorName()));
+}
+
+Converter::Converter(Converter&& c) noexcept: m_codepage(std::move(c.m_codepage)), m_converter(std::move(c.m_converter)), m_error(std::move(c.m_error)) {}
+
+void Converter::reset() {
+	std::scoped_lock l(m_lock);
+	if (m_converter) ucnv_reset(m_converter.get());
+}
+
+icu::UnicodeString Converter::convertToUTF8(std::string_view sv) {
+	std::scoped_lock l(m_lock);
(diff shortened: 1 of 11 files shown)


game/unicode.cc
```diff
@@ -10,16 +10,40 @@
 #include <unicode/ubidi.h>
 #include "compact_enc_det/compact_enc_det.h"
 
-icu::ErrorCode UnicodeUtil::m_staticIcuError = icu::ErrorCode();
-icu::RuleBasedCollator UnicodeUtil::m_dummyCollator (icu::UnicodeString (""), icu::Collator::PRIMARY, m_staticIcuError);
-icu::RuleBasedCollator UnicodeUtil::m_sortCollator  (nullptr, icu::Collator::SECONDARY, m_staticIcuError);
+std::unique_ptr<icu::RuleBasedCollator> UnicodeUtil::m_searchCollator = nullptr;
+std::unique_ptr<icu::RuleBasedCollator> UnicodeUtil::m_sortCollator = nullptr;
```
twollgam, line 14: Initialization not needed. Smartpointers are by default Null.

game/unicode.cc
```diff
@@ -10,16 +10,40 @@
...
+
+std::map<std::string, Converter> UnicodeUtil::m_converters{};
+
+Converter::Converter(std::string const& codepage): m_codepage(codepage), m_converter(nullptr, &ucnv_close) {
+	m_converter = std::unique_ptr<UConverter, decltype(&ucnv_close)>(ucnv_open(m_codepage.c_str(), m_error), &ucnv_close);
+	if (m_error.isFailure()) throw std::runtime_error("unicode/error: " + std::to_string(m_error.get()) + ": " + std::string(m_error.errorName()));
+}
+
+Converter::Converter(Converter&& c) noexcept: m_codepage(std::move(c.m_codepage)), m_converter(std::move(c.m_converter)), m_error(std::move(c.m_error)) {}
+
```
twollgam, line 24: I would prefere to spread this over 2 or more lines, cause it is not easy readable.

game/unicode.hh
```diff
@@ -2,30 +2,57 @@
...
 
+class UnicodeString;
+
+struct Converter {
+	/// Constructor/Destructor
+	Converter(std::string const& codepage);
+	Converter(Converter&& c) noexcept;
+	Converter(Converter& c) = delete;
+	Converter(Converter const& c) = delete;	
+	~Converter() {};
```
twollgam, line 24: Not needed cause the compiler will create it automatically.

game/unicode.hh
```diff
@@ -2,30 +2,57 @@
...
+
+	icu::UnicodeString convertToUTF8(std::string_view sv); ///< Do the actual conversion.
+
+	private:
+	std::string m_codepage;
+	std::unique_ptr<UConverter, decltype(&ucnv_close)> m_converter;
+	void reset(); ///< UConverters are state machines and need to be reset after each use.
```
twollgam, line 31: Put it in a new private section or move it up. I think members and functions shouldn't be mixed.

game/unicode.hh
```diff
@@ -2,30 +2,57 @@
...
 	static bool caseEqual (std::string_view lhs, std::string_view rhs, bool assumeUTF8 = false);
 	static bool isRTL(std::string_view str); ///< FIXME: This won't be used for now, but it might be useful if we eventually implement RTL translations. As-is, at least on my mac, Performous is refusing to render Arabic text, although it might be a font issue.
 	static std::string toLower (std::string_view str);
 	static std::string toUpper (std::string_view str);
 	static std::string toTitle (std::string_view str);
-	static icu::RuleBasedCollator m_dummyCollator;
-	static icu::RuleBasedCollator m_sortCollator;
+
+	static std::unique_ptr<icu::RuleBasedCollator> m_searchCollator;
+	static std::unique_ptr<icu::RuleBasedCollator> m_sortCollator;
```
twollgam, line 57: Should be private. I think they should not be accessed from outside. Access should only be indirect by calling a function.

twollgam (changes requested):
The translation stuff should be reviewed by someone else.

game/unicode.cc
Lord-Kamina: It's because they're static.

game/unicode.hh
Lord-Kamina: Which one, the destructor?

game/unicode.hh
Lord-Kamina: Actually I forgot I can probably delete this function because the converter is automatically reset when used.

game/unicode.hh
Lord-Kamina: I think that's just making it overly complicated. These are never actually used directly by UnicodeUtil.

game/unicode.cc
twollgam: Sure, but the `= nullptr` is not needed ;-)

game/unicode.hh
twollgam: Yes, there's always a default dtor if none is defined.

game/unicode.hh
twollgam: OK

game/unicode.hh
twollgam: Ok, for later, if ever.

twollgam (approved):
Reviewed changes again. This time inclusive i18n.cc. LGTM

twollgam (approved):
LGTM

PR #798, opened 2022-07-28
TranslationEngine: fix possible segfault

Lord-Kamina:
@ooshlablu reported seeing a crash on macOS after my second Unicode PR. I haven’t been able to reproduce it but from the trace, I’m thinking the issue must be that it’s trying to use “m_currentLanguage” before the class has finished constructing.

If so, it can probably be fixed by giving it a default value.

### More

### DIFF
--- a/game/i18n.cc
+++ b/game/i18n.cc
@@ -120,7 +120,7 @@ std::string TranslationEngine::getLanguageByKey(const std::string& languageKey) 
 	return allLanguages.at(languageKey);
 }
 
-const std::pair<std::string, std::string>& TranslationEngine::getCurrentLanguage() const { 
+const std::pair<std::string, std::string>& TranslationEngine::getCurrentLanguage() const {
 	return m_currentLanguage; 
 }
 
--- a/game/i18n.cc
+++ b/game/i18n.cc
@@ -1,13 +1,18 @@
 #include "i18n.hh"
 
+#include "config.hh"
 #include "configuration.hh"
 #include "fs.hh"
 #include "unicode.hh"
 
 #include <unicode/locid.h>
 
+std::pair<std::string, std::string> TranslationEngine::m_currentLanguage{"en_US.UTF-8", "English" };
+std::string TranslationEngine::m_package = PACKAGE;
+boost::locale::generator TranslationEngine::m_gen{};
+std::map<std::string, std::string> TranslationEngine::m_languages{};
 
-TranslationEngine::TranslationEngine(const char *package) : m_package(package) {
+TranslationEngine::TranslationEngine() {
 	initializeAllLanguages();
 	/* set all languages in configuration.
 	 * They are kept untranslated internally which prevents having issues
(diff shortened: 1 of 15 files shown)


twollgam (approved):
If this solution works merge it. LGTM

Lord-Kamina:
Turns out I was wrong. The real problem was `getCurrentLanguageCode`was being called form the `MidiParser`, which is constructed and put to work before instantiating fake singleton, resulting dereferencing an empty pointer. I made TranslationEngine static, removing its reliance on `Game` altogether.

@ooshlablu can you confirm if it works for you?

ooshlablu:
This fixes the segfault I was seeing on MacOS

ooshlablu (approved):
LGTM

game/songparser.cc
```diff
@@ -70,16 +70,17 @@ SongParser::SongParser(Song& s): m_song(s) {
 			type = Type::XML;	// XMLPP should deal with encoding so we don't have to.
 		}
 		else {
-			UnicodeUtil::convertToUTF8(m_ss.str(), s.filename.string());
-			if (smCheck (m_ss.str())) {type = Type::SM; } else if (txtCheck (m_ss.str())) {
+			std::string ss = UnicodeUtil::convertToUTF8(m_ss.str(), s.filename.string());
+			if (smCheck (ss)) {type = Type::SM; } else if (txtCheck (ss)) {
```
twollgam, line 74: I would like to see some line breaks ;-)

game/unicode.cc
```diff
@@ -36,10 +36,11 @@ Converter& UnicodeUtil::getConverter(std::string const& s) {
 	return m_converters.try_emplace(s, Converter(s)).first->second;
 }
 
-std::string UnicodeUtil::getCharset (std::string_view str) {
+std::string UnicodeUtil::getCharset (std::string_view& str) {
```
twollgam, line 39: Why (non const) reference here? Is str modified?

game/unicode.cc
```diff
@@ -98,7 +96,7 @@ std::string UnicodeUtil::convertToUTF8 (std::string_view str, std::string _filen
 	return ret;
 }
 
-bool UnicodeUtil::removeUTF8BOM(std::string_view str) {
+bool UnicodeUtil::removeUTF8BOM(std::string_view& str) {
```
twollgam, line 99: Reference needed?

game/songparser.cc
Lord-Kamina: Not saying I won't add them but man I seriously want to know what's wrong with your editor width setting 😂

game/unicode.cc
Lord-Kamina: It can be modified by removeUTF8BOM

twollgam (approved):

game/unicode.cc
Lord-Kamina: Yes because it potentially modifies the string_view to a substring of itself.

game/i18n.cc
```diff
@@ -120,19 +125,24 @@ std::string TranslationEngine::getLanguageByKey(const std::string& languageKey)
...
 
-const std::pair<std::string, std::string>& TranslationEngine::getCurrentLanguage() const { 
-	return m_currentLanguage; 
+std::pair<std::string, std::string> const& TranslationEngine::getCurrentLanguage() { 
+	return TranslationEngine::m_currentLanguage; 
+}
+
+std::string TranslationEngine::getCurrentLanguageCode() {
+	std::string lang(getCurrentLanguage().first);
+	return lang.substr(0, lang.size() - 6);
```
Baklap4, line 134: Whats this minus 6??

Baklap4 (commented):
Overall looks alright
Is the change to static functions really needed? Im always getting itchy if a lot of static is used...

Lord-Kamina:
There are other possible solutions, but.I think they'd require a deeper refactoring and/or to keep relying on`Game`

game/i18n.cc
Lord-Kamina: Sorry for the magic number, it's ".UTF-8"

Lord-Kamina: Maybe it'd be best to calculate the cut-off in relation to the index of the first `.`?

Baklap4 (approved):
LGTM, still on holiday, but people claimed it fixes the issue :)

Issue #797, opened 2022-07-28
Players and highscores not saved/reset on exit


BWagener:
### Performous & operating system versions

Performous: https://github.com/performous/performous/actions/runs/2715581108
OS: Windows 10 Pro N
- Edition	Windows 10 Pro N
- Version	21H2
- Installed on	‎31/‎12/‎2020
- OS build	19044.1826
- Experience	Windows Feature Experience Pack 120.2212.4180.0

### What is the problem?

Highscores and players are not saved in database.xml.

Creating a player kind of works, a player is saved to the database but their name is empty (see attached files)

### Steps to reproduce

1. Start game
2. Sing song
3. Create player to save highscore
4. Exit game

### Log(s) & preferences file(s)

[infolog.txt](https://github.com/performous/performous/files/9211495/infolog.txt)
[config.xml](https://github.com/performous/performous/files/9211514/config.txt)
[database.xml](https://github.com/performous/performous/files/9211515/database.txt)

BWagener:
I tested the most recent builds and the last working build in regards to highscore saving is the one for this commit 

https://github.com/performous/performous/commit/843511829d5e237534769426a6ca49250a7b34bf
while all builds since, starting with
https://github.com/performous/performous/commit/1512c3c97fe9bf44377573ac58ee45d6c6206440
have this bug.

PR #801, opened 2022-07-29
Database: fix non-saving players and hiscores.

Lord-Kamina:
### What does this PR do?

This turned out to be an interesting one… there were some  checks left over from the optional commits. Since we dropped that idea, we were casting players ang song ids to bool, and for some reason the compiler didn’t whine about it.

Point is, put a wrench in the code when the first player’s id (0) got converted to false.

Anyway, this should fix it, and I’ve also introduced `std::max_element` per @twollgam’s suggestion.

### Closes Issue(s)

#797 

### More

### DIFF
--- a/game/hiscore.cc
+++ b/game/hiscore.cc
@@ -83,7 +83,6 @@ void Hiscore::load(xmlpp::NodeSet const& nodes) {
 
 void Hiscore::save(xmlpp::Element *hiscores) {
 	for (auto const& h: m_hiscore) {
-		if (!h.playerid) continue;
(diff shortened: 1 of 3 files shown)


twollgam (approved):
LGTM

If the reporter of the error approve the fix is working then merge it!

BWagener:
> LGTM
> 
> If the reporter of the error approve the fix is working then merge it!

It's working perfectly now!

Baklap4 (approved):
LGTM, OP from #797 also said this fixed the issue.

PR #802, opened 2022-07-29
Move SDL_Quit() to the Window destructor.

### LINKED ISSUE #800: SDL_GLCreateContext failure

Performous
macOS Monterrey

### What is the problem?

SDL fails to create an OpenGL Context.
I can't reproduce, @ooshlablu says it apparently happens after crashing previously.

My theory is this happens because we're not invoking `SDL_Quit()` when quitting during an exception. Maybe we should add it to `~Window()` instead of using `std::atexit`.

### Steps to reproduce

_No response_

### Additional context

This is a remind for myself but if anyone else wants to give it a go before I do, you're welcome to.

_No response_

Lord-Kamina:
### What does this PR do?

We posted this bug upstream over at https://github.com/libsdl-org/SDL/issues/5912 but nobody could reproduce it. @ooshlablu later said this happens after a crash. My hypothesis is we were not cleaning SDL up on abnormal exit, which could maybe cause the issue?

I'm mentioning upstream also so it can be definitely documented that this was the cause, in case this does work.

### Closes Issue(s)

Theoretically, close #800.

### More

### DIFF
--- a/game/video_driver.cc
+++ b/game/video_driver.cc
@@ -60,7 +60,6 @@ GLuint Window::m_vbo = 0;
 GLint Window::bufferOffsetAlignment = -1;
 
 Window::Window() : screen(nullptr, &SDL_DestroyWindow), glContext(nullptr, &SDL_GL_DeleteContext) {
-	std::atexit(SDL_Quit);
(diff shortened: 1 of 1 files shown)


ooshlablu:
This fixes the SDL error for me.

ooshlablu (approved):
LGTM

PR #803, opened 2022-08-03
fix showing highscore input player name with instruments

Darkmagister:
### What does this PR do?

adding again the screen at the end of a song for the highscore with instrument not only with vocal

### Motivation

i usually use with with instrument and almost never for sing, so it was frustrating not having any highscore 

### Additional Notes

i only tested it with few songs with and without vocal and it seems to be working
i had to edit the getVocalTrack function otherwise it will crash on track without vocal

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -90,7 +90,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          upload_url: ${{ needs.create_release.outputs.upload_url }}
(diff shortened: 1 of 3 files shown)


Baklap4:
Please split the dark-green theme into a separate PR

Darkmagister:
> Please split the dark-green theme into a separate PR

sorry i've made a little mess with the commit and branches, i think i've fixed it!!

ooshlablu:
@Darkmagister Can you rebase this to current master and fix the conflicts? I'll be able to test it when it's updated, hopefully we'll get it merged quickly.

PR #806, opened 2022-08-10
Fix a bug in ConfigItem::addEnum

### LINKED ISSUE #812: Crash on duet songs

Performous: https://github.com/performous/performous/actions/runs/2863890125
OS: Windows 10 Pro N
- Edition	Windows 10 Pro N
- Version	21H2
- Installed on	‎31/‎12/‎2020
- OS build	19044.1826
- Experience	Windows Feature Experience Pack 120.2212.4180.0

### What is the problem?

The game crashes whenever a song is started in duet mode. The songs I've tested with are in UltraStar format.

### Steps to reproduce

1. Start game
2. Enter song selection
3. Hit enter on "Duet mode" or a duet part on a duet song
image
4. Game crashes

### Additional context

_No response_

[config.xml](https://github.com/performous/performous/files/9394784/config.txt)
[database.xml](https://github.com/performous/performous/files/9394785/database.txt)
[infolog.txt](https://github.com/performous/performous/files/9394787/infolog.txt)
[Cat Stevens - Father and Son.txt](https://github.com/performous/performous/files/9394850/Cat.Stevens.-.Father.and.Son.txt)
[Cat Stevens - Father and Son [MULTI].txt](https://github.com/performous/performous/files/9394851/Cat.Stevens.-.Father.and.Son.MULTI.txt)

twollgam:
### What does this PR do?

In ConfigItem::addEnum a typo is fixed that results in not initializing
m_step so that calculations do division by zero.

Also ConfigItem get setters for some values to deny setting the values
by getters.

### Closes Issue(s)

Closes #812 -> Division by zero since `m_step` wasn't correctly set.

### DIFF
--- a/game/configitem.cc
+++ b/game/configitem.cc
@@ -37,24 +37,44 @@ ConfigItem::ConfigItem(OptionList opts)
 : m_type("option_list"), m_value(opts) {
 }
 
-ConfigItem::NumericValue& ConfigItem::getMin() {
+ConfigItem::NumericValue ConfigItem::getMin() const {
 	return m_min;
 }
 
-ConfigItem::NumericValue& ConfigItem::getMax() {
+void ConfigItem::setMin(NumericValue min) {
+	m_min = min;
+}
+
+ConfigItem::NumericValue ConfigItem::getMax() const {
 	return m_max;
 }
 
-ConfigItem::NumericValue& ConfigItem::getStep() {
+void ConfigItem::setMax(NumericValue max) {
+	m_max = max;
+}
+
+ConfigItem::NumericValue ConfigItem::getStep() const {
 	return m_step;
 }
 
-ConfigItem::NumericValue& ConfigItem::getMultiplier() {
+void ConfigItem::setStep(NumericValue step) {
+	m_step = step;
+}
+
+ConfigItem::NumericValue ConfigItem::getMultiplier() const {
 	return m_multiplier;
 }
 
-std::string& ConfigItem::getUnit() {
+void ConfigItem::setMultiplier(NumericValue multiplier) {
+	m_multiplier = multiplier;
+}
+
+std::string const& ConfigItem::getUnit() const {
 	return m_unit;
+}
+
+void ConfigItem::setUnit(std::string const& unit) {
+	m_unit = unit;
 }
 
 ConfigItem& ConfigItem::incdec(int dir) {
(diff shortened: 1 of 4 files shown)


game/configitem.cc
```diff
@@ -37,26 +37,46 @@ ConfigItem::ConfigItem(OptionList opts)
 : m_type("option_list"), m_value(opts) {
 }
 
-ConfigItem::NumericValue& ConfigItem::getMin() {
+ConfigItem::NumericValue ConfigItem::getMin() const {
 	return m_min;
 }
 
-ConfigItem::NumericValue& ConfigItem::getMax() {
```
Lord-Kamina, line 44: I don't like adding setters. I think it's unnecessary clutter for something that will be used maybe two times in the code.

game/configitem.cc
```diff
@@ -193,19 +213,20 @@ std::string const ConfigItem::getValue() const {
...
+void ConfigItem::addEnum(std::string const& name) {
 	verifyType("uint");
-	if (find(m_enums.begin(),m_enums.end(),name) == m_enums.end()) {
+	if (find(m_enums.begin(), m_enums.end(), name) == m_enums.end())
 		m_enums.push_back(name);
-	}
+
 	m_min = static_cast<unsigned short>(0);
 	m_max = static_cast<unsigned short>(m_enums.size() - 1);
-	m_min = static_cast<unsigned short>(1);
```
Lord-Kamina, line 203: Facepalm. Good job catching that.

twollgam: Don't know how this could happen. My few unit test didn't catch it.

game/configitem.cc
twollgam: @Baklap4  @OznOg what do you think?

My arguments from discord:

I don't think so, cause the setters make the code more readable , findable and understandable. And modifying values from outside is a bad habit/style (except for pods) cause the class should be the master of its data.

Baklap4: since i stepped away from java and got introduced into c# i hated getters/setters from the gecko since you're just internally setting/getting a member field.

Yes if you have to do some funky stuff like lazy loading sure that's a valid method. But for just assigning + reading i'm with @Lord-Kamina 

This is especially true with this syntax (in C#):
```csharp
public int SomeValue { get; init; } // or
public int SomeOtherVal { get; private set; }
```

The first syntax only allows the constructor to fill the `SomeValue` member whilst being publicly gettable.
The second syntax only allows for private sets whilst being publicly gettable

if something like the above is available in c++ without the boilerplate of getter/setter methods i'm more than willing to use that approach

Lord-Kamina: I don't think that can be done in C++.

Also, keep in mind that this proposed interface is due to a refactor. Originally, it was the ConfigItem itself that set these values

twollgam: That is all syntactic suggar not available in c++. There you have 
```c++
public int get_SomeValue();
public void set_SomeValue(int);
```
like in c++ COM. Found some other solutions here https://stackoverflow.com/questions/8368512/does-c11-have-c-style-properties

OK, if we do not want setters here, than I would prefer also no getters. So we come back to direct access to the members? 
@Baklap4  and @Lord-Kamina what do you think?

Baklap4: Hmmm i think this gets way too complex for such a simple thing. So i think it'd be best to resort to just using member fields like we always did

I do know we have some get/set methods but those either have more stuff than just returning or setting a member field. If that's not the case though i think it'd be best to also resolve them into just member fields only

Lord-Kamina:
This looks better to me now; maybe I would have used a friend class instead of just `public` but this also works. In any case, I'd like if possible to merge #814 first, because that will likely speed PR rebasing significantly. Opinion, @Baklap4?

Baklap4 (approved):
LGTM, no need to wait for #814 as that one still has some discussions open.

PR #810, opened 2022-08-13
Add .vscode to .gitignore

qreeves:
Add `.vscode` to `.gitignore`.


Baklap4:
Please dont do this, the .vscode folder can contain launch profiles which can be made functioning for everyone. The .vs folder however can be ignored

qreeves:
Actually did not know that, always assumed it was a user config due to the auto creation.

PR #813, opened 2022-08-22
Fix a couple things on macOS bundling and CI.

Lord-Kamina:
### What does this PR do?

Fixes some things on the macOS bundling script and makes it more consistent.
Also, should fix the macOS CI builds failing due to ffmpeg5 being default in homebrew.

### More


Baklap4 (approved):
LGTM, but still awaiting pipelines :) if all green go ahead and merge

PR #814, opened 2022-08-24
Test using a different vcpkg action for CI

Lord-Kamina:
### What does this PR do?

Testing an alternative GH action for building dependencies and caching.

### More

- [x] Added/updated documentation


Lord-Kamina:
@Baklap4 There's some debug stuff I need to remove but as I said on Discord... I feel we're using the manifest mode incorrectly and we're really better off switching to what I've done in this PR; not to mention after the cache has been built, it takes a couple of minutes to get a windows build, and the cache should only need to be rebuilt when it expires or when our actual dependencies change.

Lord-Kamina:
@Baklap4 MSVC tests issue notwithstanding, I think I'm done here.

.github/workflows/build_and_release.yml
```diff
@@ -470,31 +470,44 @@ jobs:
...
       - name: Setup our dev environment
-        uses: ilammy/msvc-dev-cmd@v1.4.1
+        uses: ilammy/msvc-dev-cmd@v1
 
       - name: Setup ninja cmake generator
         uses: seanmiddleditch/gha-setup-ninja@master
 
       - name: Download and build our dependencies using vcpkg
-        uses: lukka/run-vcpkg@v10.2
+        uses: Lord-Kamina/vcpkg-action@master
```
Baklap4, line 479: What version of vcpkg will be used here? if master then this will cause rebuilds of the cache once any of our libraries gets updated.
Preferably i'd like to specify the version we want to use so that the action always resolves to a specific version instead of latest. This way we can keep consistent builds.

.github/workflows/build_and_release.yml
```diff
@@ -470,31 +470,44 @@ jobs:
...
         with:
-          vcpkgGitCommitId: '14e7bb4ae24616ec54ff6b2f6ef4e8659434ea44'
-          runVcpkgInstall: true
+          pkgs: aubio boost-iostreams boost-locale boost-program-options boost-system cairo cpprestsdk ffmpeg fftw3 fmt fontconfig gdk-pixbuf gettext glibmm glm gtest icu libepoxy librsvg libsigcpp libxml2 libxmlpp nlohmann-json openblas opencv openssl pango portaudio portmidi pthreads sdl2
+          triplet: x64-windows-release
+          extra-args: --clean-after-build
+          cache-key: win64-vcpkg
```
Baklap4, line 484: Personally i'd just run several bash/powershell commands to download, setup and build vcpkg instead of a github action. This way it's clear for everyone what's happening in the background and probably easier to debug aswell since the action can contain bugs/failures when vcpkg decides to update something

.github/workflows/build_and_release.yml
```diff
@@ -470,31 +470,44 @@ jobs:
...
+          cd "performous/build/x64-release/testing"
+          ./performous_test.exe --gtest_filter=UnitTest*
+      - name: Run unittests by ctest.
+        shell: powershell
+        run: |
+          cd "performous/build/x64-release"
+          ninja testing/test
       - name: Create Installer
         id: package_performous_x64
+        shell: powershell
```
Baklap4, line 507: the shell is per definition powershell on windows machines so no need to specify this

.github/workflows/build_and_release.yml
```diff
@@ -470,31 +470,44 @@ jobs:
...
+          set VCPKG_DEFAULT_BINARY_CACHE=${{ github.workspace }}/vcpkg_cache
           cd ${{ github.workspace }}/performous
           echo $PERFORMOUS_VERSION
-          cmake --preset "x64-release"
-          cmake --build --preset "x64-release"
-
+          cmake --preset "x64-release" -DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake -DVCPKG_TARGET_TRIPLET=x64-windows-release
+          cmake --build --preset "x64-release" -DVCPKG_TARGET_TRIPLET=x64-windows-release -DCMAKE_TOOLCHAIN_FILE=${{ github.workspace }}/vcpkg/scripts/buildsystems/vcpkg.cmake
+      - name: Run unittests directly called.
+        shell: powershell
```
Baklap4, line 496: the shell is per definition powershell on windows machines so no need to specify this

.github/workflows/build_and_release.yml
```diff
@@ -470,31 +470,44 @@ jobs:
...
+        run: |
+          cd "performous/build/x64-release/testing"
+          ./performous_test.exe --gtest_filter=UnitTest*
+      - name: Run unittests by ctest.
+        shell: powershell
```
Baklap4, line 501: the shell is per definition powershell on windows machines so no need to specify this

CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ if (APPLE)
 endif()
 
 if(WIN32)
+	set(MINGW_SYSTEM_PREFIX "C:/msys64/mingw64" CACHE PATH "Path to the mingw-w64 system root.")
```
Baklap4, line 60: mingw isn't just available for windows please set the paths aswell for unix/osx

CMakeLists.txt
```diff
@@ -188,8 +189,10 @@ if (APPLE)
 	set(CPACK_SOURCE_GENERATOR "")
 	set(CPACK_GENERATOR "Bundle")
 	set(CMAKE_XCODE_ATTRIBUTE_MARKETING_VERSION "${PROJECT_VERSION}")
+elseif(WIN32)
+	set(CPACK_GENERATOR "NSIS64")
```
Baklap4, line 193: please leave this within the else statement. nsis isn't specific for just windows.

CMakePresets.json
```diff
@@ -13,8 +13,6 @@
         "BUILD_SHARED_LIBS": "ON",
         "ENABLE_WEBSERVER": "OFF",
         "SELF_BUILT_AUBIO": "ALWAYS",
-        "CMAKE_INSTALL_PREFIX": "${sourceDir}/build/${presetName}-install",
-        "CMAKE_TOOLCHAIN_FILE": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake",
```
Baklap4, line 17: Please keep these lines since else it'll break visual studio users. You can always overwrite these at the actual cmake script.

CMakePresets.json
```diff
@@ -37,48 +35,33 @@
...
-    },
-    {
-      "name": "x86-release",
-      "displayName": "x86 Release",
-      "description": "Target Windows Release (32-bit).",
-      "inherits": "x86-debug",
+      "name": "x64-debinfo",
+      "displayName": "x64 Release with Debug info",
+      "description": "Target Windows Release with Debug info (64-bit).",
+      "inherits": "x64-debug",
```
Baklap4, line 41: Shouldn't inherit from debug. Instead it should inherit from release whilst also keeping the debug symbols (it's an option i think)

game/CMakeLists.txt
```diff
@@ -12,6 +12,21 @@ if(WIN32)
...
+		if(DEFINED ENV{ORIGINAL_TEMP})
+			set (TMP_LOCATION $ENV{ORIGINAL_TEMP})
+		elseif(DEFINED ENV{TEMP})
+			set (TMP_LOCATION $ENV{TEMP})
+		else()
+			message(WARNING "Temp folder not found. CPack won't be able to create a proper installer.")
+		endif()
+		file(WRITE "${TMP_LOCATION}/source.txt" "${CMAKE_SOURCE_DIR}")
+		file(WRITE "${TMP_LOCATION}/mingwroot.txt" "${CPACK_MINGW_SYSTEM_PREFIX}")
+	endif()
```
Baklap4, line 26: cpack by default builds/installs into your defined build and install prefix. No need to redefine that logic...

game/CMakeLists.txt
```diff
@@ -12,6 +12,21 @@ if(WIN32)
...
+	if (CMAKE_BUILD_TYPE STREQUAL "Release")
+		set (CPACK_STRIP_FILES TRUE)
+	endif()
```
Baklap4, line 29: This should live within the main cmakelists file

game/CMakeLists.txt
```diff
@@ -46,10 +61,8 @@ if(WIN32)
 		# According to MinGW tools, we need to compile the rc file, and then link it into projects:
 		# windres foo.rc foores.o
 		# gcc -o foo.exe foo.o foores.o
-		find_library(WSOCK32_LIBRARY wsock32)
-		find_library(WS2_32_LIBRARY ws2_32)
 		if(NOT CMAKE_RC_COMPILER)
-			find_program(CMAKE_RC_COMPILER windres)
+			find_program(CMAKE_RC_COMPILER "${CPACK_MINGW_SYSTEM_PREFIX}/bin/windres")
```
Baklap4, line 65: This breaks the build with msvc as it was found previously

game/CMakeLists.txt
```diff
@@ -106,6 +119,11 @@ set(SELF_BUILT_JSON "NEVER" CACHE STRING "rule to allow self build of Json depen
 set(Aubio_REQUIRED_VERSION "0.4.9")
 
 # Find all the libs that don't require extra parameters
+
+if (MINGW)
+target_link_libraries(performous PRIVATE mingw32)
```
Baklap4, line 124: Forgot an indent and shouldn't we link against mingw64?

game/CMakeLists.txt
```diff
@@ -182,6 +200,7 @@ if("AUTO" STREQUAL "${ENABLE_WEBSERVER}")
 		message(STATUS "Webserver support: Enabled (automatically found)")
 	else()
 		message(STATUS "Webserver support: Disabled (cpprestsdk not found)")
+		SET (ENABLE_WEBSERVER OFF FORCE)
```
Baklap4, line 203: This shouldn't be necessary?

game/CMakeLists.txt
```diff
@@ -303,4 +311,38 @@ if(WIN32)
...
+		elseif(DEFINED ENV{TEMP})
+			set (TMP_LOCATION $ENV{TEMP})
+		else()
+			message(WARNING "Temp folder not found. CPack won't be able to create a proper installer.")
+		endif()
+		file (READ "${TMP_LOCATION}/source.txt" CPACK_TOP_SOURCE_DIR)
+		file (READ "${TMP_LOCATION}/mingwroot.txt" CPACK_MINGW_SYSTEM_PREFIX)
+		message ("SOURCE DIR IS: ${CPACK_TOP_SOURCE_DIR}")
+		message ("HOME DIR IS: ${CMAKE_HOME_DIRECTORY}")
+		message ("INSTALL DIR IS: ${CMAKE_INSTALL_PREFIX}")
```
Baklap4, line 328: This should all be done automaticly. See previous cpack_install_dir/build_dir comment

testing/CMakeLists.txt
```diff
@@ -27,7 +27,15 @@ target_link_libraries(performous_test PRIVATE ${Boost_LIBRARIES})
 find_package(fmt REQUIRED CONFIG)
 target_link_libraries(performous_test PRIVATE fmt::fmt)
 
-target_link_libraries(performous_test PRIVATE GTest::GTest GTest::Main)
+if (${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.23)
+	target_link_libraries(performous_test PRIVATE GTest::gtest GTest::gtest_main GTest::gmock)
+elseif(${CMAKE_VERSION} VERSION_GREATER_EQUAL 3.20)
+	target_link_libraries(performous_test PRIVATE GTest::gtest
+ GTest::gtest_main)
```
Baklap4, line 34: Please single line this

testing/CMakeLists.txt
```diff
@@ -27,7 +27,15 @@ target_link_libraries(performous_test PRIVATE ${Boost_LIBRARIES})
...
+else()
+	target_link_libraries(performous_test PRIVATE GTest::GTest
+ GTest::Main)
```
Baklap4, line 37: Please single line this

Baklap4 (changes requested):
Some minor things still.
Aside from that lease get rid of the `win32/mingw64/conf.d` folder. These can all be copied over upon releasing just like the `win32/mingw64/fonts.conf`
As i think we didn't change them from wherever you've copied them

Baklap4:
Also i'm missing the change to the readme where people can download the latest mingw64 build

Lord-Kamina:
> Some minor things still. Aside from that lease get rid of the `win32/mingw64/conf.d` folder. These can all be copied over upon releasing just like the `win32/mingw64/fonts.conf` As i think we didn't change them from wherever you've copied them

I did change them (only slightly though) and I copied them because they're the specific windows files. When we did MXE builds, we were shipping linux config files (That still sorta worked but were not technically correct). I could copy them over from the mingw tree but we'd still need to run sed or something on them for a couple changes.

--vcpkg.j_son
```diff
@@ -43,5 +43,5 @@
     "aubio",
     "fmt"
   ],
-  "builtin-baseline": "14e7bb4ae24616ec54ff6b2f6ef4e8659434ea44"
+  "builtin-baseline": "acc3bcf76b84ae5041c86ab55fe138ae7b8255c7"
```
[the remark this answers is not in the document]
Lord-Kamina: Yes, sorry. I meant to delete this file, actually because AFAIK, the file solely existing activates manifest mode.

I don't like manifest mode because it makes caching a lot harder and/or hacky. The point of manifest mode is that it keeps a VCPKG tree for each project/port using it. What `lukka/run-vcpkg` is doing is hackily using manifest mode, together with an undocumented option to override the install path.

Frankly. I think it'd be much cleaner to update the README/wiki or even make a batch file to fetch the necessary dependencies for the user (could probably be largely copied from the CI itself)

vcpkg-action installs is globally in the same way Linux and macOS package managers work. By default, it uses latest master; I forked it and changed it because at the time I did this, master was broken. My version defaults to the latest stable release (i.e. it's a bit outdated now but would be automatically updated when a new version is released). We could force it to the current master, though.

Baklap4:
> I could copy them over from the mingw tree but we'd still need to run sed or something on them for a couple changes.

What changes are needed? Maybe propose them upstream so we get them into mingw itself instead of dealing with it ourselves?

.github/workflows/build_and_release.yml
Lord-Kamina: See above.

--vcpkg.j_son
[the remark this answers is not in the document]
Lord-Kamina: To be more specific, from the README of vcpkg-action: 
```
I created a simple GitHub action designed to accurately cache packages by using a "dry-run" build to generate an accurate list of packages and package versions. The dry-run is hashed as part of the cache restore key, so the cache will always accurately reflect the packages that need to be built.
```

.github/workflows/build_and_release.yml
Lord-Kamina: Related again to above. By not specifying a git ref, my version will always use the latest stable release. One should hope that stable releases will be... well, stable.

.github/workflows/build_and_release.yml
Lord-Kamina: Same as above.

--vcpkg.j_son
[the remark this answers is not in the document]
Baklap4: > because AFAIK, the file solely existing activates manifest mode.

You're correct. If it exists it will be running in manifest mode.

> Frankly. I think it'd be much cleaner to update the README/wiki or even make a batch file to fetch the necessary dependencies for the user (could probably be largely copied from the CI itself)

We also have this for linux/osx so yes this can be a solution 👍 

Personally i'm more fond of dealing with vcpkg.exe directly instead of through a github action like mentioned below
However if i'm the only one advocating for that let's use the GH-action

So basicly the todo's here are:

- [ ] 1. Change wiki to add a vcpkg.exe command to install the specified libraries from the correct vcpkg version
- [ ] 2. Remove the manifest file

.github/workflows/build_and_release.yml
Baklap4: Pretty sure else the script below wouldn't work aswell.
Maybe instead of setting it for each step, define it globally within this pipeline. The GH-action you're using works with either linux, windows or osx

game/CMakeLists.txt
Lord-Kamina: No. The library/headers are called mingw32 mingw-w64 is the name of the project that forked the original mingw32 code. This still comes from the `x86_64` version of the package.

game/CMakeLists.txt
Lord-Kamina: I added it in an attempt to have performous not link against windows networking libraries when building without the webserver.

game/CMakeLists.txt
Lord-Kamina: It is not done automatically and I swear I lost almost the entire weekend trying to get it to work. It's a fucking nightmare.

testing/CMakeLists.txt
Lord-Kamina: Will do

testing/CMakeLists.txt
Lord-Kamina: Will do also

game/CMakeLists.txt
Baklap4: Alrighty thanks for clarifying!, Then just fix the indent :)

.github/workflows/build_and_release.yml
Lord-Kamina: > Also i'm missing the change to the readme where people can download the latest mingw64 build

Forgot those, will add them also.

Lord-Kamina:
> > I could copy them over from the mingw tree but we'd still need to run sed or something on them for a couple changes.
> 
> What changes are needed? Maybe propose them upstream so we get them into mingw itself instead of dealing with it ourselves?

It's chiefly just one line I think, `<dir prefix="relative">./fonts</dir>` but it's a performous specific thing.

CMakeLists.txt
Lord-Kamina: That'd be a tad different because we'd be cross-compiling, though? Also, it's just the default. In fact, if you look at the CI run, I'm explicitly overriding this path.

CMakeLists.txt
Lord-Kamina: You got it around backwards. I made the difference because the others fail on windows. I could add NSIS to the else ALSO, though.

Baklap4: TGZ, ZIP should work on windows provided you've got that installed ofc

CMakeLists.txt
Baklap4: You're correct, then we'd be cross-compiling (which still should work i guess)
Ahh you have to define it to override it?

CMakeLists.txt
Lord-Kamina: Fair enough.

CMakePresets.json
Lord-Kamina: Release inherits from Debug anyway, just changing the release type but if it makes more sense to you this way, sure.

game/CMakeLists.txt
Lord-Kamina: The problem is actually that CPACK doesn't know the source directory where the top-level `CMakeLists.txt` is. This is chiefly a hack for that.

CMakeLists.txt
Lord-Kamina: > Ahh you have to define it to override it?

I don't think so, actually. But the idea was precisely to make it easier to use on a vanilla msys2 install (so by default it uses the default installation path on windows). You just made me realize a bug though because it's missing the `CPACK_` prefix.

Lord-Kamina:
@Baklap4 I think I've addressed most (if not all) issues. If you're ok now, the only thing left would be to rebase/tidy the history a bit because 13 commits for this is kinda insane.

Baklap4:
Aight seems alright, feel free to rebase it in less commits :)!

Lord-Kamina:
> Aight seems alright, feel free to rebase it in less commits :)!

Done.

Baklap4 (approved):
LGTM

PR #816, opened 2022-09-01
Fix/mute vocal track

### LINKED ISSUE #811: Cannot Mute Vocal Track

Performous: 1.2.0
Operating System: Windows 10 Pro 21H1

### What is the problem?

According to this [page](https://github.com/performous/performous/wiki/Real-karaoke) you should be able to mute the vocal track. I am unable to mute any of the vocal tracks for the songs that have them. Additionally, I cannot find the scripts mentioned at the above page in the install directory.

### Steps to reproduce

Start a song (music.ogg) that has a separate vocal track (vocals.ogg).
Try to press CTRL+V or CTRL+SHIFT+V.
Vocal Track does not mute.

### Additional context

It would be great to have a setting that automatically disables the vocal track (or have this automatically off for Karaoke mode) or if it could work like SingStar, where if it detects someone singing, to lower the volume of the vocal track.

[infolog.txt](https://github.com/performous/performous/files/9355981/infolog.txt)
[config.xml.txt](https://github.com/performous/performous/files/9355987/config.xml.txt)

Baklap4:
### What does this PR do?

* Updates vcpkg to the latest master version.
* Fixes a bug where the vocal track would always be set to 1.0 instead of the toggled setting the user setted.
    * Use `ctrl + v` to toggle the mute/unmute of the vocal track.
    * Use a song which contains a `vocals.ogg` file. 
* Adds an entry to config to save the prefered way for this.

### Closes Issue(s)

Closes #811

### Motivation

This bug was introduced a long time ago and ever since it didn't work anymore. Thanks for @Cossey to point this out

### More

- [x] Updated https://github.com/performous/performous/wiki/Real-karaoke to match the correct behavior

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -389,6 +389,10 @@ to save the current settings to XML.
 		<short>Suppress center channel</short>
 		<long>Suppress audio of center channel (e.g. vocals).</long>
 	</entry>
+	<entry name="audio/mute_vocals_track" type="bool" value="false">
+		<short>Mute the vocals track</short>
+		<long>Mute the vocals track if a vocals.ogg is found.</long>
+	</entry>
 
 	<!-- Paths -->
 	<entry name="paths/songs" type="string_list" hidden="false">
--- a/game/screen_sing.cc
+++ b/game/screen_sing.cc
@@ -246,7 +246,12 @@ void ScreenSing::instrumentLayout(double time) {
 			if (cs.first > 0) level = cs.second / cs.first;
 			if (m_song->music.size() <= 1) level = std::max(0.333, level);
 		}
-		m_audio.streamFade(name, level);
+		if (name == "Vocals") {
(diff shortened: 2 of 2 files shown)


Cossey:
Thanks! When this gets merged when would the next release be?
I only ask because I use this on Windows and don't have the tooling setup for building.

Baklap4:
> when would the next release be?

For every merge to master a 'latest' release will be made (this won't be tagged), but you can download the binary from the readme file :)

ooshlablu:
Tested with vocals + guitar on Ubuntu 20.04 with the db and appimage. Didn't notice any detrimental effects to gameplay. I didn't know that there was an option to mute tracks, but it definitely works in this PR. could come in handy at some point :-)

LGTM

ooshlablu (approved):
LGTM

PR #817, opened 2022-09-04
improve guitar frets, fix shader lighting model

qreeves:
### What does this PR do?

- Adds *basic* MTL support for OBJ files.
- Adds support for color keyed vertices on `3dobject`.
- Adds new `3dobjkey` shader which uses the the keyed parameters.
- Fixes the lighting model in `core.frag` so shading is correct.

### Motivation

Wanting to improve the look and feel of Performous, especially with regard to instrument tracks.

### Additional Notes

As OBJ files do not natively support color keying, any material name starting with `ColorKey` will be set as a keyed color. This allows having fret objects which do not color the entire model, but rather specific sections of the model which take on color with this attribute.


ooshlablu:
Tested with vocals + guitar with the deb and appimage on Ubuntu 20.04. Didn't notice any issues. Polish looks great!

data/shaders/core.frag
```diff
@@ -47,37 +47,48 @@ void main() {
...
-	vec3 refl = reflect(-l, n);
+	vec3 refl = -reflect(l, n);
 	float spec = dot(refl, n);
 	if (power > 0.0) {
-	power *= pow(spec, 100);
-	#ifdef ENABLE_SPECULAR_MAP
-	power *= texture(specularTex, fragIn.texCoord);
-	#endif
-	frag.rgb += vec3(power, power, power);
+		power *= pow(spec, 100);
```
Lord-Kamina, line 83: Tabs shall be your undoing 😂

Lord-Kamina (changes requested):
I don't see any issues other than those erroneous indents. @Baklap4?

data/shaders/core.frag
Lord-Kamina: I seem to have pointed my comment anywhere but where the actual issue is.

Baklap4:
LGTM except for above mentioned indents

Baklap4 (changes requested):
I'm not quite sure why the workflow fails on this branch...
Is it possible to rebase this branch on latest master and see if anything changes?

qreeves:
My branch was force pushed and I've officially lost interest in contributing now.

Lord-Kamina:
@qreeves I don't think it was force-pushed, except possibly rebased against master (which shows up as a force push but is done from the PR interface itself) because the repository settings required that in order for it to be mergeable? (And we've turned that off now)

Lord-Kamina:
@qreeves i.e. This: 
5D5E2C95-015E-4586-87BE-23DA36F66EA0 (old picture, for reference)

PR #824, opened 2022-09-10
Fix FONTCONFIG_PATH on windows.

Lord-Kamina:
Small error I made for the MinGW-w64 builds.

### More

### DIFF
--- a/game/platform.cc
+++ b/game/platform.cc
@@ -17,7 +17,7 @@ std::uint16_t Platform::shortcutModifier(bool eitherSide) {
 
 Platform::Platform() {
 	#if (BOOST_OS_WINDOWS)
-	_putenv_s("FONTCONFIG_PATH",".\\etc\\");
+	_putenv_s("FONTCONFIG_PATH",".\\");
 	#endif
 }
 


Lord-Kamina:
@qreeves could you please test this and check whether there's any issue with fonts?

game/platform.cc
```diff
@@ -17,7 +17,7 @@ std::uint16_t Platform::shortcutModifier(bool eitherSide) {
 
 Platform::Platform() {
 	#if (BOOST_OS_WINDOWS)
-	_putenv_s("FONTCONFIG_PATH",".\\etc\\");
+	_putenv_s("FONTCONFIG_PATH",".\\");
```
Baklap4, line 20: To what is this relative? The executable? AppData? ProgramData?

Baklap4: Just checked this. For vcpkg the path should be: `<vcpkg_root>/installed/<triplet>/etc`  not sure for mingw64

Lord-Kamina: Relative to the executable. I just put it in the same dir because it doesn't really make much sense to have an etc dir just to put a fonts dir under which to put the config files IMO.

Baklap4 (approved):
LGTM

PR #830, opened 2022-09-21
Remove boost for case conversion and trim

twollgam:
### What does this PR do?

The pr removes the dependency for boosts case conversion. It is part of issue #39.

### Closes Issue(s)

None, but is part of #39.

### Motivation

Remove dependency of boost.

### DIFF
--- a/game/util.cc
+++ b/game/util.cc
@@ -1,5 +1,6 @@
 #include "util.hh"
 
+#include <algorithm>
 #include <iomanip>
 #include <sstream>
 #include <stdexcept>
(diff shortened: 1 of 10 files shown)


Baklap4 (approved):
LGTM, @ooshlablu any chance you can test this PR and see if we don't run into weird errors regarding case conversion. Some things to check out is the cache directory, and if all different songparsers still work correctly

ooshlablu:
I tested this with guitar/bass and vocals using the AppImage and debs on Ubuntu 20.04 & 22.04, and everything looks fine to me. Game play seems to be unchanged, search seems to work how it did before, setting config options seems fine, and all of my songs are found and the ones I tried work fine.

LGTM

twollgam:
@Baklap4 please review the issue again. I think I rebased the pr what causes the need for a new review :-(

game/util.cc
```diff
@@ -32,3 +36,112 @@ std::string format(std::chrono::seconds const& unixtime, std::string const& form
...
+	return s;
+}
+
+std::string replace(std::string const& s, char from, char to) {
+	auto result = s;
+
+	return replace(result, from, to);
+}
+
+std::string& replace(std::string& s, char from, char to) {
```
Baklap4, line 80: Might be better to add a formatter whether to ignore case or not. Now this won't replace things: `replace('a', 'A')` and it's not clear why.

game/util.cc
```diff
@@ -32,3 +36,112 @@ std::string format(std::chrono::seconds const& unixtime, std::string const& form
...
+	return result;
+}
+
+std::string erase(std::string const& s, std::string const& toErase) {
+	auto result = s;
+
+	return erase(result, toErase);
+}
+
+std::string erase(std::string& s, std::string const& toErase) {
```
Baklap4, line 65: Might be better to add  a formatter (see below)

game/util.cc
```diff
@@ -32,3 +36,112 @@ std::string format(std::chrono::seconds const& unixtime, std::string const& form
...
+std::string toLower(std::string const& s) {
+	//return boost::algorithm::to_lower_copy(s);
+	auto result = s;
+
+	std::for_each(result.begin(), result.end(), [](auto& c){ c = static_cast<char>(std::tolower(c));});
+
+	return result;
+}
+
+std::string toUpper(std::string const& s) {
```
Baklap4, line 50: Might be better to add  a locale since toUpper could do something else with weird characters.

Baklap4 (changes requested):
Some minor changes to the util functions. We might have to consider ignore case and locales and right to left languages

game/util.cc
```diff
@@ -32,3 +36,112 @@ std::string format(std::chrono::seconds const& unixtime, std::string const& form
...
+	return s;
+}
+
+std::string trimLeft(std::string const& s, std::locale const& locale) {
+	auto const start = std::find_if(s.begin(), s.end(), [&locale](char c){return !std::isspace(c, locale);}) - s.begin();
+
+	return s.substr(start);
+}
+
+std::string& trimLeft(std::string& s, std::locale const& locale) {
```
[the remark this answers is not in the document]
twollgam: Good idea 👍 But I prefer front and back.

game/util.cc
twollgam: Don't know if I changed it as you expected. Please check that.

Baklap4 (approved):
Looking good to get rid of Boost!

Let's get this tested, but codewise looking good!

BWagener:
Tested on windows, looks good to me.

Baklap4:
Pipeline fails and conflicts

Baklap4 (approved):

PR #834, opened 2022-10-10
Check thread->joinable() before calling join()

Lord-Kamina:
Especially for the ones being held in smart pointers, not doing this can occasionally trigger exceptions.

### More


Baklap4 (commented):
The code itself seems fine, but to my understanding, `thread->join()` is actually safe to call right? It only isn't safe to call when we wanna join threads when called from another thread AKA  a fork?

What exceptions were you getting before this?

Baklap4 (approved):

PR #837, opened 2022-11-12
macOS arm64 build

taynaron:
### What does this PR do?

Allows macOS arm64 (M1, M2, etc) builds of Performous.

### Closes Issue(s)

None I am aware of.

### Motivation

I want to build Performous myself so I can work on it.

### More

### Additional Notes

This is meant to start a discussion of how folks would like to integrate the macOS arm64 build into their build pipeline.I don't expect this to be merged without discussion and changes.


Lord-Kamina:
This can't be merged as is, because at the very least it will break intel builds. That said, it's good to know it's essentially a few very straight forward changes to search paths (Odds are macports wouldn't require even those).

Anyway, I have been planning for a bit now to remake the macOS bundling script using some sort of cli tool helper, to make it less unwieldy (been eyeing docopts, mostly). I'll probably integrate these changes when I get around to that.

taynaron:
Absolutely, I didn't intend for this to be immediately mergeable. If you want, we could do a 'least effort' merge - just fix the two arm64 locations to be hardware/flag dependent - and think about the long-term solution later.

Baklap4:
Would also be nice to include this within the github workflows, but sadly it's not supported yet (at least not the managed versions) it is available on self-hosted ones: https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware/

taynaron:
new checkin should be safe to check in. Uses the host architecture to determine which to build - not ideal if someone wanted to, say, build a universal build. That said, it's an improvement, and has no downsides that master doesn't also have AFAICT.

osx-utils/performous-app-build.sh
```diff
@@ -331,7 +331,7 @@ function main {
 			  -DCMAKE_CXX_COMPILER="${CXXPATH}" \
 			  -DCMAKE_EXPORT_COMPILE_COMMANDS="ON" \
 			  -DCMAKE_CXX_FLAGS="-Wall -Wextra" \
-			  -DCMAKE_OSX_ARCHITECTURES="x86_64" \
+			  -DCMAKE_OSX_ARCHITECTURES=${CMAKE_HOST_SYSTEM_PROCESSOR} \
```
Lord-Kamina, line 334: This is actually incorrect. At this point we're in bash, so you 're passing in an empty string.
OTOH, I don't really think that argument is necessary at all nowadays, so perhaps you could remove it entirely.

Lord-Kamina (changes requested):

osx-utils/performous-app-build.sh
taynaron: Oh dang, I guess I misunderstood what was happening. That said, you're right that removing the flag entirely works too. I've checked that in.

taynaron: I've also added you as a collaborator on my fork so you can mess with my branch if you like.

Lord-Kamina (approved):

Baklap4 (approved):

Issue #838, opened 2022-11-12
XML Types clash crashes Performous on start


taynaron:
### Performous & operating system versions

Performous: 1.2.0
Operating system: macOS 13 (arm64)

### What is the problem?

Performous crashes on an arm64 mac with:
<img width="372" alt="Screenshot 2022-11-11 at 8 39 14 PM" src="https://user-images.githubusercontent.com/1496984/201488612-0d3d3546-c255-484e-98e0-339db472b40f.png">
infolog has:
`core/error: /Users/taynaron/Library/Preferences/Performous/config.xml:3: Error while reading entry: Entry type mismatch: audio/backend: schema type = int, config type = uint
More details might be available in "/Users/taynaron/Library/Caches/Performous/in`

### Steps to reproduce

1. have an existing preferences file (from x86_64 build)
2. run arm64 Performous
3. see above error

### Log(s) & preferences file(s)

logger/notice: Logging all notices, warnings and errors. Log file: /Users/x/Library/Caches/Performous/infolog.txt
core/notice: Performous 1.2.0 starting...
  Internationalization: Enabled
  MIDI Hardware I/O:    Enabled
  Webcam support:       Disabled
core/error: /Users/x/Library/Preferences/Performous/config.xml:3: Error while reading entry: Entry type mismatch: audio/backend: schema type = int, config type = uint
More details might be available in "/Users/x/Library/Caches/Performous/infolog.txt".
If you think this is a bug in Performous, please report it at 
  https://github.com/performous/performous/issues

PR #840, opened 2022-11-14
fix/uint config mismatch

Lord-Kamina:
This was considered in the original implementation of the unsigned configitems. It was later removed in (presumably) a merge error.

I’ve also removed a duplicate function, probably due to the same thing.

### Closes Issue(s)

#838 

### Motivation
QA

### More

### DIFF
--- a/game/configuration.cc
+++ b/game/configuration.cc
@@ -108,7 +108,11 @@ void ConfigItemXMLLoader::update(ConfigItem& item, xmlpp::Element& elem, unsigne
 			item.setLongDescription(getText(elem, "long"));
 		}
 		else {
-			const auto type = getAttribute(elem, "type");
+			std::string type{getAttribute(elem, "type")};
+			if (item.getType() == "uint" && type == "int") {
+				type = "uint"; // Convert old config values.
+				std::clog << "configuration/info: Converting config value: " + getAttribute(elem, "name") + ", to type uint." << std::endl;
+			}
 			if (!type.empty() && type != item.getType())
 				throw std::runtime_error("Entry type mismatch: " + getAttribute(elem, "name") + ": schema type = " + item.getType() + ", config type = " + type);
 		}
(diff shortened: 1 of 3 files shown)


game/configuration.cc
```diff
@@ -108,7 +108,11 @@ void ConfigItemXMLLoader::update(ConfigItem& item, xmlpp::Element& elem, unsigne
 			item.setLongDescription(getText(elem, "long"));
 		}
 		else {
-			const auto type = getAttribute(elem, "type");
+			std::string type{getAttribute(elem, "type")};
+			if (item.getType() == "uint" && type == "int") {
+				type = "uint"; // Convert old config values.
```
Baklap4, line 113: Let's double check if this is the only type conversion or if we have some more.
I see we do define: `uint`, `int`,  and `float`

Lord-Kamina: Yes but I don't think I changed any other types. Checking the history of schema.xml should do the trick either way

Baklap4 (approved):
LGTM, @taynaron can you try out this fix :)?

Lord-Kamina:
> I'd love to! Can someone please approve and merge https://github.com/performous/performous/pull/837 so I don't have to keep manually applying my changes there to other branches?

Not for now. I'm working on rewriting that whole script, and plan to add better support for it then, but the PR as is (or at least, last I checked earlier today) would break it for everyone else.

If you don't mind me commiting to your branch, I could probably make a temporary stopgap to fix it in line with the current style of the script.

Baklap4:
> I'd love to! Can someone please approve and merge https://github.com/performous/performous/pull/837 so I don't have to keep manually applying my changes there to other branches?

For now it might be easier to make a local branch (which you already have) to contain the changes of #837 and when you work branch of from that branch. Then when you want to create the PR rebase it on master so the commit(s) from #837 aren't there any more.
If you then have to rework the other feature (in this case #840) you locally rebase with #837 (`git rebase -i 837`) and when you're done rebase it again with master `git rebase -i master`. The rebase shouldn't give any conflicts in most of the other code you make. Imho this is easier than constantly cherry-picking the commit and removing it again

taynaron:
I'm all for you committing to my branch if it'll help move things forward :)

taynaron:
I have tested, and the arm version now can run with prefs created on x86_64. I don't have the setup to allow me to test the converse, but I see no reason it wouldn't work.

taynaron:
@Baklap4 Tested, LGTM!

Issue #731, opened 2022-04-23
Ubuntu 1804 EOL cleanup


Baklap4:
## Cleanup

Ubuntu will hit EOL at the end of april. Same goes for Debian later at 30th of june.
We're currently still supporting 1804 but this has proven to become rater difficult as of lately. Think of troubles with nlohmanjson, cmake and maybe some other stuff.

This issue will be used to determine if and what we want to change/upgrade since there are new nice features on the way from dependencies we're using.

So feel free to share ideas and things so we can implement them at a later time.

### TODO

- [ ] Something

Baklap4:
Docker has been moved to the main repository.
Next step would be to phase out Debian 10 and Ubuntu 1804

PR #842, opened 2022-11-21
Migrate docker repo to main repo

ooshlablu:
### What does this PR do?
This PR migrates the performous-docker repo into the main one so changes can be made to the containers and tested within a single PR instead of having to make changes to the container repo and then here.

### Closes Issue(s)
https://github.com/performous/performous/issues/731
https://github.com/performous/performous-docker/issues/15

### Notes
https://github.com/orgs/performous/packages/container/package/deps must be deleted before this is merged. Currently, that container registry is owned by https://github.com/performous/performous-docker.
The `Required Checks` for this repo will also need to be updated to remove the old Linux ones and add the new ones.


.github/workflows/build_and_release.yml
```diff
@@ -73,216 +72,17 @@ jobs:
...
-  # https://github.com/performous/performous-docker since
-  # github actions provides only Ubuntu
+  # Pull in the Linux build workflow
   Linux_Packages:
-    # The upload step needs the release to be created first so it
-    # can access the upload_url output. This step also needs the output from
-    # the version dteremination job to name things correctly.
-    name: Create Linux packages
+    name: Build the Linux packages
+    uses: ./.github/workflows/linux.yml
```
Baklap4, line 78: Can we also do this for windows/osx? this greatly reduces the content for the actual pipeline

Baklap4 (changes requested):
Lets combine `run_ctest.sh` and `run_gtest.sh` with a switch to use one of both and auto select the one installed (if any at all)

.github/workflows/build_and_release.yml
ooshlablu: Yeah, my thought was to do just Linux in this PR and then do the rest in another PR to make my changes more visible. I think doing all at once in this PR would overwhelm those reviewing that don't have as in-depth knowledge about the pipeline that we do. Also, I know that @Lord-Kamina is making a new script/method for the MacOS builds, so I wanted to give him the opportunity to make that change :-)

ooshlablu:
> Lets combine `run_ctest.sh` and `run_gtest.sh` with a switch to use one of both and auto select the one installed (if any at all)

@Baklap4 I moved all the testing stuff into 1 script with switches to trigger either only `ctest` or `gtest`. I don't think that the script should auto detect if a testing suite is installed or not because I feel like calling the script has the expectation that the stuff is installed, and if it's not there then it's a failure.
I put in a comment around those if statements to detect the OS. When Debian 10 is retired those can go away, and we really don't have to do any detection.

docker/run_tests.sh
```diff
@@ -0,0 +1,70 @@
...
+    "c")
+      RUN_GTEST=false;;
+    "g")
+      RUN_CTEST=false;;
+    "D")
+      CLONE_DIRECTORY=${OPTARG};;
+    "t")
+      GTEST_TESTING_DIRECTORY=${OPTARG};;
+    "T")
+      CTEST_TESTING_DIRECTORY=${OPTARG};;
```
Baklap4, line 42: We might wanna switch up these options since it's kind of a bad practice to have capitalized letters do a different thing.

What might even be better is to have one "testing" directory which gets used by both ctest/gtest so we end up with one option: `TESTING_DIRECTORY`

Baklap4 (changes requested):

docker/run_tests.sh
ooshlablu: Ok, the options are switched. I agree with the testing for both being in the same place, but not sure I'd be able to do it properly on my own. I think we should make an issue, and enlist the help of @twollgam who originally made the testing stuff :-)

Baklap4 (approved):
LGTM, All issues have been addressed, let's get this merge train going!

twollgam (commented):
LGTM

PR #843, opened 2022-12-05
Rewrite macOS packaging script

Lord-Kamina:
### What does this PR do?

1) Fixes some errors in our CMake scripts and makes use of some standard CMake variables, as per the `GNUInstallDirs` module. This should largely not affect anybody.
2) Merges most of the packaging logic into CMake itself, making for a less hacky process. I've also taken the opportunity to include the version info in the Copyright field for the `.app` bundle, since that's essentially just free text.
3) Fixes a minor error with the macOS package assets.
3) Completely rewrites the packaging script in Python3, using [docopt-ng ](https://github.com/jazzband/docopt-ng), which also allows us to provide a nice help message to document how to use the command.

There are some caveats here:

* We still depend on dylibbundler.
* We now depend, *solely for producing macOS packages* on Python3 (which is not by default shipped with macOS as Python 2.7 once was, but is instead usually installed as part of the macOS Developer CLI tools, which are required anyway). We also have a dependency on a couple python modules, which I have bundled here as a python virtual-env (hence the ridiculous amount of changed files). This replaces the current dependency on node and npm, though.
* Doing this PR, I've noticed Apple have deprecated the command usually used to embed SLA in installer packages. This will become a problem eventually, so we should begin discussing what to do. If we require the license be shown to the user, maybe we could do it via an `SDL_ShowSimpleMessageBox` on first launch.
* @Tronic ~~I know you'll hate me for embedding several hundreds of files worth of python shit, but I sincerely believe the improved code and usability are worth it and will make life easier for everyone down the line when it comes to producing macOS releases.~~ Disregard me being dumb. I can have my cake and eat it too with a requirement file. Please let me know whether it works for you.

### More


Tronic:
> several hundreds of files worth of python shit

If it replaces complex shell script, ideally all shell script, I am all for it. I'll give this a test run but if there are problems, I won't be able to spend more time debugging at this time.

Lord-Kamina:
> > several hundreds of files worth of python shit
> 
> If it replaces complex shell script, ideally all shell script, I am all for it.

I'm still ironing a couple final kinks in the PR itself but the script should be pretty much fully functional. Most likely you will still need to use the script, in order to deal with some weird things in where Homebrew places certain files but with this, it's theoretically possible to produce a functional build without using the script at all.

>  I'll give this a test run but if there are problems, I won't be able to spend more time debugging at this time.

Probably not much more debugging on your part will be required, since the CI now uses Homebrew.

Lord-Kamina:
> Looks like I am once again missing a lot of development packages (not installed or not found, even in traditional build). I'll try to get back to this at a (much) later time when I can risk getting my dev system messed up by those installs.

When you eventually have time, you should really consider replicating my setup: install macports to an external disk but mount said disk under a folder on your path (or even manually adding it to path when necessary, for extra safety), configured directly in `/etc/fstab`, then you could just plug it and unplug it when necessary and completely isolate it from your regular work environment.

Tronic:
I think pretty much everyone is on Homebrew already. Maybe a sandboxed install of that? Do you know how to make one without VMs?

Lord-Kamina:
> I think pretty much everyone is on Homebrew already. 

Not really, I think this is largely an ideological divide that will be around forever. 

> Maybe a sandboxed install of that? Do you know how to make one without VMs or external disks?

For homebrew, I'm not sure it's possible. For macports it should be relatively straightforward because it's by design made to keep itself separate from the OS and tucked away in it's own folder tree. Possibly evwn a standard install and then taking it out of the PATH would do the trick. I suggested the external drive only because it's effectively a physical on/off button.

taynaron:
I'd agree with @Tronic that brew largely won the user war ~5 years ago. Also, @Lord-Kamina homebrew on arm64 installs to a homebrew-specific directory, so if "you can remove it from $PATH" == "sandboxed" then macports and homebrew are at feature parity there.

.github/workflows/build_and_release.yml
```diff
@@ -396,32 +403,31 @@ jobs:
             aubio fftw glm nlohmann-json fmt
            brew link --force --overwrite ffmpeg@4
 
+      - name: Setup Python
+        id: setup_python
+        uses: actions/setup-python@v4 
+        with:
+            python-version: 'pypy3.9'
```
taynaron, line 410: if we're using python for the first time, why lock on a python version over 2 years old?

.gitignore
```diff
@@ -1,15 +1,17 @@
 [bB]uild/*
+[bB]uild.xcode
 CMakeLists.txt.user*
 *~
 *.mo
 dylibbundler.log
 performous-app-build.sh
```
taynaron, line 7: is this correct? This is set to ignore but the script is checked in.

osx-utils/macos-bundler.py
```diff
@@ -0,0 +1,342 @@
...
+	-f --flat-output  Put output directly in the output folder, without versioned folders.
+	-j <n>, --jobs <n>  Argument passed to make, specifying the max number of jobs to run. Defaults to the output of os.cpu_count() [default: {str(cpu_count())}]
+	-k --no-clean  Don't erase tree structure of the .app before beginning work.
+	-n --no-regenerate  Don't regenerate the CMake build system.
+	-v <version>, --package-version <version>  Sets the version on the Info.plist of the created .app. By default, it's calculated according to the latest stable version and hash of the latest commit.
+	-x --xcode-project  Generate an XCode project suitable for debugging. Note: to use this project, it's necessary to build the install target from XCode and then the performous target (or, alternatively, mirror the settings for the performous scheme on the install target)
+	--script-debug  Print the resolved arguments and options passed to this utility.
+
+Environment:
+	--arch <architecture>...  Target architecture names passed to the compiler. [default: x86_64]
```
taynaron, line 209: would it be reasonable to make --arch's default the system architecture?

osx-utils/macos-bundler.py
```diff
@@ -0,0 +1,342 @@
...
+	--cc <path/to/compiler>  Change C compiler [default: /usr/bin/clang]
+	--cxx <path/to/compiler>  Change C compiler [default: /usr/bin/clang++]
+	-p <prefix>, --prefix <prefix>  Set prefix path for searching of libraries and headers. By default, the tool tries to detect whether MacPorts or Homebrew are installed and the prefix is set accordingly. Note: If both are installed, you can choose.
+	(--prefer-macports | --prefer-homebrew)  Prefer either MacPorts or Homebrew.
+	-s <path>, --source <path>  Path to the Performous source. Defaults to ../
+	-o <path>, --output <path>  Path where the .app will be built. Defaults to <performous-source>/osx-utils/out[/xcode]
+	-t <target>, --target <target>  macOS target version. Defaults to the currently running version, as reported by platform.mac_ver() [default: {str(float('.'.join(platform.mac_ver()[0].split('.')[:2])))}]"""
+
+if __name__ == "__main__":
+	arguments = docopt(docstring = usageHelp, default_help = False)
```
taynaron, line 219: why use docopt over the python-included argparse?

taynaron (commented):
compiles on arm64. Comments/questions/concerns added inline.

.github/workflows/build_and_release.yml
Lord-Kamina: It's the one that comes bundled with the latest xcode command line tools and, do we honestly need a newer one?

.gitignore
Lord-Kamina: It's not the script. I initially made a vlrtualenv inside that folder.

osx-utils/macos-bundler.py
Lord-Kamina: Actually yes, come to think of it.

osx-utils/macos-bundler.py
Lord-Kamina: Because it gives us a great user experience (my main motivation in doing this) and makes it easier change down the line.

.gitignore
Lord-Kamina: Not but I think it won't hurt to keep it for a bit.

osx-utils/macos-bundler.py
Tronic: Python's argparse is quite bad actually. Needs quite a bit of code and still has severe limitations (requiring monkeypatching its code to get to good parsers, a lot of trickery to get well formatted help). I have no experience with docopt yet but I do plan to try it on my next project that needs a simple parser (thanks @Lord-Kamina).

Lord-Kamina: If you do (and want the python version) make sure to use `docopt-ng` because the original is essentially abandoned.

.github/workflows/linux.yml
```diff
@@ -1,7 +1,7 @@
 name: Build Linux Packages
 
 on:
-  # Run on a schedule to get monthly updates
+  # Run on a schedule to get monthly updates #
```
ooshlablu, line 4: This was probably a mistake :-)

ooshlablu (approved):
LGTM. The only qualm I have is about the changes to the commenter, but as I mentioned we'll run it and see how it goes. If it comes to it, we can just revert those changes.

.github/workflows/linux.yml
Lord-Kamina: Yes and no. It was on purpose but I did it so the containers would be rebuilt.

ooshlablu: Oh, I meant the extra `#`

ooshlablu (approved):
LGTM

PR #757, opened 2022-05-07
Player screen

twollgam:
### What does this PR do?

Add a new screen for player management.

### Closes Issue(s)

#574 

### Motivation

The current player management is very basic. With this PR one can manage players ingame.

PR #846, opened 2022-12-13
Refactoring of text rendering and text measurement´

twollgam:
### What does this PR do?

- Refactor text rendering. SRP: The rendering of text to a texture is moved to a new class TextRenderer.
- Add a measure function to class TextRenderer. 
- Add class Size as helper class. 

### Closes Issue(s)

None

### Motivation

For the PR #757  (Player Screen) drawing a cursor for text input is needed. One needs the position of text per character. This pr adds this possibilities.


twollgam:
@ooshlablu if you test this one have a look at text rendering. It should not make a difference to master..

ooshlablu:
I tested this with the deb and AppImage on Ubuntu 20.04 + 22.04. Everything looks fine to me, didn't notice any differences in gameplay, which should be the point. :+1:

Baklap4:
@twollgam will this also fix the issue that text overflows in the X direction (out of screen) during singing?

game/graphic/Size.cc
```diff
@@ -0,0 +1,24 @@
+#include "Size.hh"
```
Baklap4, line 1: Personally i'm fine with PascalCasing filenames and classes/interfaces and I actually prefer it, but we haven't done that before. We used snake_casing up until now. Not sure what others think of that @Lord-Kamina @ooshlablu

game/opengl_text.cc
```diff
@@ -212,11 +135,13 @@ SvgTxtThemeSimple::SvgTxtThemeSimple(fs::path const& themeFile, float factor) :
...
-		m_cache_text = _text;
-		m_text.text = _text;
-		m_opengl_text = std::make_unique<OpenGLText>(m_text, m_factor);
+void SvgTxtThemeSimple::render(std::string const& text) {
+	if (!m_opengl_text.get() || m_cache_text != text) {
+		m_cache_text = text;
+
+		auto renderer = TextRenderer();
+
+		m_opengl_text = std::make_unique<OpenGLText>(renderer.render(text, m_text, m_factor));
```
Baklap4, line 144: Do we need instance level access or can we make the renderer go static? We're now creating TextRenderer objects for every render call which isn't necessairy a good thing i think?

Baklap4 (changes requested):

ooshlablu:
> @twollgam will this also fix the issue that text overflows in the X direction (out of screen) during singing? Or can this be used to fix that specific problem?

It doesn't in my tests, but I believe this is the scaffolding necessary to make a proper fix for that issue.

game/graphic/TextRenderer.cc
```diff
@@ -0,0 +1,140 @@
...
+
+	PangoStyle parseStyle(std::string const& fontstyle) {
+		if (fontstyle == "normal") return PANGO_STYLE_NORMAL;
+		if (fontstyle == "italic") return PANGO_STYLE_ITALIC;
+		if (fontstyle == "oblique") return PANGO_STYLE_OBLIQUE;
+		throw std::logic_error(fontstyle + ": Unknown font style (opengl_text.cc)");
+	}
+    
+    void alignFactor(float& factor) {
+        factor *= 2.0f;  // HACK to improve text quality without affecting compatibility with old versions
```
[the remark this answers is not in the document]
twollgam: I did not test it, but I think it is necessary to increase the visual quality of the rendered text.

game/graphic/Size.cc
twollgam: That is a good point. I like if file name is exact the class name. Actual some classes are "hidden" behind file names, but may be file names follow no scheme ... (f.e. the screen classes)

game/opengl_text.cc
twollgam: I thought about it, but making the functions static seems to me like premature optimiszation. If we need (or want to have)  an alternative renderer, which could be implemented by interface, the static way is counterproductive, That leads me keep is non static. As alternative:
```
static TextRenderer renderer;
```
Then the consumer avoids frequent initialization.

game/graphic/TextRenderer.cc
```diff
@@ -0,0 +1,140 @@
...
+#include <pango/pangocairo.h>
+
+#include <memory>
+
+namespace {
+	PangoAlignment parseAlignment(std::string const& fontalign) {
+		if (fontalign == "start") return PANGO_ALIGN_LEFT;
+		if (fontalign == "center" || fontalign == "middle") return PANGO_ALIGN_CENTER;
+		if (fontalign == "end") return PANGO_ALIGN_RIGHT;
+		throw std::logic_error(fontalign + ": Unknown font alignment (opengl_text.cc)");
```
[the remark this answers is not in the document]
twollgam: @Baklap4 shouldn't we use `std::invalid_argument` here?

twollgam:
> > @twollgam will this also fix the issue that text overflows in the X direction (out of screen) during singing? Or can this be used to fix that specific problem?
> 
> It doesn't in my tests, but I believe this is the scaffolding necessary to make a proper fix for that issue.

@Baklap4  That is right. This pr is only a refactoring and it adds some new functions, which are needed for the PlayerScreen-PR. But there are no changes for user experience.

game/graphic/TextRenderer.cc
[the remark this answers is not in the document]
Baklap4: It was in there before. Let's keep it that way for now

game/graphic/Size.cc
Baklap4: Personally i'm a fan of having one file per class/header/test

That might be another refactor as you say some classes are hidden which isn't that nice

game/opengl_text.cc
Baklap4: Agree to this approach (consumer of the class defining a static renderer)
We wouldn't want the methods itself to be static like you said, that's counterproductive

Lord-Kamina:
> @twollgam will this also fix the issue that text overflows in the X direction (out of screen) during singing? Or can this be used to fix that specific problem?

No. The issue here is we have some hardcoded values and hacks. We haven't implemented text alignment properly (all our text is left or center aligned, can't remember., and we position it with OpenGL) nor any semblance of bounds checking.

The solution is to properly implement text alignment and wrapping (and ideally detecting how many lines fit on a given screen), but this also necessitates fixing the broken Dimensions class.

twollgam:
> > @twollgam will this also fix the issue that text overflows in the X direction (out of screen) during singing? Or can this be used to fix that specific problem?
> 
> No. The issue here is we have some hardcoded values and hacks. We haven't implemented text alignment properly (all our text is left or center aligned, can't remember., and we position it with OpenGL) nor any semblance of bounds checking.
> 
> The solution is to properly implement text alignment and wrapping (and ideally detecting how many lines fit on a given screen), but this also necessitates fixing the broken Dimensions class.

Measure text dimensions could be the base for implementing text wrapping. But it should wait for the fix of the Dimensions class.

game/graphic/Size.cc
twollgam: The current naming is not consistent. Always lower case, sometimes with _, but not always snake. 

My opinion: I make filenames lowercase in this pr and would love if we can go the way @Lord-Kamina  suggested: extra pr switching to PascalCase :+1: :+1:

ooshlablu: :+1: @twollgam

twollgam: done

game/opengl_text.cc
```diff
@@ -212,11 +135,13 @@ SvgTxtThemeSimple::SvgTxtThemeSimple(fs::path const& themeFile, float factor) :
...
-void SvgTxtThemeSimple::render(std::string _text) {
-	if (!m_opengl_text.get() || m_cache_text != _text) {
-		m_cache_text = _text;
-		m_text.text = _text;
-		m_opengl_text = std::make_unique<OpenGLText>(m_text, m_factor);
+void SvgTxtThemeSimple::render(std::string const& text) {
+	if (!m_opengl_text.get() || m_cache_text != text) {
+		m_cache_text = text;
+
+		static TextRenderer renderer;
```
Baklap4, line 142: Would it make sense to make this a member field within opengl_text.hh?

game/opengl_text.cc
```diff
@@ -246,22 +171,25 @@ void SvgTxtTheme::draw(Window& window, std::string _text) {
 
 void SvgTxtTheme::draw(Window& window, std::vector<TZoomText>& _text, bool lyrics) {
 	std::string tmp;
-	for (auto& zt: _text) { tmp += zt.string; }
+
+	for (auto& zt: _text)
+		tmp += zt.string;
```
Baklap4, line 176: Please add the braces even though its a single statement in the forloop. This way i don't have to scan the code for more indentations (that's the reason it's been single lined i guess) to see where the loop ends.

game/opengl_text.cc
```diff
@@ -246,22 +171,25 @@ void SvgTxtTheme::draw(Window& window, std::string _text) {
...
+
 	if (m_opengl_text.size() != _text.size() || m_cache_text != tmp) {
 		m_cache_text = tmp;
 		m_opengl_text.clear();
+		auto renderer = TextRenderer();
 		for (auto& zt: _text) {
-			m_text.text = zt.string;
-			auto openGlPtr = std::unique_ptr<OpenGLText>(std::make_unique<OpenGLText>(m_text, m_factor));
+			auto openGlPtr = std::make_unique<OpenGLText>(renderer.render(zt.string, m_textstyle, m_factor));
```
Baklap4, line 183: we're making a lot of unique smart pointers here. Shouldn't we make the instance first, fill it and push it back to the vector and then clearing it and repeating the process of filling/inserting in the vector/clear?

Baklap4 (changes requested):
still a couple of things left, but these are things i'm not so sure about, so if its all nonsense im talking just make me approve the PR :P

game/opengl_text.cc
twollgam: I think that this would actually have no benefit.

game/opengl_text.cc
twollgam: Why? I didn't get the point ... sorry

Baklap4 (approved):

PR #847, opened 2022-12-20
Fix whammy and godmode mapping for the X-Plorer

ooshlablu:
### What does this PR do?

Fixes the mapping for the X-Plorer guitar, godmode and whammy were swapped.

### Motivation
Noticed is while playing recently. I thought there was a bug in "Star Power" and whammy, or it wasn't implemented or something.


Baklap4 (approved):
I don't own an X-Plorer Guitar, but you do so this is probably fine :)

PR #849, opened 2022-12-28
Seperate the workflows into different files to get a better overview.

### LINKED ISSUE #848: Move the AppImage/MacOS/Windows Actions worflows into separate files like the Linux Packages
### Suggestion

In https://github.com/performous/performous/pull/842, the Linux packaging workflow was broken out from the main `build_and_test.yml` to its own workflow file, `linux.yml` which is pulled in from the main workflow. The rest of the packaging workflows should be separated in a similar way to make life easier for contributors that don't have a deep knowledge of GitHub actions, since everything will be concentrated in relevant files rather that all dumped into the main workflow.

### Use case

_No response_

The Linux workflow was broken out from the main workflow like this: https://github.com/performous/performous/blob/master/.github/workflows/build_and_release.yml#L76-L84, and moved to a separate file: https://github.com/performous/performous/blob/master/.github/workflows/linux.yml

Baklap4:
### What does this PR do?

* Cleans up the workflow file by having it split up per environment

### Closes Issue(s)

Closes #848

### Notes
- [ ] Before this PR can get merged we need to change the required actions. This can be done after approval


.github/workflows/appimage.yml
```diff
@@ -0,0 +1,96 @@
+name: Build AppImage Packages
+
+on:
+  # Run on a schedule to get monthly updates
+  schedule:
+    - cron: "0 0 28 * *"
```
ooshlablu, line 6: I just want to make sure that this is being used as intended. The reason I added the `cron` statement to the Linux workflow was to make sure that the underlying containers get refreshed with updated deps each month so we don't have to do a security dance if a dep gets pwnd. This could be useful for making sure that the README has packages available with the 90-day default expiry time.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,101 @@
+name: Build MacOS Package
+
+on:
+  # Run on a schedule to get monthly updates
+  schedule:
+    - cron: "0 0 28 * *"
```
ooshlablu, line 6: Same statement as above.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,220 @@
+name: Build Windows Packages
+
+on:
+  # Run on a schedule to get monthly updates
+  schedule:
+    - cron: "0 0 28 * *"
```
ooshlablu, line 6: Same comment as above.

ooshlablu (approved):
LGTM. Just make sure those cron statements are necessary/intended, which is not a blocker.

PR #850, opened 2022-12-30
Fix release url for dependent workflows

ooshlablu:
### What does this PR do?

Fixes the upload to release stages with the required `inputs` rather than the old `needs`.

### Motivation

Would have broke if we tried doing a release, but would have been fine for everything else.


Baklap4 (approved):
lgtm

PR #853, opened 2023-01-04
fix highscore not showing and track with no vocal

Darkmagister:
### What does this PR do?

adding again the screen at the end of a song for the high score with instrument not only with vocal

### Motivation

i usually use with with instrument and almost never for sing, so it was frustrating not having any highscore

### Additional Notes

i only tested it with few songs with and without vocal and it seems to be working
i had to edit the getVocalTrack function otherwise it will crash on track without vocal

updated to new master, #803

### DIFF
--- a/game/screen_sing.cc
+++ b/game/screen_sing.cc
@@ -266,7 +266,7 @@ void ScreenSing::activateNextScreen()
 	}
 
 	// Score window visible -> Enter quits to Players Screen
-	if(!config["game/karaoke_mode"].ui() && !m_song->hasDance() &&!m_song->hasDrums() &&!m_song->hasGuitars()) {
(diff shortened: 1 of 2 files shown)


ooshlablu:
It seems to be working as intended:
```xml
  <hiscores>
    <hiscore playerid="0" songid="2467" track="Guitar - Medium" level="0">8132</hiscore>
    <hiscore playerid="0" songid="2467" track="Vocals" level="0">6009</hiscore>
  </hiscores>
```
I've tested with the debs and the AppImage for Ubuntu 20.04 & 22.04, and have not noticed any detrimental effects to game play or existing functionality.
LGTM

ooshlablu (approved):
LGTM

PR #854, opened 2023-01-24
Try a specific revision for the MSVC build

ooshlablu:
### What does this PR do?
Locks the `vcpkg` version to `master` for now at least. There was an issue with `libiconv` where the deps disappeared from the download mirror: https://github.com/performous/performous/actions/runs/4002705248/jobs/6870182159

Unfortunately, that issue popped up and was cleaned up last week, which was one week after the latest release. This will likely be resolved in the near-ish future when a new `vcpkg` release is made.

### Closes Issue(s)
No linked issue, but our master branch is currently broken since some of the deps `404` when trying to build on the master branch.

### Motivation

I noticed that the master builds were failing after https://github.com/performous/performous/pull/846 was merged.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -46,6 +46,7 @@ jobs:
           triplet: x64-windows-release
           extra-args: --clean-after-build
           cache-key: win64-vcpkg
+          revision: 4715616818c2ac225a13dd05e31c3bbbf904ec65
 
       - name: Run cmake to configure the project and build it
         env:


Lord-Kamina:
The default is not master, but rather the latest release as determined by voodoo.

Baklap4 (changes requested):
Instead of targeting master, which will cause a cache invalidation everytime they push a change, target a certain commit. (which for now can be the latest master hash)

if a commit hash is not possible we should make it possible within the action of @Lord-Kamina

Baklap4 (approved):

PR #856, opened 2023-01-31
Increase transparency of wave.svg provided by user Neas

twollgam:
### What does this PR do?

User Neas mentioned:

> Currently it seems that the leading player has nearly no transparency on their pitch wave so you cant really see others. Is there an option somewhere to make it more transparent? 

He increased transparency of the wave.svg which looks better. This pr only contains the modified file.

### Closes Issue(s)

None.


Lord-Kamina:
A while ago, I wanted to PR a rework of several graphics, including turning a couple things we currently use raster for into SVGs. I never did submit it, but maybe I should.

ooshlablu:
I tested this with the deb and AppImage on Ubuntu 20.04 && 22.04, and it mostly looks good. This background does make it hard to see when it is used however: https://github.com/performous/performous/blob/master/data/backgrounds/blue_band.svg. Maybe that needs to be toned down as well. IT works fine with all of the other backgrounds, and looks ok for all the Album covers I have tested, though I do suspect an album cover with a lot of blue will cause issues.

twollgam:
Any suggestion? @ooshlablu why has the original svg no the same problem? Isn't it blue on blue?

Baklap4:
Hmmm transparency is nice, but when a darker blue (background) is interfering with a lighter blue (microphone) that can indeed cause an issue. Personally i think most of the users will have custom background and or video's and thus won't notice this, but the argument still stands. I'm alright with the upcoming change :)

Baklap4:
@Lord-Kamina any chance you can add those commit in a pr so we can review those seperately?

Baklap4 (approved):

ooshlablu:
@twollgam Here are some good examples of the interference with that background:
Screenshot from 2023-02-01 18-42-50
Screenshot from 2023-02-01 18-42-52

It is easier to pick up where the vocals are with the others:
Screenshot from 2023-02-01 18-43-39

Would it be possible to lighten up the blue in `blue_band.svg`? I think that would help.

Everything else looks fine to me.

PR #835, opened 2022-11-08
Fix sigsegv in drawCovers().

Lord-Kamina:
### What does this PR do?

Fixes a potential `nullptr` dereference in `drawCovers`.

@ooshlablu noticed a crash in screeen_songs while testing #545.

Upon investigating I reproduced it and tracked it here. No idea why it's not causing a crash on master but did on the PR.

### Motivation

QA

### More

- [ ] Added/updated documentation

PR #865, opened 2023-02-18
Split video_driver into one file per class and fix crash 

twollgam:
### What does this PR do?

Two things:
1. It contains a refectoring for video_driver.hh/cc. There are multiple classes inside the files which are splitted into one file per class.
2. Fix a crash that appears while destructing `Window` where shaders must be destructed before the gl context is gone. Cause actually that is not obvious and only mentioned in a comment, goal is to force the right destruction sequence by code.

### Closes Issue(s)

No issue opened for the bug/crash.

### Motivation

Fix a bug and avoid inserting the bug accidentally in future-

### More

- [x] Moving classes to own files
- [x] Fix bug by restructuring code
- [x] Code review
- [x] Testing

### DIFF
--- a/game/graphic/window.cc
+++ b/game/graphic/window.cc
@@ -64,8 +64,12 @@ Window::Window()
 		GLattrSetter attr_d(SDL_GL_DEPTH_SIZE, 24);
 		GLattrSetter attr_db(SDL_GL_DOUBLEBUFFER, 1);
 		Uint32 flags = SDL_WINDOW_HIDDEN | SDL_WINDOW_RESIZABLE | SDL_WINDOW_OPENGL;
-		if (config["graphic/highdpi"].b()) { flags |= SDL_WINDOW_ALLOW_HIGHDPI; }
-		else { SDL_SetHintWithPriority("SDL_HINT_VIDEO_HIGHDPI_DISABLED", "1", SDL_HINT_OVERRIDE); }
+		if (config["graphic/highdpi"].b()) {
+			flags |= SDL_WINDOW_ALLOW_HIGHDPI;
+		}
+		else {
+			SDL_SetHintWithPriority("SDL_HINT_VIDEO_HIGHDPI_DISABLED", "1", SDL_HINT_OVERRIDE);
+		}
 		int width = config["graphic/window_width"].i();
 		int height = config["graphic/window_height"].i();
 		int windowPosX = config["graphic/window_pos_x"].i();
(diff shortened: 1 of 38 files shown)


ooshlablu:
@twollgam I tested this with the `deb`s and AppImage on Ubuntu 20.04 & 22.04 with vocals + guitar/bass and everything looks great. I tested at least 3 songs in-a-row with each combo to make sure there weren't any sneaky crashes that popped up, and it all worked fine. It also fixes the issue seen in this comment: <https://github.com/performous/performous/pull/835#pullrequestreview-1304218992>

LGTM!

game/graphic/window.cc
```diff
@@ -312,7 +317,11 @@ void Window::render(Game &game, std::function<void (void)> drawFunc) {
 	updateStereo(stereo ? getSeparation() : 0.0f);
 	glerror.check("setup");
 	// Can we do direct to framebuffer rendering (no FBO)?
-	if (!stereo || type == 2) { view(stereo); drawFunc(); return; }
+	if (!stereo || type == 2) {
```
Baklap4, line 320: Maybe nice to refactor `type` to an enumeration? so we have staticly easy to rename types?
If this is too hard because of the configuration not quite supporting enums then let's stick with this for now

game/guitargraph.cc
```diff
@@ -59,10 +60,10 @@ void GuitarGraph::initGuitar() {
...
 	for (auto const& elem: m_song.instrumentTracks) {
 		std::string index = elem.first;
-		if (index != TrackName::DRUMS 
-			&& index != TrackName::DRUMS_SNARE 
-			&& index != TrackName::DRUMS_CYMBALS 
-			&& index != TrackName::DRUMS_TOMS 
+		if (index != TrackName::DRUMS
+			&& index != TrackName::DRUMS_SNARE
+			&& index != TrackName::DRUMS_CYMBALS
+			&& index != TrackName::DRUMS_TOMS
```
Baklap4, line 66: `||` and `&&` go at the end of the line, not the beginning of the line.

so:
```
if (x == TrackName::Drums &&
    x == TrackName::Drums_SNARE &&
    x == TrackName::Drums_TOMS) {
}
```

game/guitargraph.cc
```diff
@@ -82,9 +83,9 @@ void GuitarGraph::initDrums() {
 	// Copy all tracks of drum type to m_instrumentTracks
 	for (auto const& elem: m_song.instrumentTracks) {
 		std::string index = elem.first;
-		if (index == TrackName::DRUMS 
-			|| index == TrackName::DRUMS_SNARE 
-			|| index == TrackName::DRUMS_CYMBALS 
+		if (index == TrackName::DRUMS
+			|| index == TrackName::DRUMS_SNARE
+			|| index == TrackName::DRUMS_CYMBALS
```
Baklap4, line 88: || and && go at the end of the line, not the beginning of the line.

so:

if (x == TrackName::Drums &&
    x == TrackName::Drums_SNARE &&
    x == TrackName::Drums_TOMS) {
}

game/screen_intro.cc
```diff
@@ -183,13 +184,13 @@ void ScreenIntro::populateMenu() {
 			for (auto const& item: submenu.items) {
 				ConfigItem& c = config[item];
 				opts.emplace_back(MenuOption(c.getShortDesc(), c.getLongDesc()));
-                                opts.back().changer(c);
+								opts.back().changer(c);
```
Baklap4, line 187: Not quite sure why this spacing is needed? (might be my browser)
Allign it with `opts`

game/screen_intro.cc
```diff
@@ -183,13 +184,13 @@ void ScreenIntro::populateMenu() {
...
 			}
 			configmain.emplace_back(MenuOption(submenu.shortDesc, submenu.longDesc, imgConfig));
-                        configmain.back().submenu(std::move(opts));
+						configmain.back().submenu(std::move(opts));
```
Baklap4, line 190: Not quite sure why this spacing is needed? (might be my browser)
Allign it with configmain

game/screen_intro.cc
```diff
@@ -183,13 +184,13 @@ void ScreenIntro::populateMenu() {
...
 		} else {
 			configmain.emplace_back(MenuOption(submenu.shortDesc, submenu.longDesc, imgConfig));
-                        configmain.back().screen(submenu.name);
+						configmain.back().screen(submenu.name);
```
Baklap4, line 193: Not quite sure why this spacing is needed? (might be my browser)
Allign it with configmain

Baklap4 (changes requested):

game/graphic/window.cc
twollgam: Tried it, looks better IMO

game/graphic/window.hh
```diff
@@ -21,6 +21,8 @@ static inline float virtH() { return float(screenH()) / screenW(); }
 
 /// handles the window
 class Window {
+	enum class Stereo3dType {RedCyan, GreenMagenta, OverUnder};
```
Baklap4, line 24: I'm not sure if it's done in other enums, but please specify the (integer) value so we're always sure which int maps to which value and we can't break it if we add a type inbetween or in front.

Baklap4 (changes requested):
Not sure if it's common practice but yea i've seen these things go wrong..

Baklap4 (approved):

PR #866, opened 2023-02-26
Move the cron trigger to the correct workflow and set it to run weekly

ooshlablu:
### What does this PR do?

Move the cron trigger to the correct workflow and set it to run weekly.

### Motivation
The old way the `cron` trigger was firing would miss the steps required to name the artifacts and, therefore, the builds would fail. The `cron` trigger was bumped up to a weekly frequency to make sure that we have a functioning cache for `vcpkg` builds.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -1,10 +1,6 @@
 name: Build AppImage Packages
 
 on:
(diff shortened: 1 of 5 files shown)


Baklap4 (approved):
LGTM

PR #867, opened 2023-02-26
Common functions library

ooshlablu:
### What does this PR do?
Move the code that figures out the name and location of the Linux packages and AppImages to a shell script that can be sourced and the function called directly in the workflows. This is to eliminate drift that may occur between common code areas that must be solved in multiple workflows, and makes the workflows simpler and cleanlier. I did the same thing for the Composer repo a few months back: https://github.com/performous/composer/pull/49

Also, the official Github actions we were using to create releases and upload artifacts to them have been deprecated, will no longer be maintained, and the actions runners are starting to complain about them because they use the old `set-output` method and node 12:
https://github.com/actions/create-release/issues/119
https://github.com/actions/upload-release-asset/issues/78

### More
**This should be merged after https://github.com/performous/performous/pull/866.**

Here's a release, created in my repo, that is using the new methods in this PR: https://github.com/ooshlablu/performous/releases/tag/3.4.5
The actions run for that was: https://github.com/ooshlablu/performous/actions/runs/4273227492/jobs/7438991455
Just like the old jobs, releases will be created as drafts and will require a member of the org to publish them before they are officially released. I published the release I linked so others who are not part of my org/repo can view it.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -31,13 +31,13 @@ jobs:
       - name: Checkout Git
         id: checkout_git
         uses: actions/checkout@v3
-        with:
-          path: performous
 
       - name: Build the AppImage
         id: build_appimage
         run: |
-          cd performous
+          # Pull in our common build functions
+          . .github/workflows/build_functions.sh
+
           PACKAGE_VERSION=${{ inputs.package_complete_version }}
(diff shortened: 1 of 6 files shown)


Baklap4 (approved):
LGTM

PR #868, opened 2023-03-15
Filter screen

twollgam:
### What does this PR do?

This PR adds a new screen to create a complex song filter. It reduces the displayed songs on song screen.

### Closes Issue(s)

None

### Motivation

Actually filtering on song screen is possible but only simple by one criteria. The filter screen adds the possibility to combine different criteria into one filter.


Baklap4:
Screenshot looks nice! i don't quite like the "dropdown" buttons, it's too bright compared to the background

Also a question: How would one clear the advanced filter? (note i haven't tested it yet)

twollgam:
Sadly size depends on the theme and we are not so far to customize that too. But there is a pr for progressiv theme configuration  (wished by @marwin89 ).

The default theme has a little bit bigger font:
image

But I think the arrows could be smaller.

twollgam:
> Also a question: How would one clear the advanced filter? (note i haven't tested it yet)

Should I add a button "Clear" or "Reset"?

Baklap4:
Maybe we can utilize the escape button while in the songbrowser. It now clears the current filtering, but it can also clear the advanced filtering if wished (or maybe using shift + escape)
Also within the songbrowser is it clear that the advanced filter is on?

twollgam:
I decided to add a reset button cause it is easier to understand for users even if the are new to performous.

twollgam:
@Baklap4 here is a new screenshot:

image

I think the button makes sense cause the user experience is better as when there will be only short cuts. 

Showing that extended filtering is "on" in the song browser is a good idea. How can we archive that? Or new pr?

twollgam:
@ooshlablu would be nice if you could test this one again.
@Baklap4  @HetorusNL  I think the pr is now feature complete and needs a code review.

game/screen_songfilter.cc
```diff
@@ -0,0 +1,313 @@
...
+
+		result.languages.emplace_back(none);
+		result.genres.emplace_back(none);
+
+		auto languages = std::set<std::string>();
+		auto genres = std::set<std::string>();
+
+		for(auto const& song : songs.getSongs()) {
+			languages.emplace(song->language);
+//			std::cout << song->genre << std::endl;
```
Baklap4: can be removed

game/songfilter.cc
```diff
@@ -0,0 +1,113 @@
...
+	if(matches(song.genre, m_genres))
+		return containsNoCase(song.genre, m_genre);
+
+	return m_genre == "Other";
+}
+
+
+LanguageFilter::LanguageFilter(std::string const& language)
+: m_language(language) {
+	//std::cout << "set language " << language << std::endl;
```
Baklap4: Commented prints `std::cout` can be removed or replaced with debug/trace log messages

game/ui/effect/sinus.cc
```diff
@@ -0,0 +1,24 @@
...
+Sinus::Sinus(float durationInSeconds, float min, float max)
+: m_durationInSeconds(durationInSeconds), m_min(min), m_max(max) {
+}
+
+void Sinus::setConsumer(std::function<void(float, EffectContext& context)> consumer) {
+	m_consumer = consumer;
+}
+
+namespace {
+	float const pi(atanf(1.f) * 4.f);
```
kt--: I'm interested to know why you do it this way, instead of the various math.h constants?

game/ui/list.cc
```diff
@@ -0,0 +1,256 @@
...
+std::any const& Item::getUserData() const {
+	return m_userData;
+}
+*/
+void Item::setUserData(std::any const& data) {
+	m_userData = data;
+}
+
+
+const size_t List::None = size_t(-1);
```
kt--: It's not clear what your intention is here.  

Obviously `size_t` is an unsigned type.  Did you mean to use `ssize_t` with the `-1`?  If you just want a size_t "MAX_INT", perhaps it's better to use something from `std::numeric_limits`.

kt-- (approved):
Generally this is all really nice code.   The only thing I'm worried about is those couple of floats that don't seem to be initialised (but maybe I missed something?).

Personally, I would make a named constant of the `size_t(-1)` that represents a "this is unassigned" value so it's only in one spot.  (Or use `ssize_t` instead).  But this is only a  minor improvement.  I'm a pedant about this sort of thing, because I once spent ages  debugging some negative enums that were compiled to large-ints and broke the logic.

PR #870, opened 2023-03-22
﻿Refactoring tight coupling between Song and SongParser

twollgam:
### What does this PR do?

The tight coupling between Song and SongParser will be broken. A very simple interface is used. 

### Closes Issue(s)

None

### Motivation

This will help in upcoming prs to create unit tests more easy.


ooshlablu:
Tested with singing on Ubuntu 20.04 with both the deb and AppImage, didn't notice any issues.

Baklap4 (changes requested):
Overall changes look good, but i think we can do better.
1. The interface is very neat
2. Create a factory which creates all parsers
3. This factory first determines what to parser it needs then uses that
4. Move the parse specific methods to their class and mark them private
5. Cleanup the Songparser header/code file (by removing all the specific parse methods) (and maybe get rid of it entirely) since we'll have specific classes for each parser and don't have a need for a base SongParser class imho
6. Write unittests for specific parsers
7. (optional) write some documentation about how to create a new parser (for example, lrc)

PR #871, opened 2023-03-22
CI: Use RelWithDebInfo, fixes logic to build tests and run them on every platform.

Lord-Kamina:
### What does this PR do?

A couple of small things and housekeeping but chiefly, build RelWithDebInfo and make tests build and run on every platform in the CI checks.

### More


docker/build_performous.sh
```diff
@@ -65,15 +65,26 @@ else
...
   EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DUSE_BOOST_REGEX=1"
 fi
 
 ## Set more cmake flags for Debian 10
 # Debian Buster has system Aubio 0.4.5, this is not enough
 # because performous requires a minimum version of 0.4.9.
-if ([ "${ID}" = "debian" ] && [ "${VERSION_ID}" = "10" ]); then
+if ([[ "${ID}" == "debian" && "${VERSION_ID}" == "10" ]]); then
   EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DSELF_BUILT_AUBIO=ALWAYS -DSELF_BUILT_JSON=ALWAYS"
+  echo 'deb http://deb.debian.org/debian buster-backports main' >> /etc/apt/sources.list
```
ooshlablu, line 77: This whole section should be moved to the relevant Dockerfiles for the OS's. The relative Debian commands should be moved to Dockerfile.debian and the 20.04 stuff to Dockerfile.ubuntu

.github/workflows/linux.yml
```diff
@@ -142,7 +142,7 @@ jobs:
           # For some reason, building with 'RelWithDebInfo' (default) doesn't work 
           # specifically on PRs, so we'll just build Releases for everything for now
           #if [ '${{ github.event_name }}' != 'pull_request' ]; then
-            RELEASE_TYPE="-R"
```
ooshlablu, line 145: I don't think this is doing what you think :-)

-R is a flag that the build script uses, so it should only ever be set when releases are made. In this case what should be done is the `if` line above should be uncommented, and the `fi` below. This should stay as -R for builds on master IMO, but I think we need to decide what releases happen where and stuff anyway. As is, this with the comments removed will build release builds for `master` and when we do actual releases via tags. Everything else (which should just be PRs) will be relwithdebinfo.

ooshlablu (changes requested):
A few changes around Dockerfiles and the build script.

.github/workflows/linux.yml
Lord-Kamina: Actually, I think we should always do RelWithDebInfo,unless maybe we ship debug symbols separately

ooshlablu: Yeah, we should probably make that decision with the group on Discord. I don't have a very strong opinion either way really. If we decide to always do ship with debug then obviously this section can be removed.

docker/Dockerfile.debian
```diff
@@ -1,37 +1,52 @@
...
-      libboost-iostreams-dev libboost-locale-dev libboost-system-dev \
-      libboost-program-options-dev libssl-dev libcpprest-dev \
-      libportmidi-dev libopencv-dev libaubio-dev nlohmann-json3-dev \
-      libfmt-dev ca-certificates file && \
-    if [ ${OS_VERSION} -eq 11 ]; then \
-       apt-get -y install libgmock-dev libgtest-dev; fi && \
-    apt-get clean && \
-    mkdir /root/performous
+RUN <<EOR
+#!/bin/bash -x
```
ooshlablu, line 20: This line (and below) are probably being treated as comments. This should be `set +x`. Also make sure this is in line with the commands below since it is part of the same grouping.

docker/Dockerfile.debian
```diff
@@ -1,37 +1,52 @@
...
+		echo 'deb https://apt.kitware.com/ubuntu/ bionic main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null
+		echo 'deb http://deb.debian.org/debian buster-backports main' | tee /etc/apt/sources.list.d/backports.list
+		export_platform_flags '-DSELF_BUILT_AUBIO=ALWAYS -DSELF_BUILT_JSON=ALWAYS'
+		apt-get -y update
+		apt-get -y install libfmt-dev/buster-backports
+	fi
+EOR
+
+RUN <<EOR
+#!/bin/bash -x
```
ooshlablu, line 36: Same comment as above

docker/Dockerfile.fedora
```diff
@@ -1,17 +1,26 @@
...
-        portaudio-devel help2man redhat-lsb opencv-devel portmidi-devel libjpeg-turbo-devel \
-        pango-devel jsoncpp-devel glm-devel openblas-devel fftw-devel cpprest-devel \
-        aubio-devel json-devel rpm-build fmt-devel gtest-devel gmock-devel && \
-    dnf clean all && \
-    mkdir /root/performous
+ENV PLATFORM_CMAKE_FLAGS="-DUSE_BOOST_REGEX=1"
+ENV PACKAGE_TYPE="RPM"
+
+RUN <<EOR
+#!/bin/bash -x
```
ooshlablu, line 12: Same as with Ubuntu

docker/Dockerfile.ubuntu
```diff
@@ -1,28 +1,46 @@
...
+
+## Set up environment variables so the tzdata install doesn't
+## hang on asking for user input for configuration
+ARG DEBIAN_FRONTEND="noninteractive"
+ARG TZ="America/New_York"
+ENV PACKAGE_TYPE="DEB"
+
+## Install the deps and create the build directory
+RUN <<EOR
+#!/bin/bash -x
```
ooshlablu, line 14: Same with Ubuntu

docker/Dockerfile.ubuntu
```diff
@@ -1,28 +1,46 @@
...
+		wget -O ./key.asc https://apt.kitware.com/keys/kitware-archive-latest.asc 2>/dev/null
+		gpg -v -o /usr/share/keyrings/kitware-archive-keyring.gpg --dearmor ./key.asc
+		echo 'deb [signed-by=/usr/share/keyrings/kitware-archive-keyring.gpg] https://apt.kitware.com/ubuntu/ focal main' | tee /etc/apt/sources.list.d/kitware.list >/dev/null
+		apt-get -y update
+		apt-get -y install cmake
+	fi
+EOR
+
+RUN <<EOR
+#!/bin/bash -x
```
ooshlablu, line 27: Same with Ubuntu

ooshlablu (changes requested):
Fix those shebang lines

ooshlablu (approved):
Tested with the AppImage and debs on Ubuntu 20.04 & 22.04 and everything seems to be working as expected.

PR #872, opened 2023-03-24
Fix unit test for format function

twollgam:
### What does this PR do?

Fixes a bug in a unit test for the format of time.

### Closes Issue(s)

None


twollgam:
I close this one cause in other branches the test is disabled and it is too difficult to fix it.

PR #873, opened 2023-04-03
Fix CI badge link

Lord-Kamina:
### What does this PR do?

Fix the link on the CI badge.

But also, I'm using it as a testing ground because the sha for artifacts is still not correct in case of force pushes, and I wanna see if I can fix it.

### More


.github/workflows/build_and_release.yml
```diff
@@ -45,6 +45,16 @@ jobs:
           repository: ${{ github.event.repository.full_name }}
           ref: '${{ github.event.head.ref }}'
           fetch-depth: 0
+      - name: Debug various env stuff
```
Baklap4, line 48: is this one needed in master?

CMakeLists.txt
```diff
@@ -90,7 +90,9 @@ elseif(APPLE)
 	set(CMAKE_INSTALL_MANDIR "${CMAKE_INSTALL_PREFIX}/Resources/man/man6")
 endif()
 
-set(CMAKE_CXX_FLAGS_RelWithDebInfo "${CMAKE_CXX_FLAGS_RelWithDebInfo} -O3")
+if(NOT MSVC)
+	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -O3")
+endif()
```
Baklap4, line 95: Let's also add optimizations for msvc. I think it was -O2 (https://learn.microsoft.com/en-us/cpp/build/reference/o-options-optimize-code?view=msvc-170) optimize for speed

Baklap4 (changes requested):

CMakeLists.txt
Lord-Kamina: /O2 but it's already on, I believe.

PR #875, opened 2023-04-20
Treat warnings as errors

twollgam:
### What does this PR do?

To avoid introducing new warnings they will be treated as errors. 

### Closes Issue(s)

None

### Motivation

Make it easier for developers to find warnings.

### DIFF
--- a/game/CMakeLists.txt
+++ b/game/CMakeLists.txt
@@ -311,6 +311,12 @@ if(WIN32)
 	set_target_properties(performous PROPERTIES OUTPUT_NAME "Performous")
 endif()
 
+if(WIN32)
+	target_compile_options(performous PUBLIC /WX)
+else()
+	target_compile_options(performous PUBLIC -Werror)
+endif()
+
 # Generate config.hh
 configure_file("${CMAKE_CURRENT_SOURCE_DIR}/config.cmake.hh" "${CMAKE_CURRENT_BINARY_DIR}/config.hh" @ONLY)
 include_directories("${CMAKE_CURRENT_BINARY_DIR}")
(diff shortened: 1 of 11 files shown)


game/CMakeLists.txt
```diff
@@ -311,6 +311,12 @@ if(WIN32)
 	set_target_properties(performous PROPERTIES OUTPUT_NAME "Performous")
 endif()
 
+if(WIN32 AND MSVC)
+	target_compile_options(performous PUBLIC /WX)
+else()
+	target_compile_options(performous PUBLIC -Werror)
```
Baklap4, line 317: just wondering if this also works for clang?
If yes let's go!

Baklap4: Might be better to just heck for MSVC isntead of WIN32, as MSVC _can_ be used on unix....

twollgam: > just wondering if this also works for clang? If yes let's go!

It should cause clang uses gccs command line AFAIK

twollgam: > Might be better to just heck for MSVC isntead of WIN32, as MSVC _can_ be used on unix right?

Not that I know. msvc is visual studio only. There is a mac version, but it only supports .net and no c++ AFAIK.

Baklap4: 👍

Baklap4 (approved):

Baklap4:
If we wanna compile with treat warnings as errors on windows we need to specify a warning level instead of `/Wall` i think `/W4` is more than plenty for now @twollgam

PR #878, opened 2023-04-30
Enabled webserver support on windows.

Baklap4:
### What does this PR do?

Enabled webserver support on windows (using cpprestsdk)
Will get superceded by restinio in a later pr.
Hopefully this doesn't break unix

### Closes Issue(s)

Closes https://github.com/performous/performous/issues/632

### Motivation

Webserver was not enabled + broken on windows build due to the `U` macro.
This has been fixed in this pr.

### More


ooshlablu:
Did a quick test with the deb and AppImage on Ubuntu 20.04. I didn't notice any difference in game play, and the webserver functions correctly. I was able to create a playlist, play it, etc, as before.

Tronic:
I found this C++ web server, looks quite excellent in every way. Faster in benchmarks where cpprestsdk apparently deadlocks and dies... Notice run_async() which makes a background threads automatically, so Performous can keep running at the same time.

https://crowcpp.org/master/

Baklap4:
We're not gonna 'just replace the already working backend of the webserver'. This is just a support PR to make the webserver available on windows, which should've been already merged on 1.2.0 launch

In another PR we're changing this to Restinio + NLohmannJson which is scheduled for a 1.4.0 (not the next release but the release after).

Baklap4:
Made this up to date with latest master @ooshlablu

ooshlablu (approved):
LGTM

PR #879, opened 2023-05-01
Cmake minimum version required bump

Baklap4:
### What does this PR do?

* Bumped cmake version required to 3.15 in all files.
* Sets /w4 error level for msvc instead of /wall. Usually /w4 is good enough, `/wall` has issues with stl of mickeysoft

Documentation for cmake says so: https://discourse.cmake.org/t/how-to-set-warning-level-correctly-in-modern-cmake/1103

### Closes Issue(s)

None

### Motivation

Wanting to enable treat warnings as errors on Windows.
Still some warnings to fix on windows before we can enable it again.

### Notes

Hopefully i didn't break any cmake code. I purposely didn't update any syntax that can be done when we clean up cmake code in its entirety

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14.5)
+cmake_minimum_required(VERSION 3.15)
 project(Performous CXX C)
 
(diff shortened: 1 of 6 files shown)


twollgam (approved):

ooshlablu:
Looks good with both the package ancd AppImage on Ubuntu 20.04

PR #883, opened 2023-06-12
Feature/commenttag

Baklap4:
### What does this PR do?

Trying to introduce new (hopefully) standardized fields.
This helps in distinguishing where the mapped file came from and which program made it at which version.
Aside from this `#COMMENT` is sometimes used which is now also supported

### Closes Issue(s)

None

### Motivation

I'm trying to get a couple new standardized fields into the ultrastar format.
This allows hosting websites to add their name to song files whilst creator tools like composer add their specific version and appname.

This way everyone knows where the txt came from and which program is used to create it.

Since the varying sources of which you can download songs from differ in quality this is a nicely added bonus to see where it came from.

Might be a niche for just me, but if everyone opts in it'd make for a way better system and maybe even a newly introduced system which combines all the sources out there.... (that's my endgoal)

### Testing

You can test this by adding the following tags:
```
#SOURCE:USDB
#APP:Composer
#APPVERSION:1.0.0
#COMMENT:Some comment
```

All tags are opt in. But if everyone plays by the rules every txt will have these fields eventually

The tags are shown within the webserver-ui: Playlist and search screen.
All extra tags are seperated by the `|` symbol. And the popup if you click a song also mentions the tags + content if they're there.

### Meanwhile...
Meanwhile i'm trying to convince other karaoke creators and downloadable websites to implement this for every song txt they have. But currently everyone is against it....

### DIFF
--- a/data/themes/default/www/js/playlist-song-modal-body.js
+++ b/data/themes/default/www/js/playlist-song-modal-body.js
@@ -136,6 +136,50 @@ $(function () {
     }
 
     /*
+        If our song contains a source show it.
+        Else hide the complete div.
+    */
+    if (songObject.Source.length > 1) {
+        $("#playlist-song-modal-body #modal-source-div").show();
+        $("#playlist-song-modal-body #modal-source-value").text(songObject.Source);
+    } else {
+        $("#playlist-song-modal-body #modal-source-div").hide();
+    }
+
+    /*
+        If our song contains an app show it.
+        Else hide the complete div.
+    */
+    if (songObject.App.length > 1) {
+        $("#playlist-song-modal-body #modal-app-div").show();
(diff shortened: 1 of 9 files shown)


Baklap4:
Undrafted this. Ready for review :)
Also updated the Original post

Baklap4:
Waiting for finalizing: https://github.com/UltraStar-Deluxe/format/issues/35

Baklap4:
The gods have decided. A new tag is introduced to credit hosting website. The key is: `#PROVIDEDBY`

See: https://github.com/UltraStar-Deluxe/format/issues/35#issuecomment-1819221345

Will fix this in a next commit

Baklap4:
Was replaced by https://github.com/performous/performous/pull/975 since github didn't let me refresh this PR.... kind of weird...

PR #885, opened 2023-06-27
Move the AppImage builder to Ubuntu 22.04

ooshlablu:
### What does this PR do?

Move the AppImage builder to Ubuntu 22.04

### Motivation
An issue popped up in `user_support` on Discord that a user was not able to run the AppImage on a newer version of Suse. Running it would crash with the following error:
```
libstdc++.so.6: version 'GLIBCXX_3.4.29' not found
```
Ubuntu 22.04 is using a newer version of the various `libc6` components, so I suspected bumping the runner version up would fix the issue.

### Additional Notes

`libfuse2` was added to the dependencies installed on the builder because it is required to run AppImages for now and 22.04 does not come with it installed by default.

We do still have some workarounds for Arch and Fedora, so those should be tested with the AppImage as well.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -17,7 +17,7 @@ jobs:
   # Create the AppImage
   AppImage:
     name: Create the AppImage
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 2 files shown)


Baklap4 (approved):
Lgtm

Issue #886, opened 2023-06-27
Access Violation in repositioning songs


Baklap4:
### Performous & operating system versions

Latest git version on Windows 11

### What is the problem?

Changing the position (through webserver) from 2nd place to 1st place crashes the program with a ViolationException
![image](https://github.com/performous/performous/assets/3918517/d73916b2-87bd-4de4-970c-cb4728991f06)

### Steps to reproduce

1. Start performous
2. Play a song
3. go to localhost:8000
4. add 2 songs through the WebUI
5. Go to view playlist in webUI
6. Click the 2nd song
7. Press Move Up to move it up by 1.
8. Crash

### Additional context

```
void PlayList::swap(unsigned index1, unsigned index2) {
	std::lock_guard<std::mutex> l(m_mutex);
	std::shared_ptr<Song> song1 = m_list[index1];
	m_list[index1] = m_list[index2];
	m_list[index2] = song1;
}
```
This code is getting hit.
Here we see `song1` is a copy of the reference to `m_list[index1]`. Which then gets overwritten with `m_list[index2]`. `song1` in this case is not the original `index1` content which results into a violation exception i think.

How to fix this?

PR #887, opened 2023-06-27
Remove Playlist SetPosition

Baklap4:
### What does this PR do?

* Removes the SetPosition function which was only called from requesthandler.
* Changes requesthandler to use the swap method on Playlist.

Less code, better maintainability

### Closes Issue(s)

#886 

### Motivation

I want to be able to use the up/down/set-position functionality in webserver.


ooshlablu (approved):
LGTM
I was able to produce the bug with the current version on the README.
I tested this build on Ubuntu 20.04 & 22.04 and it seems to have fixed the issue. No other functionality or game play was affected.

PR #888, opened 2023-07-03
Playlist fixes

Baklap4:
### What does this PR do?

* Webserver alert messages keys have been changed. But UI hasn't been updated. Alert messages now work again
* Using Set Position from the modal now works correctly and no longer swaps the songs
    * For example: setting position of last song to the first one used to swap the first and last song whilst it should just move the last song to the front.
* The Move Up functionality in playlist screen didn't work this is also fixed now.

### Closes Issue(s)

Some issues mentioned within the discord channel.

### Motivation

Playlist should work just fine. And the 'null' messages from the web-ui should've been already fixed when the keys were changed.


ooshlablu (approved):
LGTM. Tested with the AppImage and deb on Ubuntu 20.04 and 22.04. Fixes all the issues mentioned, and game play appears to be unaffected overall.

PR #896, opened 2023-08-02
Update Slovak translation.

jose1711:
### What does this PR do?

Update Slovak translation so that all messages in .pot are translated, few fixes too.

### Motivation

Slovak translation was a bit out of sync.


Baklap4:
Can you please target the 1.3.0-rc1 branch so this gets included in the official release?

PR #897, opened 2023-08-06
Allow all connections

Baklap4:
### What does this PR do?

While testing 1.3.0-rc1 we found that on windows we cannot accept all connections.
This is a problem within cpprestsdk as we bind wrongly for Windows.

Added an if-statement to check for OS and added the correct binding address.

### Closes Issue(s)

None

### Motivation

All connections for webserver should work just fine on Windows as it also does for Linux/MacOS

### Additional Notes

Windows users still need Admin privileges to run the webserver with All connections.

### DIFF
--- a/game/webserver.cc
+++ b/game/webserver.cc
@@ -1,5 +1,6 @@
 #include "webserver.hh"
 #include "game.hh"
+#include "platform.hh"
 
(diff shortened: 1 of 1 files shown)


ooshlablu (approved):
LGTM

Issue #900, opened 2023-08-06
"Refresh database" in the webserver is not working unless the "Perform" screen is visited


HetorusNL:
### Performous & operating system versions

Performous: 1.3.0-rc1
Operating system: Windows 11 Pro 22H2 x64

### What is the problem?

When Performous is started and the web server is opened, the "Refresh database" button is not working if the "Perform" screen isn't visited before. After visiting the "Perform" screen, the "Refresh database" button is working correctly and reloads the database

### Steps to reproduce

1. Open Performous
2. Open web server
3. Click "Refresh database" on the web server
4. [No songs are loaded] <- the issue
5. Click "Perform" in Performous
6. Click "Refresh database" on the web server
7. [Songs are loaded correctly] <- what should happen at step 4.

Baklap4:
Closing since #901 has been merged to 1.3.0 branch

PR #901, opened 2023-08-06
Fix: apply the filter to the songs before passing them to the web server

HetorusNL:
### What does this PR do?

Edit: call the `m_songs.update()` function from `requesthandler.cc` before constructing the json and passing the songs to the web server.
~This PR adds a call to `filter_internal()` after the songs are loaded from cache/filesystem.~
~This way, the web server can load the songs after this filtering is done.~

### Closes Issue(s)

#900 

### Motivation

Being able to use the web server before having to enter the "Perform" screen.

### More

### DIFF
--- a/game/requesthandler.cc
+++ b/game/requesthandler.cc
@@ -100,6 +100,8 @@ void RequestHandler::Get(web::http::http_request request)
         } else if(query == "sort=creator&order=descending") {
             m_songs.sortSpecificChange(10, true);
         }
+        // make sure to apply the filtering
+        m_songs.update();
         web::json::value jsonRoot = SongsToJsonObject();
         request.reply(web::http::status_codes::OK, jsonRoot);
         return;


ooshlablu (approved):
Tested with Ubuntu 20.04. This definitely resolves the issue, and I have not noticed any detrimental effects in any of the options screen or during game play.

Baklap4 (changes requested):
This is like sticking a bandaid... the problem seems to go away, but its actually still there.
This will only be loaded once the whole library is loaded. In between this time the refresh button still returns an empty array.

Try it while loading a couple 100 songs.

A better place to look would be the requestprovider and why the endpoint returns an empty array instead of the (partially) filled songscollection

Baklap4 (approved):
As discussed on discord, i think this works for now. But we need to refactor the code within songs.cc but thatll be another pr

PR #902, opened 2023-08-06
Fixed song sorting from within the web server

HetorusNL:
### What does this PR do?

Fixed the sorting of songs on artist/title/language/edition/creator.
Added sorting class of the 'creator' field of songs.
Added `operator=(Type value)` to the Cycle class:
* Otherwise the constructor is used which creates a new Cycle instance with `Cycle(value, 1, 0)` default max/min values
* This fails the checks resulting in an `throw std::logic_error(..)` and a HTTP 500 returned to the web server

### Closes Issue(s)

#903 

### Motivation

Wanting to sort the songs in the web server.

### More

### DIFF
--- a/game/requesthandler.cc
+++ b/game/requesthandler.cc
@@ -95,6 +95,10 @@ void RequestHandler::Get(web::http::http_request request)
             m_songs.sortSpecificChange(3);
         } else if(query == "sort=edition&order=descending") {
             m_songs.sortSpecificChange(3, true);
+        } else if(query == "sort=creator&order=ascending") {
+            m_songs.sortSpecificChange(10);
+        } else if(query == "sort=creator&order=descending") {
+            m_songs.sortSpecificChange(10, true);
         }
         web::json::value jsonRoot = SongsToJsonObject();
         request.reply(web::http::status_codes::OK, jsonRoot);
--- a/game/utils/cycle.hh
+++ b/game/utils/cycle.hh
@@ -11,6 +11,7 @@ class Cycle {
 	Cycle& backward();
 
 	operator Type() const;
+	Cycle& operator=(Type value);
 
 	Type getMin() const;
 	Type getMax() const;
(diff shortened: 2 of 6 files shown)


game/utils/cycle.hh
```diff
@@ -57,6 +58,11 @@ Cycle<Type>& Cycle<Type>::set(Type value) {
 	return *this;
 }
 
+template<class Type>
+Cycle<Type>& Cycle<Type>::operator=(Type value) {
```
Baklap4, line 62: Since we have some unittests provided by @twollgam please also create a test for the overload 👍

Baklap4 (changes requested):

game/utils/cycle.hh
HetorusNL: Added and tested the unittests for `Cycle::operator=(Type)` 👍

ooshlablu:
I tested this on Ubuntu 20.04 with the deb and AppImage and it looks good. Sorting works properly (at least for the metadata categories that exist in my sources) in the WebUI. I was able to create playlists from the sorted item and stuff like that, so that all looks good. I didn't notice any changes to game play experience, which is expected.

LGTM

game/requesthandler.cc
```diff
@@ -95,6 +95,10 @@ void RequestHandler::Get(web::http::http_request request)
             m_songs.sortSpecificChange(3);
         } else if(query == "sort=edition&order=descending") {
             m_songs.sortSpecificChange(3, true);
+        } else if(query == "sort=creator&order=ascending") {
+            m_songs.sortSpecificChange(10);
+        } else if(query == "sort=creator&order=descending") {
+            m_songs.sortSpecificChange(10, true);
```
Baklap4, line 101: Hmmm better would be to change `sortSpecificChange` to accept an enum. This way we don't spread magic numbers throughout the code.
If it's too much for this, then please create an issue mentioning this refactor so we can do it later.

The changes itself work :)

Baklap4 (changes requested):

game/requesthandler.cc
HetorusNL: It might be better to do that in a separate PR indeed, as also the `addSongOrder(..)` and other related code needs to work with these enum class values.

Baklap4: Created issue: #904

Baklap4 (approved):
Issue has been created, can be merged now :)

Issue #893, opened 2023-07-31
Loading utf8 files on windows fails when 'Beta: Use Unicode UTF-8 for worldwide language support' is not enabled


HetorusNL:
### Performous & operating system versions

Performous: 1.2.0.394 (latest msvc 2023-07-30)
Operating system: Windows 11 Pro 22H2 x64

### What is the problem?

When trying to import files with non-ascii characters, like `Can’t` or `böhse onkelz`, the application crashes: it doesn't start and Event Viewer shows the following error: 
![image](https://github.com/performous/performous/assets/17402329/0e26afa2-3675-4c5c-9c95-cf335ff14c84)
![image](https://github.com/performous/performous/assets/17402329/432bd316-6f86-45dc-a75f-524b6287a78e)

### Steps to reproduce

1. Add a song with for example `ö` in the file path
2. Add the folder with that song to the song paths in Performous
3. Performous crashes trying to load the files (with aforementioned error)
4. Subsequent launches result in an immediate crash without showing the GUI

### Log(s) & preferences file(s)

[infolog.txt](https://github.com/performous/performous/files/12222352/infolog.txt)

Baklap4:
@Lord-Kamina what would be the preffered way to handle this?
So far as i can see the current proposed changes are working, but in discord there was a little discussion going on...

For the 1.3.0 release this is not a blocker, but we might introduce a new patch version soon after to get this fixed (1.3.1)

PR #908, opened 2023-08-22
Fixes UTF8 Paths with UnicodeUtil class

Baklap4:
### What does this PR do?

This PR fixes an issue on windows that crashes Performous when importing songs with characters like `ö` and `’` while not having the beta utf8 support enabled in settings.

Without this change the `Song::getDurationSeconds` function crashes on `avformat_open_input` while supplying a path as mentioned above.

### Closes Issue(s)

#893 

### Motivation

Running Performous on windows while not having to enable utf8 support.

### More

### Additional Notes

Tested on Win11 with the beta utf8 support disabled.

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -4,6 +4,7 @@
 #include "config.hh"
 #include "screen_songs.hh"
 #include "util.hh"
+#include "unicode.hh"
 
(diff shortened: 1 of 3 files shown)


Baklap4:
Alternative was merged #934 . Closing it

PR #910, opened 2023-08-30
Improve the FindCed.cmake script

abompard:
### What does this PR do?

It gives the option to use the system-installed `compact_enc_det` instead of building it from git.

### Motivation

This is necessary to build Flatpaks, because the builder runs in a container without network access.


Baklap4:
This looks good!
@ooshlablu can you test this? It'll be merged, to master, when 1.3.0 is released (somewhere soon)

Thanks for the PR @abompard

Baklap4 (approved):

PR #911, opened 2023-08-30
Add the Aubio include dirs

abompard:
### What does this PR do?

This adds the Aubio include dirs to the compiler path as detected using pkgconfig.

### Motivation

Now that we can use the system aubio instead of downloading and rebuilding it (bf5e42c0), we also need to add the include path.


Baklap4:
This is looking good! Thanks for the PR @abompard @ooshlablu can you give this a test run/build

Baklap4:
@ooshlablu this just need a test run to see if it's still running 👍

ooshlablu:
Just did a quick sing-test on Ubuntu 20.04, and seems fine. Will try with guitars and stuff in a little bit, but I think this change will leave all of that unaffected.

Baklap4 (approved):

ooshlablu:
Tested with guitars and singing, everything looks good.

PR #912, opened 2023-08-30
Use DATADIR as installation path for the desktop file and the icon

abompard:
### What does this PR do?

This uses the DATADIR build-specified directory to install the desktop file and the icon.
It also uses the more standard DATADIR/icons/hicolor/scalable/apps directory instead of DATADIR/pixmaps.

Reference: https://specifications.freedesktop.org/icon-theme-spec/latest/ar01s03.html

### Motivation

Before this PR, the installation paths were hardcoded to `/usr/share`, which does not work well with a Flatpak installation, because everything is in `/app`

### DIFF
--- a/data/CMakeLists.txt
+++ b/data/CMakeLists.txt
@@ -3,8 +3,8 @@ set(PIXMAP_FILE      "themes/default/icon.svg")
 
 # Install launcher and fonts on system level
 if(UNIX AND NOT APPLE)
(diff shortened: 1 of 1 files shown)


Baklap4:
LGTM once 1.3.0 is released we'll merge this into master :)

Baklap4 (approved):

PR #913, opened 2023-08-30
Add Appstream metadata

abompard:
### What does this PR do?

This PR adds the [Appstream](https://www.freedesktop.org/software/appstream/docs/) metadata file.

### Motivation

An Appstream metadata file is required to show up in gnome-software, or to be listed as a GUI application in Flatpak.

### Additional Notes

Reference: https://www.freedesktop.org/software/appstream/docs/

I've added the only screenshot I found on performous.org, but it's a carroussel, I'd perfer to add other higher-resolution screenshots if there are any.

### DIFF
--- a/data/performous.metainfo.xml
+++ b/data/performous.metainfo.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.performous.Performous</id>
+  <metadata_license>CC0</metadata_license>
+  <project_license>GPL-2.0-or-later</project_license>
+  <name>Performous</name>
+  <summary>Music and rhythm / party game</summary>
+  <description>
+    <p>
+      A karaoke, band and dancing game where one or more players perform a
+      song and the game scores their performances.
+    </p>
+    <p>
+      Supports songs in UltraStar, Frets on Fire and StepMania formats.
+      Microphones and instruments from SingStar, Guitar Hero and Rock Band
+      as well as some dance pads are auto-detected.
+    </p>
+  </description>
+  <screenshots>
+    <screenshot type="default">
+      <image type="source" width="1536" height="96">http://performous.org/imgs/showcase.jpg</image>
+    </screenshot>
+  </screenshots>
+  <url type="homepage">https://performous.org</url>
+  <content_rating type="oars-1.1" />
+  <launchable type="desktop-id">performous.desktop</launchable>
+  <releases>
+    <release version="1.2.0" date="2022-03-27">
+      <artifacts>
+        <artifact type="source">
+          <location>https://github.com/performous/performous/archive/refs/tags/1.2.0.tar.gz</location>
+          <checksum type="sha256">c88438bb3b9400a2254a99284375157534ae38befc48ba27254bc3b9e7ef50ef</checksum>
+        </artifact>
+      </artifacts>
+    </release>
+  </releases>
+</component>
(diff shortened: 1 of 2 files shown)


data/performous.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+  <summary>Music and rhythm / party game</summary>
+  <description>
+    <p>
+      A karaoke, band and dancing game where one or more players perform a
+      song and the game scores their performances.
+    </p>
+    <p>
+      Supports songs in UltraStar, Frets on Fire and StepMania formats.
+      Microphones and instruments from SingStar, Guitar Hero and Rock Band
+      as well as some dance pads are auto-detected.
```
Baklap4, line 16: Is there any way to translate this?

data/performous.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+  <launchable type="desktop-id">performous.desktop</launchable>
+  <releases>
+    <release version="1.2.0" date="2022-03-27">
+      <artifacts>
+        <artifact type="source">
+          <location>https://github.com/performous/performous/archive/refs/tags/1.2.0.tar.gz</location>
+          <checksum type="sha256">c88438bb3b9400a2254a99284375157534ae38befc48ba27254bc3b9e7ef50ef</checksum>
+        </artifact>
+      </artifacts>
+    </release>
```
Baklap4, line 35: Is there any way to automate this instead of manually typing it on every release?
Also how should pre-releases be formed? (would it be just another release or marked with something?)
If the latter let's also add rc1 in here

Baklap4 (changes requested):

data/performous.metainfo.xml
abompard: Yes, these files can be translated using Intltool or Itstool, for example: https://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Translation.html
I see that we have a script in `lang/xml_gettext.py` that extracts strings from `schema.xml`. Should we switch to something else or extend it?

Baklap4: We currently use po-files for translations (by using xgettext see the `lang` folder). It would be nice if we can support this from the gecko

The xml_gettext.py extracts strings from schema.xml, might be nice to extend this to also include this metadata file

abompard: OK! I take it you don't want to use something like [itstool](https://itstool.org/) or intltool for both schema.xml and the metainfo.xml file?
(not sure what "from the gecko means, sorry (not a native English speaker here))

abompard: I've added a commit that uses itstool for XML translations, just as a POC of its use. It's of course for you to decide if you want to add it as a dependency, but I find it more manageable if we want to translate more XML in the future. It also removes the need to maintain `xml_gettext.py`. And last thing, it's capable of building a translated XML file using the `.mo` files, which would have been pretty hard to do with `xml_gettext.py`.

abompard: Hmm, I don't understand why CI fails, I've added the itstool package to the dockerfiles, I see it being installed, but cmake still doesn't find it.

abompard: Any idea what could be going on @Baklap4 ?

ooshlablu: I _think_ at one point something similar happened, anc IIRC it was because of something wrong with `FIND_PACKAGE` or one of the other built-in versions. I may be mis-remembering tho

Baklap4:
I'll take a look at this when 1.3.0 is released (somewhere in the upcoming days)

Baklap4:
Gotta take a look at itstool if it's working on windows else we might need xml_gettext.py back in

PR #917, opened 2023-09-24
Showing error for songs that do not contain audio.

### LINKED ISSUE #127: Performous should not list broken songs
If a song is incomplete (missing MP3), it should be reported as broken and not be shown in the song browser.

Baklap4:
### What does this PR do?

The awesome project of https://github.com/bohning/usdb_syncer creates txt files without mandatoring the mp3 file to be there.
These files get loaded within performous but the songs won't work since there's no audio. For the keen eye, if you enabled autoplay with a certain timeout. You could identify this by seeing the offset of the timeout (in my case 10 seconds) as the duration for the song when added to the queue. This is not really a good way to tell the user it's a faulty song.

This feature tries to solve that issue.

From the webserver point you get to see:
Searching for a song:
image

Adding a faulty song:
image

Database view:
image

From within performous you'll get a message that the song couldn't be opened:
image

### Closes Issue(s)

# Closes #127

### Motivation

Telling the enduser to be able to play the song, but it can't be played is not nice. So we shouldn't do this.
This tells the user something is up with the song, please check later.

### More

### Additional Notes

When you fixed the song one should bust the cache entirely.
Tried to make it auto-bust but we have a thing called "guessFiles" which is fucking this up. Can be done at a later time when we refactor this.

### DIFF
--- a/game/requesthandler.cc
+++ b/game/requesthandler.cc
@@ -131,6 +131,7 @@ void RequestHandler::Get(web::http::http_request request)
             songObject[utility::conversions::to_string_t("Language")] = web::json::value::string(utility::conversions::to_string_t(song->language));
             songObject[utility::conversions::to_string_t("Creator")] = web::json::value::string(utility::conversions::to_string_t(song->creator));
             songObject[utility::conversions::to_string_t("Duration")] = web::json::value(song->getDurationSeconds());
+            songObject[utility::conversions::to_string_t("HasError")] = web::json::value::boolean(song->loadStatus == Song::LoadStatus::ERROR);
             jsonRoot[i] = songObject;
             i++;
         }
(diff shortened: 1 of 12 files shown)


Baklap4:
@ooshlablu this is easy to test:
1. Grab a txt file and specify a `#MP3` source
2. Make sure there's no MP3 or any audio file available in the songfolder

Song should be turning up with a ⚠️ and note in game. This should still be the case after caching  en restarting

game/requesthandler.cc
```diff
@@ -169,7 +170,13 @@ void RequestHandler::Post(web::http::http_request request)
         if(!songPointer) {
             request.reply(web::http::status_codes::NotFound, "Song \"" + utility::conversions::to_utf8string(jsonPostBody[utility::conversions::to_string_t("Artist")].as_string()) + " - " + utility::conversions::to_utf8string(jsonPostBody[utility::conversions::to_string_t("Title")].as_string()) + "\" was not found.");
             return;
-        } else {
+        }
+        else if (songPointer->loadStatus == Song::LoadStatus::ERROR)
+        {
+            request.reply(web::http::status_codes::NotFound, "Song \"" + utility::conversions::to_utf8string(jsonPostBody[utility::conversions::to_string_t("Artist")].as_string()) + " - " + utility::conversions::to_utf8string(jsonPostBody[utility::conversions::to_string_t("Title")].as_string()) + "\" Song load status is error. Please check what's wrong with it.");
```
twollgam, line 176: Would be good to break the line in to smaller pieces.

twollgam (changes requested):
LGTM only one line is a little bit too long ;-)

ooshlablu:
@Baklap4 I tested it on Ubuntu 20.04 & 22.04 and it works. Didn't notice any effects on gameplay or other features, so :+1:

game/requesthandler.cc
Baklap4: Fixed

twollgam (commented):
There are minor issue like indent. Don't know if you want to fix it.

Else it LGTM so feel free to merged it with or without fix.

twollgam (approved):

PR #925, opened 2023-10-18
Change from NEVER to AUTO for ced, aubio and json in cmake

twollgam:
### What does this PR do?

The default value NEVER is changed to AUTO for the libraries ced, aubio and json in cmake. NEVER has the meaning, that the library is never build with performous. Therefore a system version of the library is required. If those are not available cmake failes.

Changing the default to AUTO is the better choice, cause in this case cmake will look for a system library and if none is available it will be build with performous.

### Closes Issue(s)

None.

### Motivation

It is a pain for a developer if cmake doesn't work out of the box. The new default values will make it much easier to build performous from sources.

### DIFF
--- a/game/CMakeLists.txt
+++ b/game/CMakeLists.txt
@@ -121,9 +121,10 @@ find_package(PkgConfig REQUIRED)
 pkg_check_modules(deps REQUIRED IMPORTED_TARGET glib-2.0 gio-2.0 gobject-2.0)
 
 set(SELF_BUILT_GIT_BASE "https://github.com/performous" CACHE STRING "base path of the git trees used for self built dependencies")
-set(SELF_BUILT_AUBIO "NEVER" CACHE STRING "rule to allow self build of Aubio dependency library [NEVER*|AUTO|ALWAYS]")
-set(SELF_BUILT_CED "NEVER" CACHE STRING "rule to allow self build of Ced dependency library [NEVER*|AUTO|ALWAYS]")
-set(SELF_BUILT_JSON "NEVER" CACHE STRING "rule to allow self build of Json dependency library [NEVER*|AUTO|ALWAYS]")
+set(SELF_BUILT_AUBIO "AUTO" CACHE STRING "rule to allow self build of Aubio dependency library [NEVER|AUTO*|ALWAYS]")
+set(SELF_BUILT_JSON "AUTO" CACHE STRING "rule to allow self build of Json dependency library [NEVER|AUTO*|ALWAYS]")
+# This is AUTO cause CED is not available on ubuntu
+set(SELF_BUILT_CED "AUTO" CACHE STRING "rule to allow self build of Ced dependency library [NEVER|AUTO*|ALWAYS]")
 
 set(Aubio_REQUIRED_VERSION "0.4.9")
 


yoda-jm:
I would like to keep the default as NEVER instead of AUTO because the default behavior should be predictible (or else it become complicated for developer to debug with what version it was built).
However we can probably improve the error for those libraries, by having a nice error message if they are not found in the NEVER case to set the correct CMake switch to ALWAYS (and we can probably completly get rid of AUTO)

Lord-Kamina:
> I would like to keep the default as NEVER instead of AUTO because the default behavior should be predictible (or else it become complicated for developer to debug with what version it was built). However we can probably improve the error for those libraries, by having a nice error message if they are not found in the NEVER case to set the correct CMake switch to ALWAYS (and we can probably completly get rid of AUTO)

I agree re: predictability but not with where you go from there.
IMO, if it's properly documented, using "AUTO" is predictable and failures due to not following instructions or reading the docs are the packager's responsibility.

Baklap4 (changes requested):
Auto is the better choice here, it has predictable behavior: Try system if not available build it.
This works for pretty much all usecases.

Is the apple build script in need of change? since i think we overrule it there
If not let's get it merged :)

twollgam:
> I would like to keep the default as NEVER instead of AUTO because the default behavior should be predictible (or else it become complicated for developer to debug with what version it was built).
> However we can probably improve the error for those libraries, by having a nice error message if they are not found in the NEVER case to set the correct CMake switch to ALWAYS (and we can probably completly get rid of AUTO)

I agree if you want to get something predictable. Only your conclusion is wrong, cause NEVER leads to undefined situations. You will never know on which platform what libraries are used. The most predictable situation you can get if the default is ALWAYS. And you convinced me to use that. @Lord-Kamina and @Baklap4  what do you think about ALWAYS?

The reason that this issue/pr exists is that NEVER as default for ced makes the build system unusable on debian and ubuntu (there may be others). And that's a no go.

yoda-jm:
I think the NEVER/ALWAYS depends on the library:
- when there is no chance to have compatible version installed locally (system) we need to use ALWAYS
- when there is a good chance that it is possible to have a compatible version installed locally (system) we need to use NEVER.

Things that should be considered:
- status of the library in Ubuntu latest
- status of the library in Ubuntu LTS
- status of the library in current Debian
- status of the library in Gentoo, ArchLinux, latest RHEL, ...

I think if the library is ok un Ubuntu latest we can choose NEVER for example, because people compiling Performous from scratch are often on latest.
Also if in all those distributions the version packages is working with Performous we need to completly delete the support for in-tree library.

Baklap4:
> I agree if you want to get something predictable. Only your conclusion is wrong, cause NEVER leads to undefined situations. You will never know on which platform what libraries are used. The most predictable situation you can get if the default is ALWAYS. And you convinced me to use that. @Lord-Kamina and @Baklap4 what do you think about ALWAYS?


My preference would still be `AUTO` (like in the pr) as this would fix most library-find issues on each system. I can see the Always option work too, but this creates a dependency towards the internet which isn't favorable i think.

> The reason that this issue/pr exists is that NEVER as default for ced makes the build system unusable on debian and ubuntu (there may be others). And that's a no go.

Agreed

Baklap4 (approved):

PR #926, opened 2023-10-22
Fix macOS Builds.

Lord-Kamina:
### What does this PR do?

This is the clean version of the patches to fix macOS builds.

An explanation of the problem, to avoid stepping on the same stone again: Performous runs a loop with pretty much all possible combinations of base paths and common subfolders to find the configuration schema. Not finding this file makes performous get stuck in a loop so early in the program, not even the log is running yet.

In a previous PR, I had changed a CMake path used for installing data files into an absolute path, not realizing that this path gets hardcoded into the game as a possible subfolder to locate the configuration schema. The macOS package currently relies on this specific folder and by making it an absolute path, it meant that the game never found the needed file and thus became stuck in a loop.

### More


game/webcam.hh
```diff
@@ -38,6 +38,9 @@ class Webcam {
 	Dimensions const& dimensions() const { return m_texture.dimensions; }
 
   private:
+#ifdef __clang__
+	[[maybe_unused]]
+#endif
```
Baklap4, line 43: Can be removed i think?

Lord-Kamina: It can't.
gcc is stupid and emits a warning about not emitting a warning in the first place, and that breaks the build due to `-Werror`

Baklap4: That's a shame. Then that's alright for now. Maybe it's fixed in some later version of gcc

Baklap4 (approved):

PR #929, opened 2023-10-29
Texture management

twollgam:
### What does this PR do?

It introduces texture management. Currently textures used twice are loaded twice. The new texture management loads the textures only once.

### Closes Issue(s)

None.

### Motivation

There will follow prs where some textures are used multiple times.

Additionally this pr can be the base for using mega textures.

### More

- Actually the user experience should not be changed.
- The use of class Texture (and OpenGLTexture) will be much more easy now, cause moving, copying and assigning is now possible.


game/graphic/bitmap.hh
```diff
@@ -54,3 +58,4 @@ void writePNG(fs::path const& filename, Bitmap const& bitmap, unsigned stride =
 void loadPNG(Bitmap& bitmap, fs::path const& filename);
 void loadJPEG(Bitmap& bitmap, fs::path const& filename);
 
+void load(Bitmap& bitmap, fs::path const& name);
```
Baklap4, line 61: Hmmm would it be handy to be more explicit about visibility? I know methods/fields are private per default.. but i found this out only after googling.. It might be a bit more verbose but makes it very explicit. In c# we even get warnings when we don't specify a visibility modifier

game/graphic/texture_loader.cc
```diff
@@ -0,0 +1,190 @@
...
+    public:
+        static TextureLoader& instance() {
+            if (!m_instance)
+                m_instance = std::make_unique<TextureLoader>();
+
+            return *m_instance;
+        }
+        static void destroy() {
+            m_instance.reset();
+        }
```
Baklap4, line 24: Hmmm singleton here, do we need it to be singleton? or is it possible to make it transient (everytime new when called and going out of scope when not needed anymore) lifetime so bitmaps get disposed when they're not in use anymore? this also prevents a migration to possible dependency injection

Baklap4 (changes requested):
I like how you moved the classes and cleaned the code up more :D! i don't quite like the singleton approach, but maybe that's a first step in the right direction and can be dismissed for now?

game/graphic/texture_loader.cc
twollgam: The TextureLoader was a singleton before this. So I decided not to change it to much. And we could remove it, but that will be a whole bunch of changes. Then every place where we load/define a texture must be modified.

game/graphic/bitmap.hh
twollgam: Sorry, I don't understand your point. I think the image loading code is very ugly and I would like to change it one day, but before I want to merge the branches which exists already.

game/graphic/texture_loader.cc
twollgam: And there are still too many places where the singleton is needed. Would be nice but it is not the scope of this pr

Baklap4 (approved):

PR #892, opened 2023-07-30
Fixed crash importing songs with unicode in path on non-utf8 locale

HetorusNL:
### What does this PR do?

This PR fixes an issue on windows that crashes Performous when importing songs with characters like `ö` and `’` while not having the beta utf8 support enabled in settings.

Without this change the `Song::getDurationSeconds` function crashes on `avformat_open_input` while supplying a path as mentioned above.

### Motivation

Running Performous on windows while not having to enable utf8 support.

### More

- [ ] Added/updated documentation

PR #934, opened 2023-11-08
set the locale to UTF-8 on windows during startup of Performous

### LINKED ISSUE #893: Loading utf8 files on windows fails when 'Beta: Use Unicode UTF-8 for worldwide language support' is not enabled

HetorusNL:
### What does this PR do?

This PR sets the locale on windows machines to UTF-8 during the startup of Performous.

Background of the issue:
_This PR fixes an issue on windows that crashes Performous when importing songs with characters like ö and ’ while not having the beta utf8 support enabled in settings._
_Without this change the Song::getDurationSeconds function crashes on avformat_open_input while supplying a path as mentioned above._

### Closes Issue(s) / PR's

Closes #893 
Closes #892 
Closes #902 

### Motivation

Running Performous on windows while not having to enable utf8 support.

### More

### Additional Notes

Tested on Win11 with the beta utf8 support disabled.

Related PRs fixing the same issue:
- https://github.com/performous/performous/pull/892
- https://github.com/performous/performous/pull/908

### DIFF
--- a/game/main.cc
+++ b/game/main.cc
@@ -262,15 +262,8 @@ static void fatalError(const std::string &msg) {
 	std::clog << "core/error: " << errMsg << std::endl;
 }
 
-void setupPreconditions() {
-#ifdef WIN32
-	// set the locale to UTF-8 on windows
-	setlocale(LC_ALL, ".UTF8");
-#endif
-}
-
 int main(int argc, char** argv) try {
-	setupPreconditions();
+	Platform::setupPlatform();
 	std::srand(static_cast<unsigned>(std::time(nullptr)));
 	// Parse commandline options
 	std::vector<std::string> devices;
--- a/game/main.cc
+++ b/game/main.cc
@@ -262,7 +262,15 @@ static void fatalError(const std::string &msg) {
 	std::clog << "core/error: " << errMsg << std::endl;
 }
 
+void setupPreconditions() {
+#ifdef WIN32
+	// set the locale to UTF-8 on windows
+	setlocale(LC_ALL, ".UTF8");
+#endif
+}
+
 int main(int argc, char** argv) try {
+	setupPreconditions();
 	std::srand(static_cast<unsigned>(std::time(nullptr)));
 	// Parse commandline options
 	std::vector<std::string> devices;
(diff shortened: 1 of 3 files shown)


twollgam (dismissed):
LGTM

twollgam:
Testet it on Windows 11. Where other versions crash works this one perfect. 👍

ooshlablu:
I tested on Ubuntu 20.04 & 22.04 just to be sure, and didn't notice any changes there. Everything functions as expected. :+1:

Lord-Kamina:
Question... is this really necessary? What happens if we pass the file paths through UnicodeUtil before using them on `song::getDurationSeconds`?

And, assuming UnicodeUtil just doesn't cut it for this (I doubt it would have much trouble but 🤷🏼‍♂️), could we move the fix inside the Platform class?

HetorusNL:
Also when writing to and reading from the json cache, unicode-related issues occur (has been noted on discord some while ago).
And the UnicodeUtil is used in #908 , but while testing this the cache isn't working for songs with unicode characters and the log is bloated with encoding warnings from UnicodeUtil.

Baklap4:
If it's possible let's move it to the `Platform` class since platform specific stuff lives here.

This should be a fairly easy fix for windows so that's pretty nice.

I think for now we should go with this PR and close the related ones. At some point we can remove this again to replace it with fully `std::u8string` support at which point we won't need the unicode util class anymore. @HetorusNL can you create a, maybe a couple, of issues to remove this and unicodeutil and replace it with `std:u8string`

Baklap4 (approved):
LGTM

Issue #949, opened 2023-11-26
Song thumbnail disappears after stopping the song


berarma:
### Performous & operating system versions

Performous 1.3.0-45-gcb10e1c6 (latest master)
OS: Debian 12.2

### What is the problem?

Song thumbnails disappear from the browser screen after exiting from the song playback.

### Steps to reproduce

1. Select a song and play it.
2. Press Esc and select Exit from the menu.
3. The thumbnail for that song is no longer visible.
4. Repeating the same with another song makes the thumbnail for the previous song appear and the one for the last played song disappears.

### Log(s) & preferences file(s)

The image shows the missing thumbnail after having played the song and exited:

![Captura desde 2023-11-26 18-16-35](https://github.com/performous/performous/assets/162766/49efc2af-2939-473a-8f04-e506cef19a46)

Baklap4:
@twollgam this is since the texture rework branch. We might have to fix this before doing the release...

twollgam:
I will look for it next week ...

ooshlablu:
There is also an issue that popped up with instruments along with this issue. Instruments are no longer able join the game, some text pops up stating "Push Green to join".
I'm not sure when the instrument issue popped up, but I did notice the issue of the album cover disappearing along with the instrument issue when I tried to run latest master build last night.

PR #950, opened 2023-11-26
possible fix of black cover

twollgam:
### What does this PR do?

Possible bugfix of #949 

### Closes Issue(s)

#949


berarma:
In my laptop with Wayland the issue didn't happen. In my desktop computer it's working well with this fix.

@Baklap4 Do your covers turn empty after playing and exiting the song or are they empty from the start?

Baklap4:
These were empty from the start, but yesterday they still had an image... probably something weird on my system (i did delete all the cache and database.xml, but no dice)

PR #952, opened 2023-11-27
Add editorconfig file

berarma:
This file can be read by many editors and they adjust their settings to the project.

It makes it easier switching between projects.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
```
Baklap4, line 8: i think this should be spaces?

Baklap4 (changes requested):

.editorconfig
```diff
@@ -0,0 +1,10 @@
...
+insert_final_newline = true
+trim_trailing_whitespace = true
```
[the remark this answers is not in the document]
berarma: I think this doesn't affect the code, only the visualization. That's why I left it out as a personal preference.

`tab_width` doesn't need to be set since it already defaults to the same value as indent_size.

Baklap4:
Also we can use https://pre-commit.com/ to force this upon commit for a config see for example: https://github.com/UltraStar-Deluxe/format/blob/main/.pre-commit-config.yaml

.editorconfig
berarma: Ehh... no problem, except that the code is currently indented with tabs.

Baklap4:
Bash only would be the current open pr i guess

.editorconfig
Baklap4: question is asked in discord :)

Baklap4:
Might be nice to extend the editorconfig with ide specific styles to match them
for C++ Visual Studio 2022 it can be found here: https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019&viewFallbackFrom=vs-2022

IDE's to include: Visual Studio 2022 (windows), XCode  (mac), ..... on linux

berarma:
> Might be nice to extend the editorconfig with ide specific styles to match them for C++ Visual Studio 2022 it can be found here: https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019&viewFallbackFrom=vs-2022
> 
> IDE's to include: Visual Studio 2022 (windows), XCode (mac), ..... on linux

Please, send me the settings. I don't use any of those editors. There's a lot of settings there and I don't know which ones would you want to use.

Baklap4:
> > Might be nice to extend the editorconfig with ide specific styles to match them for C++ Visual Studio 2022 it can be found here: https://learn.microsoft.com/en-us/visualstudio/ide/cpp-editorconfig-properties?view=vs-2019&viewFallbackFrom=vs-2022
> > IDE's to include: Visual Studio 2022 (windows), XCode (mac), ..... on linux
> 
> Please, send me the settings. I don't use any of those editors. There's a lot of settings there and I don't know which ones would you want to use. We can also do it later in other PRs.


Scratch it for now, a lot of ide's only support the basic tabs/endofline stuff not the more style things like braces and spaces between items etc. So this can be skipped

.editorconfig
berarma: Done.

.editorconfig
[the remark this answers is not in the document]
berarma: This is done too. Since now indent uses spaces, indent_size is important.

Issue #937, opened 2023-11-11
Text colors in asian style have too less constrast in performous+ theme


twollgam:
### Performous & operating system versions

Performous 1.3+ alpha
OS: Windows 11

### What is the problem?

Other than the default theme the text colors in asian style of performous+ theme differ not enough.

marwin89:
@twollgam  thanks for QA 👋 , I fixed this in Pullreq:
- #932 

Is this what you meant? Another color for highlight lyrics? 

![Screenshot-lyrics-colored](https://github.com/performous/performous/assets/63517874/a080a585-61a3-4893-8b07-0339293642d7)
BTW: the fonts don't get filled well, there's still a white pixeling. Don't know why. 🤔

Baklap4:
In discord this was also mentioned:

> would it be possible to get more stroke color feedback on the singing syllables? currently the letters are white, with a light-grey fill-stroke and that's very hard to read on bright whiteish backgrounds (which a lot of video's are)

Some suggestions from the event were also to fix the outline to be more prominent (the 1px black outline might be a bit too easy to miss)

PR #964, opened 2024-01-20
Draft: Performous+ Theme - adjust lyrics highlight color, stroke opacity and width

BWagener:
### What does this PR do?

* slight adjustments to the wonderful Performous+ theme
  * changes the lyric text highlight color to the same as in the default theme
  * set stroke opacity for all lyrics texts to 1
  * set stroke width for all lyric texts to 1.5

This also slightly reduces the odd white color between the outline and highlighted lyric color.

### Closes Issue(s)

Addresses a point mentioned in #937 https://github.com/performous/performous/issues/937#issuecomment-1812446174. The lyrics' black outline is a bit thin which can make the lyrics hard to read on light backgrounds

### DIFF
--- a/data/themes/performous+ (Black)/sing_lyricscurrent.svg
+++ b/data/themes/performous+ (Black)/sing_lyricscurrent.svg
@@ -63,10 +63,10 @@
   <text
      xml:space="preserve"
      id="text3481"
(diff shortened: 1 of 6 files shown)


marwin89 (approved):

Baklap4:
Merge will wait till 1.3.1 is released, but after that we can probably merge it right away :)

Thanks for the PR!

Baklap4:
Hmm i've changed the base to 1.3.1 instead of master so it gets included within the upcoming 1.3.1 release.

PR #965, opened 2024-01-26
Draft: Performous+ Theme - adjustments to notelines.svg (fixes invisible note labels)

BWagener:
Currently note labels are invisible for me

Before

image

After

image

This also makes the vertical line marking the current note to sing outlined so it's visible on any background.


Baklap4:
Hmm i've changed the base to 1.3.1 instead of master so it gets included within the upcoming 1.3.1 release.

Baklap4 (approved):
LGTM

PR #971, opened 2024-02-17
Speaker selector first

twollgam:
### What does this PR do?

The PR changes the sequence of the devices in the audio device screen. The speaker is now the first device. Before it was the last one.

Cause the speaker is the device that is most often to be configured the new sequence increase user comfort.

### Closes Issue(s)

None.

### Motivation

I have to configure the speaker device most often and use never more than 2 mics so the new sequence increases my user comfort.

### Testing
For the user there should nothing have changed but the position of the speaker selector that must be on the left side now.
Attention to the colors. They have been refactored but must not have been changed.


game/screen_audiodevices.cc
```diff
@@ -13,10 +13,27 @@ namespace {
 	static const int unassigned_id = -1;  // mic.dev value for unassigned
 	static const float yoff = 0.18f; // Offset from center where to place top row
 	static const float xoff = 0.45f; // Offset from middle where to place first column
+	auto const colorNames = std::vector<std::string>{"out=", "blue", "red", "green", "yellow", "fuchsia", "oranage", "purple", "aqua", "white", "gray", "black" };
```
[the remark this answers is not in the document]
twollgam: Yes, you are right. I did a refactoring to avoid the multiple storing of the color names. 👍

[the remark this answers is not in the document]
twollgam: Unit tests are added too. 😃

testing/common.hh
```diff
@@ -13,8 +13,10 @@ static const float pi2 = pi * 2.f;
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
 
+using ::testing::_;
 using ::testing::Contains;
 using ::testing::ElementsAre;
+//using ::testing::FieldsAre;
```
Baklap4, line 19: Maybe os/compiler check so we don't have to comment things out/enable them

Baklap4 (changes requested):

testing/microphones_test.cc
```diff
@@ -0,0 +1,71 @@
...
+    EXPECT_NEAR(0.5f, sut.g, 0.0001f);
+    EXPECT_NEAR(0.5f, sut.b, 0.0001f);
+    EXPECT_EQ(1.0f, sut.a);
+}
+
+TEST(UnitTest_Microphones, getMicrophoneConfig_sequence) {
+    auto const sut = getMicrophoneConfig();
+
+    EXPECT_THAT(sut, ElementsAre(
+        //FieldsAre("blue", _), // not available on all platforms
```
twollgam, line 58: Found this. @Baklap4 remove this too?

Baklap4: Yes i'd think so yes :)

testing/common.hh
twollgam: Removed it.

testing/microphones_test.cc
twollgam: Removed it

Baklap4 (approved):
Looking good, tested it and saw the speaker icon on the front (left side)
Might be good to check this on Linux / Mac aswell

twollgam:
@ooshlablu would be nice if you could test this one on linux (and mac?).

BWagener:
I tested the AppImage on the Steam Deck. No issues there.

PR #974, opened 2024-02-28
Replace boost-functions

twollgam:
### What does this PR do?

Remove more boost functions by std:: or own replacements to get more independent from boost libs.

### Closes Issue(s)

None, but related to #39 .

### DIFF
--- a/testing/utiltest.cc
+++ b/testing/utiltest.cc
@@ -1,4 +1,4 @@
-#include <gtest/gtest.h>
+#include "common.hh"
 
 #include "game/util.hh"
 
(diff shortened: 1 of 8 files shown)


testing/utiltest.cc
```diff
@@ -1,44 +1,78 @@
...
+    EXPECT_EQ(0.0, smoothstep(-1., 1., -2.0));
+    EXPECT_EQ(1.0, smoothstep(-1., 1., 2.0));
+}
+
+TEST(UnitTest_Utils, format) {
+    auto const time = std::chrono::seconds(7260);
+
+    // next two tests are not working when compiled with mingw. see https://sourceforge.net/p/mingw-w64/bugs/793/
+    //EXPECT_EQ("01/01/70 02:01", format(time, "%D %R"));
+    //EXPECT_EQ("02:01:00 1970-01-01", format(time, "%T %F"));
```
Baklap4, line 38: Maybe we should add a compiler-if-check here so they are exclued for mingw (or we delete these two entirely) 

i don't like commented code which does nothing :P

Baklap4 (changes requested):

Baklap4 (approved):
Let's get this tested, and see if anything weird shows up if not, the review is approved! @ooshlablu

BWagener:
Tested on windows, looks good to me.

Baklap4 (approved):

PR #975, opened 2024-03-01
Added support for Comment and ProvidedBy tags so we can distinguish m…

### LINKED ISSUE #961: Support for `VERSION` tag
### Suggestion

Within the Ultrastar community we've decided upon a common format. This can be found on [our website](https://usdx.eu/format).
One of the new tags introduced for compatibility reasons is the `#VERSION` tag.

We can use this tag to determine which features we want to use as dictated by the source.

As seen on the website:
> The version attribute helps the game and apps to understand how uptodate a song file is and how it should be treated. Ideally all txt files will be upgraded to the latest version. The value are version numbers according to semantic versioning. Song Editors, Managing tools or txt-hostings should set this automatically.

### Todo

- [ ] Implement the version tag. Details about each version can [be found here](https://github.com/UltraStar-Deluxe/format/issues/13#issuecomment-1819253742) and to have a list of tags available in each version [can be found here](https://docs.google.com/spreadsheets/d/13fj_wtRHbSjaOaS1ehmCogoeFV9MyqY1eeiqkrJfhDA/edit?usp=sharing)

### Use case

* Gradually introduce new features which come in new format versions
* Have a standard way to determine what features to enable for the specific txt in question instead of best effort

Our versions will start with this structure
0.1.0 -> Bare Minimum, Minimum vial product (mvp)
0.2.0 -> Introduce the tags we currently have
0.3.0 -> Deprecate old # tags we don't wanna use (duplicates, technical not nessecary)
1.0.0 -> Removal of deprecated tags and final list of currently used tags (2023)
1.1.0 -> First new features we want to introduce, deprecated #mp3 and chaotic time units (2023)
~1.2.0 -> more new features (2024)
2.0.0 -> Removal of #MP3 and introduce unified time units (=doing breaking changes) (2025)

also linking the vocaluxe issue: https://github.com/Vocaluxe/Vocaluxe/issues/628 just for reference

Baklap4:
### What does this PR do?

Introduces a new field `PROVIDEDBY`
This helps in distinguishing where the mapped file came from and which program made it at which version.
Aside from this `#COMMENT` is sometimes used which is now also supported

### Closes Issue(s)

closes #961

### Motivation

I'm trying to get a couple new standardized fields into the ultrastar format.
This allows hosting websites to add their name to song files whilst creator tools like composer add their specific version and appname.

This way everyone knows where the txt came from and which program is used to create it.

Since the varying sources of which you can download songs from differ in quality this is a nicely added bonus to see where it came from.

Might be a niche for just me, but if everyone opts in it'd make for a way better system and maybe even a newly introduced system which combines all the sources out there.... (that's my endgoal)

### Testing

You can test this by adding the following tags:
```
#PROVIDEDBY:https://usdb.animux.de
#COMMENT:Some comment
```

All tags are opt in. But if everyone plays by the rules every txt will have these fields eventually

The tags are shown within the webserver-ui: Playlist and search screen.
All extra tags are seperated by the `|` symbol. And the popup if you click a song also mentions the tags + content if they're there.

### Update
I've recently added a couple of new fields and fallback keys these are
`#VERSION` -> Defines the version of the mapped song. This version applies to the Format project https://github.com/UltraStar-Deluxe/format logic which changes through versions has not been added yet as everything is at this moment backwards compatible. Starting with version 2.0.0 we need to add specific handlers for the different versions.
`#YEAR` -> Defines the year of when the song came out
`#AUDIO` -> Is going to replace the `#MP3` header
`#INSTRUMENTAL` -> Aside from Audio we can now specify the instrumental track within the same file. This allows us to mix them in software for several usecases
`#END` -> Specifies the time in milliseconds from the beginning of the audio file at which the song ends. This value can be used to stop playback before any silence or outro at the end of the audio file
`#TAGS` -> A list of arbitrary tags we can help searching for songs

Replaced `GOLDEN2` enum value to `GOLDENRAP` as this is what the G stands for

### DIFF
--- a/data/themes/default/www/js/playlist-song-modal-body.js
+++ b/data/themes/default/www/js/playlist-song-modal-body.js
@@ -136,6 +136,28 @@ $(function () {
     }
 
     /*
+        If our song contains a ProvidedBy show it.
+        Else hide the complete div.
+    */
+    if (songObject.ProvidedBy.length > 1) {
+        $("#playlist-song-modal-body #modal-providedby-div").show();
+        $("#playlist-song-modal-body #modal-providedby-value").text(songObject.ProvidedBy);
+    } else {
+        $("#playlist-song-modal-body #modal-providedby-div").hide();
+    }
+
+    /*
+        If our song contains a comment show it.
+        Else hide the complete div.
+    */
+    if (songObject.Comment.length > 1) {
+        $("#playlist-song-modal-body #modal-comment-div").show();
+        $("#playlist-song-modal-body #modal-comment-value").text(songObject.Comment);
+    } else {
+        $("#playlist-song-modal-body #modal-comment-div").hide();
+    }
+
+    /*
         Click event handlers to either change a position of the song or remove the song.
     */
     $("#playlist-song-modal-body #modal-move-song-up").click(function () {
--- a/data/themes/default/www/js/playlist.js
+++ b/data/themes/default/www/js/playlist.js
@@ -22,10 +22,15 @@ $("#refresh-playlist").click(function () {
 
             $.each(database, function (iterator, songObject) {
                 totalTime += songObject.Duration + timeout;
+                var songMeta = "";
+                songMeta += songObject.Language.length > 0 ? " | " + songObject.Language : "";
+                songMeta += songObject.Edition.length > 0 ? " | " + songObject.Edition : "";
+                songMeta += songObject.ProvidedBy.length > 0 ? " | " + songObject.ProvidedBy : "";
+                songMeta += songObject.Comment.length > 0 ? " | " + songObject.Comment : "";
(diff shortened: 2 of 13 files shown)


Baklap4:
Updated the PR to fix conflicts and add some more tags (listed in `###UPDATE`)

ooshlablu:
I tested this and didn't notice any difference in gameplay, for the files I have at least. Most of what I have is in FoFix format, so I couldn't verify everything. Muting the vocals with `crtl` + `v` definitely works. :+1: from my side, but would be good if someone with different formats could give it a run-through.

Baklap4:
@Lord-Kamina @twollgam if either one of you can review this i can update the format guys to tell we support the latest, we're currently not doing much with it. But i have some pr's planned using them in the near future

PR #976, opened 2024-03-23
Add guitar tuner

twollgam:
### What does this PR do?

This pr adds a guitar tuner screen to performous. It is an alpha feature that must be enabled in the configuration.

It has two objectives:

1. Tune a six string guitar ;-) As I want to add real guitars as instruments in the future this is a very basic start.
2. Test the current note/tone detection.

### Closes Issue(s)

None.

### Motivation

I like to add real guitars as instruments and this is a very basic start. 

### More

Nothing yet.

### Additional Notes

It seems that my impression that deep notes are not exactly detected will be proved.


Baklap4:
I think this is still a draft and doesn't need review yet?

Issue #942, opened 2023-11-19
Add marking songs defect ingame


twollgam:
### Suggestion

Extend the menu which opens at pressing ESC while playing.

### Use case

Downloaded songs with usdb syncer are sometimes not usable. In this case it would be nice to mark them unusable. That gives an easy way to filter or sort them out.

marwin89:
This is a good idea. 👋 

- Did you know that USDB SYNCER will use "Song-Checked"-Marker in Future? So you can filter songs that definitely work. 💡
- QA Team of USDB is building up right now and will start using song-check marker again. ✔️🎵

Baklap4:
What would the technical details be @twollgam ?
Where are we gonna store this information

It might be usefull for the #240 aswell

BWagener:
I really like this feature.

Another use case which is similar and might warrant consideration is having the same song multiple times. I'd love to be able to just let the game ignore all but one version of the same song.

An option to disable a song would be great for that. The only difference to the feature proposed in this issue would be the intent. 

The songs I'd like to disable using this feature would be songs that are duplicate, but don't need fixing.

Baklap4:
This one is closed right as #978 is already merged into master? (if not please reopen)

BWagener:
I suggest rethinking the current implementation.

Currently the "broken" flag is saved as a song attribute in the database.

This is problematic in cases where a song is present multiple times from different sources. All versions of the song regardless of source are considered broken as there's no information which file was originally marked as broken.

My suggestion would be to specifically mark a file as broken and ignore such files. A separate list could be saved to the db.

PR #978, opened 2024-03-25
add new option in song screen menu that exists and mark the song broken

twollgam:
### What does this PR do?

Add new option in song screen menu that exists and mark the song broken.
The information is stored in the database and can later be used for song filtering

### Closes Issue(s)

#942 

### Motivation

Many songs from usdb are not synchronous so the are not performable. This is the first step to be able to filter out those songs.  Later #868 will be extended to apply a filter for broken songs.


Baklap4:
Is this to be reviewed? Or is it still in draft phase?

HetorusNL:
The implementation looks good to me. I would, however, suggest to rename 'Abort' to something more explicitly suggesting that this is 'marking the song as broken'.

Don't know if I'm necroing something, but the whole tabs vs spaces, what's the intended style for Performous? And if there is a style, it might be a good idea to (auto?) format all files and get it over with (and at the same time invalidating all PR's if github can't handie this..)

twollgam:
> I would, however, suggest to rename 'Abort' to something more explicitly suggesting that this is 'marking the song as broken'.

Any suggestions?

1. Abort cause defect
2. Abort defect song
3. Mark song defect and quit
4. ...

twollgam:
> Don't know if I'm necroing something, but the whole tabs vs spaces, what's the intended style for Performous? And if there is a style, it might be a good idea to (auto?) format all files and get it over with (and at the same time invalidating all PR's if github can't handie this..)

You mean indent? There is a PR for rejecting wrong formatted files locally. I think github gives no support server side. But a pipeline could fail for wrong formatted files.

HetorusNL (approved):

Issue #7, opened 2012-11-18
Kiosk/arcade mode and related functions


Tronic:
Add features needed for running Performous in embedded systems such as arcade cabinets.
- Kiosk/arcade mode (disable exiting the game and limit other functions, auto-reset after idle for a certain period)
- Allow playing a single song given as command-line argument (no menus)

We are waiting for a commercial sponsor to pay for the development.

PR #980, opened 2024-04-03
Kiosk mode -> limit playlist and boot straight to songs.

nieknooijens:
…imit playlist size

<!--
PLEASE READ THIS MESSAGE.

### What does this PR do?

Adds the kiosk mode menu with settings allowing you to limit the playlist size and boot straight to the song screen.

### Closes Issue(s)
none

### Motivation

Too little personell at tomo conventions, so we're automating the karaoke booth.
To avoid abuse of the playlist system we limit the maximum amount of songs.
We also just want to connect power, and have the system boot straight into karaoke mode so any idiot can set it up.
Starting the webbrowser on a different screen will be done by shell script.

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -36,6 +36,11 @@ to save the current settings to XML.
 			<short>Paths</short>
 			<long>Setup song and data paths.</long>
 		</entry>
+		<entry name="kioskmode">
+			<short>Kiosk mode</short>
+			<long>Setup automatic song play for event displays</long>
+		</entry>
+		
 	</menu>
 
 	<!-- Gameplay preferences -->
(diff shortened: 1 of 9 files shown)


nieknooijens:
windows build fails in portaudio so that's outside of my PR.

twollgam:
No, it is in main.cc:
```
D:\a\performous\performous\performous\game\main.cc(161): error C3861: 'sleep': identifier not found
```
Do you really need a sleep? If yes use `std::this_thread::sleep_for`.

Baklap4:
Possible closes #7 ?

twollgam:
@nieknooijens my question: why sleep exactly one second? The problem with the fixed duration is that on some systems it is to long on other it is too short. So what happens if you remove the sleep? Or do we need to wait that all songs are loaded?

nieknooijens:
> @nieknooijens my question: why sleep exactly one second? The problem with the fixed duration is that on some systems it is too long on other it is too short. So what happens if you remove the sleep? Or do we need to wait until all songs are loaded?

I wait until at least one song has loaded. So untill the songs array isn't empty anymore. If it is empty, I'll wait a second and try again. The reason for the sleep is that it makes the process yield rather than busy-wait. So if you have a single CPU, you give it time to switch threads instead of busy waiting.

nieknooijens:
> Possibly closes #7 ?

Probably not all of it, but it's a start.

twollgam:
@nieknooijens ok I understand. Then I think you should use 10ms which will respond more smooth.

Baklap4:
Yes please :D if @ooshlablu can test this, that'd be great

twollgam (changes requested):
There are some minor changes to do. All in all it is good 👍

ooshlablu:
@nieknooijens Can you rebase this on current master to get the Ubuntu 24.04 packages?

nieknooijens:
so that´s the rebase...

PR #989, opened 2024-05-01
Draft: optimize most sung and highest score sorting

BWagener:
### What does this PR do?

Note: for now this is just a draft as the highest score sorting is not done yet and I would like the pipelines run and test around.

This is my attempt to improve sorting performance for most sung and highest score sorting options.

Currently with a large library using these sorting options is very slow.

### Motivation

Have all sorting options be efficient.


game/screen_sing.cc
```diff
@@ -261,7 +261,6 @@ void ScreenSing::instrumentLayout(double time) {
 
 void ScreenSing::activateNextScreen()
 {
-	m_database.addSong(m_song);
```
BWagener, line 264: Looking for advice why this was being called

game/hiscore.cc
```diff
@@ -120,17 +110,27 @@ void Hiscore::load(xmlpp::NodeSet const& nodes) {
...
 void Hiscore::save(xmlpp::Element *hiscores) {
-	for (auto const& h: m_hiscore) {
-		xmlpp::Element* hiscore = xmlpp::add_child_element(hiscores, "hiscore");
-		hiscore->set_attribute("playerid", std::to_string(h.playerid));
-		hiscore->set_attribute("songid", std::to_string(h.songid));
-		hiscore->set_attribute("track", h.track);
-		hiscore->set_attribute("level", std::to_string(h.level));
-		hiscore->set_attribute("unixtime", std::to_string(h.unixtime.count()));
-		hiscore->add_child_text(std::to_string(h.score));
+	for (auto const& [song_id, song_hiscores] : m_hiscore_map) {
```
BWagener, line 113: we could sort by timestamp, song and/or player id before saving.
As the map is unordered entries in the db will be shuffled (stable order, just not by song or player id) right now.

game/songitems.cc
```diff
@@ -36,73 +42,67 @@ void SongItems::save(xmlpp::Element* songs) {
 
 SongId SongItems::addSongItem(std::string const& artist, std::string const& title, std::optional<SongId> _id) {
 	SongItem si;
+
+	if (_id.has_value() && m_songs_map.find(_id.value()) != m_songs_map.end()) {
+		si = m_songs_map.at(_id.value());
+	}
+
 	si.id = _id.value_or(assign_id_internal());
```
BWagener, line 50: assign_id_internal() would always be evaluated here, as value_or can't short circuit

game/hiscore.cc
```diff
@@ -40,45 +42,41 @@ void Hiscore::addHiscore(HiscoreItem&& item) {
 		throw std::runtime_error("No track given");
 	if (!reachedHiscore(item.score, item.songid, item.level, item.track))
 		return;
-	m_hiscore.insert(std::move(item));
-}
 
-Hiscore::HiscoreVector Hiscore::queryHiscore(std::optional<PlayerId> playerid, std::optional<SongId> songid, std::string const& track, std::optional<unsigned> max) const {
```
BWagener, line 46: While this method is very generic and could be used from many contexts, it was only being used from a single one. Accessing one of the maps was the more efficient alternative so I replaced the usage and deleted this function.

game/players.cc
```diff
@@ -66,7 +66,7 @@ void Players::addPlayer (std::string const& name, std::string const& picture, st
 	pi.name = name;
 	pi.picture = picture;
 
-	pi.id = id.value_or(assign_id_internal());
```
BWagener, line 69: assign_id_internal() would always be called, as value_or doesn't short circuit.

game/songitems.hh
```diff
@@ -32,13 +32,6 @@ struct SongItem
 	std::string artist;
 	std::string title;
 
-	/** This shared pointer is stored to access all song
-	  information available.
-	  E.g. the full artist information can be accessed using
-	  this pointer.
-	 */
-	std::shared_ptr<Song> song;
-
```
BWagener, line 41: it was not being used anywhere

game/songs.cc
```diff
@@ -367,9 +373,17 @@ void Songs::filter_internal() {
 		std::shared_lock<std::shared_mutex> l(m_mutex);
 		SongCollection(m_songs.begin(), m_songs.end()).swap(m_filtered);  // Invalid regex => copy everything
 	}
+	prof("filtered");
 	sort_internal();
 }
 
+// TODO: determine appropriate time and location to call this function
```
BWagener, line 380: Looking for help on this

game/songitems.cc
BWagener: assign_id_internal() would always be evaluated here, as value_or can't short circuit

game/songitems.cc
```diff
@@ -36,73 +42,54 @@ void SongItems::save(xmlpp::Element* songs) {
 
 SongId SongItems::addSongItem(std::string const& artist, std::string const& title, std::optional<SongId> _id) {
 	SongItem si;
-	si.id = _id.value_or(assign_id_internal());
```
BWagener, line 39: assign_id_internal() would always be evaluated here, as value_or can't short circuit

twollgam:
@BWagener there are type issues (look at the mac build). I think somewhere the song id is an `int` but should be `SongId`.

BWagener:
> @BWagener there are type issues (look at the mac build). I think somewhere the song id is an `int` but should be `SongId`.

Ah thanks, I was mistakenly assuming it's an unrelated issue as the other builds were successful.

Would it be okay to explicitly cast them to SongId? I intentionally defined the id as a signed int on the Song class to handle missing ids. With the changes made in this PR it is guaranteed every loaded song has an ID, I just needed to track the state of a new song to check it [here](https://github.com/performous/performous/pull/989/files#diff-a5a04d3e4f566fbc805534bb50d9f5004e5744a5b6529459e5cab0d38d2a04a1R59)

An alternative would be using an optional<SongId> instead, what do you think?

twollgam:
IIRC I started with a value for missing/undefined ids but it was refactored away with a std::optional. Even if that was not my own idea I would suggest this solution. Reason: You tell (with code) where an id is defined and where it could be undefined. That does not work if you simply use an `int` or `SongId`. This is where in the past comments were used.

ooshlablu:
@BWagener Can you rebase this on current master to get the Ubuntu 24.04 packages?

BWagener:
> @BWagener Can you rebase this on current master to get the Ubuntu 24.04 packages?

@ooshlablu done

PR #990, opened 2024-05-02
Fix issues on MacOS with ffmpeg

twollgam:
This PR fixes issues on MacOS with ffmpeg. 

Some time ago the homebrew package for ffmpeg@4 does not provide the av*-libs anymore for cmake projects. Building ffmpeg from source solves the problems.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -33,7 +33,6 @@ jobs:
         cmake
         cpprestsdk
         dylibbundler
-        ffmpeg@4
         fftw
         fmt
         fontconfig
@@ -78,6 +77,12 @@ jobs:
         id: install_deps
         run: |
            brew install $MACOS_DEPS
+
+      - name: Install dependency FFMPEG from source
+        id: install_ffmpeg_deps
+        run: |
+           brew install ffmpeg@4 --build-from-source 
+
       - name: Setup Python
         id: setup_python
         uses: actions/setup-python@v5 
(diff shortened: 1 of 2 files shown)


ooshlablu (approved):
LGTM

.github/workflows/macos.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         cmake
         cpprestsdk
         dylibbundler
-        ffmpeg@4
+        ffmpeg@6
```
Baklap4, line 36: Can't we just remove this entirely as we're building from ffmpeg soure?

Lord-Kamina:
Walk ne through what's going on here? This looks messy AF

.github/workflows/macos.yml
twollgam: ffmpeg@6 is pulled in by dependencies ...

But I think someone (@ooshlablu ;-) ) takes the changes and integrate the needed into his pr.

twollgam: @Baklap4 so yes, we can remove the entry :)

.github/workflows/macos.yml
```diff
@@ -78,6 +78,12 @@ jobs:
         id: install_deps
         run: |
            brew install $MACOS_DEPS
+
+      - name: Install FFMPEG Dependencies
```
twollgam, line 82: The name is not matching. Better `Install dependency FFMPEG from source`?

ooshlablu (approved):

Issue #992, opened 2024-05-05
Songs with sequential collatable artist or title are duplicated in db on every load


BWagener:
### Performous & operating system versions

Performous: master [41b7637](https://github.com/performous/performous/commit/41b7637aed8abbb8b3b7d257179b630a616db4e7)
Operating system: Windows 11 Pro N and Steam OS 3.5.19

### What is the problem?

When a song's artist or title starts with two or more "collatable" words (words defined in `config[game/sorting_ignore]`) it is duplicated in the DB whenever it is loaded from cache.

### Steps to reproduce

Change a song's artist or title to fit the criteria above, some examples:

- Alejandro Sanz - **A la** primera persona
- Ed Sheeran - **The A** Team
- **La La** Land (Emma Stone) - Audition (The Fools Who Dream)

Start up the game, close it, start it again, close it.

First save:
```
    <song id="0" artist="Alejandro Sanz" title="la primera persona,A"/>
    <song id="1" artist="Ed Sheeran" title="A Team,The"/>
    <song id="3" artist="La Land (Emma Stone),La" title="Audition (The Fools Who Dream)"/>
    ...
```

Second save:
```
    <song id="0" artist="Alejandro Sanz" title="primera persona,A,la"/>
    <song id="1" artist="Ed Sheeran" title="Team,The,A"/>
    <song id="3" artist="Land (Emma Stone),La,La" title="Audition (The Fools Who Dream)"/>
    ...
    <song id="10" artist="Alejandro Sanz" title="la primera persona,A"/>
    <song id="11" artist="Ed Sheeran" title="A Team,The"/>
    <song id="13" artist="La Land (Emma Stone),La" title="Audition (The Fools Who Dream)"/>
    ...
```

Third save:
```
    <song id="0" artist="Alejandro Sanz" title="primera persona,A,la"/>
    <song id="1" artist="Ed Sheeran" title="Team,The,A"/>
    <song id="3" artist="Land (Emma Stone),La,La" title="Audition (The Fools Who Dream)"/>
    ... 
    <song id="10" artist="Alejandro Sanz" title="primera persona,A,la"/>
    <song id="11" artist="Ed Sheeran" title="Team,The,A"/>
    <song id="13" artist="Land (Emma Stone),La,La" title="Audition (The Fools Who Dream)"/>
    ...
    <song id="20" artist="Alejandro Sanz" title="la primera persona,A"/>
    <song id="21" artist="Ed Sheeran" title="A Team,The"/>
[...]

PR #993, opened 2024-05-05
save unmodified song artist and title to db

BWagener:
### What does this PR do?

save unmodified song artist and title to db instead of collated values

### Closes Issue(s)

#992


twollgam (dismissed):
LGTM

PR #997, opened 2024-05-15
Initial work to incorperate use of Replay Gain

kt--:
_Draft_ RFC request only.

I'm trying to use the existing Replay Gain (Ref: https://en.wikipedia.org/wiki/ReplayGain ) functions in ffmpeg to normalise the audio volume in performous.

### What does this PR do?

Implements loading and logging of any Replay Gain meta information loaded from audio files.  Files might not have any Replay Gain information in them.  To add it, you can use a tool like `loudgain`.  

### Closes Issue(s)

This PR does not close any issues.

### Motivation

When one has audio from different sources, more often than not, the base-level volume of the audio is different.  These differences are prevalent even in professionally mastered audio recordings simply due to their jurisdiction.  The long-term idea is to somewhat normalise the "loudness" of the audio track, such that continual _manual_ volume adjustments don't need to be made while enjoying performous.

I've raised this Draft PR as a discussion point, because I don't understand how the volume setting code works in performous.  I have implemented Replay Gain normalisation before, but I need some pointers on how performous handles track volumes so I can implement it in a design-sympathetic manner.

**UPDATE 2024-06-05**

EDIT AGAIN: I've broken it all trying to fix the whitespace changes.   (angry-swearing-emoji)

EDIT AGAIN:  Gah!  I've got heaps whitespace changes, maybe I'll just close this and re-merge.

After prompting from @twollgam, I nuked the use of libav's filter, and simply applied the replay gain factor in the correct spot[1].  It seems to be working.  I will test it over the next week or so, and if I'm happy take it out of `Draft`.

As the data from the read from the ffmpeg input stream, it is stuffed into the audio buffer.  During this phase the adjustment is made by multiplying the gain to each sample.  

So far, its making loud songs softer, and soft songs louder. 

[1] _Possibly_ correct spot ;)

### DIFF
--- a/game/ffmpeg.hh
+++ b/game/ffmpeg.hh
@@ -22,6 +22,7 @@ extern "C" {
   struct AVCodecContext;
   struct AVFormatContext;
   struct AVFrame;
+  struct AVStream;
   void av_frame_free(AVFrame **);
   struct SwrContext;
   void swr_free(struct SwrContext **);


game/ffmpeg.cc
```diff
@@ -305,7 +305,7 @@ bool FFmpeg::readReplayGain(const AVStream *stream)
 	// assert(stream)
 	m_replaygain = 0.0;
 	if (stream != nullptr) {
-		int replay_gain_size = 0;
+		size_t replay_gain_size = 0;
```
twollgam: Seams to be wrong cause ffmpeg documentation says that it is an `int`.

PR #1000, opened 2024-06-11
Read any Replay Gain meta-information, use it to normalise the audio "loudness".

kt--:
NOTE: _I closed my other pull request since it was simply quicker to re-merge the changes manually (with `meld`) rather than muck around in the leavings of `git`.  I had managed to accidentality pull in a huge whack of white-space changes that needed to disappear.  Please accept my apologies._

### What does this PR do?

This change implements loading and use of any [Replay Gain](https://en.wikipedia.org/wiki/ReplayGain) meta information found in the audio files. Files might not have any Replay Gain information in them. To add it, you can use  tools like [loudgain](https://github.com/Moonbase59/loudgain) or [mp3gain](https://mp3gain.sourceforge.net/).  One can easily run `loudgain` over all of Performous' audio tracks using the UNIX "find" command, with the --exec option.

Using the patch generally results in soft audio being made louder and vice-versa.  But don't expect that all music becomes "perfectly" normalised.  A particularly loud piece of music might still be perceived as quite loud - but not enough to warrant manual volume-knob intervention.

It's been tested for many weeks on hundreds of pure-audio and audio-with-video files.

### Closes Issue(s)

Does not close any issues.

### Motivation

When one has audio from different sources, more often than not, the base-level volume of the audio is different. These differences are prevalent even in professionally mastered audio recordings simply due to their jurisdiction. The long-term idea is to somewhat normalise the "loudness" of the audio track, such that continual manual volume adjustments don't need to be made while enjoying Performous.

#### How to Test

- Set your output volume to whatever your normal playback volume is
- Identify some music items in your Performous song list that are either too soft, or too loud
- Play the item in Performous, noting how loud/soft it is
- Use `loudgain` to create the meta-information in the audio file (see below)
- Play the item again in Performous
  - Note the debug log when the gain is found: ffmpeg/debug: readReplayGain() - GAIN IS [-3.73] dB, Linear Gain is [0.65]
  - Listen again to the volume of the track, soft tracks should be made louder, loud tracks softer.
- I found that _generally_ the patch lowers the volume of most tracks sourced from video-viewing websites, since these places normalise the sound to a louder base-level.
- For very formal testing, I guess you could manually adjust the audio volume in Audactiy, and verify it's corrected.

There is a new configuration item in [Audio] that turns the feature on/off.

### More

The libav function that fetches the gain data `av_stream_get_side_data()` is marked as deprecated.  Probably it will need to be migrated to the C++(?) non-deprecated version when all the rest of the API is done.

### Additional Notes

To quickly add Replay Gain meta information, an efficient way is to install "loudgain" (see above), then execute it for every `.mp3` in the music archive.  I've never had an issue with it, but maybe make a backup first.  Best kick this off, watch for a few seconds, then go have a long tea break.

```
$ find . -iname \*mp3 -exec loudgain -q -I3 -S -L -a -k -s e {} \;
$ find . -iname \*m4a -exec loudgain -q -a -k -s e {} \;
```

(Note: last time I looked there was a bug with loudgain and m4a files, but also a patch mentioned in the github comments)

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -24,7 +24,7 @@ jobs:
   # Create the MacOS dmg
   MacOS_Packages:
     name: Create MacOS dmg
-    runs-on: macos-12
+    runs-on: macos-13
     env:
       MACOS_DEPS: >-
         aubio


game/ffmpeg.cc
```diff
@@ -301,6 +364,8 @@ AudioFFmpeg::AudioFFmpeg(fs::path const& filename, int rate, AudioCb audioCb) :
 
 double FFmpeg::duration() const { return double(m_formatContext->duration) / double(AV_TIME_BASE); }
 
+double FFmpeg::replayGain() const { return m_replaygain; }
```
twollgam, line 367: Name it `getReplayGain`.

game/ffmpeg.hh
```diff
@@ -49,10 +50,14 @@ class FFmpeg {
 	/// duration
 	double duration() const;
 
+	/// replay gain, in +/- decibels.  Can be zero, and is zero if not defined for the track
+	double replayGain() const;
```
twollgam, line 54: To avoid the comment name it `getReplayGainInDecibel`.

game/ffmpeg.hh
```diff
@@ -134,6 +141,7 @@ class AudioBuffer {
 
 	const unsigned m_sps;
 	const double m_duration{ 0 };
+	const double m_replaygain{0.0};  // units: dB
```
twollgam, line 144: To avoid the comment name it `m_replaygainInDecibel`.

game/ffmpeg.cc
```diff
@@ -277,6 +300,46 @@ FFmpeg::FFmpeg(fs::path const& _filename, int mediaType) : m_filename(_filename)
...
+  *
+  *           Ref: https://en.wikipedia.org/wiki/ReplayGain
+  *                https://www.smithsonianmag.com/smart-news/music-does-get-louder-every-year-835681/
+  *
+  * \note     The libav function av_stream_get_side_data() has been deprecated
+  *           It will need to be updated along with all the others
+  *
+  * \returns  True if a Replay Gain value was found in the stream
+  */
+bool FFmpeg::readReplayGain(const AVStream *stream)
```
twollgam, line 321: Make it void cause the returnvalue is nowhere used.

game/ffmpeg.cc
```diff
@@ -277,6 +300,46 @@ FFmpeg::FFmpeg(fs::path const& _filename, int mediaType) : m_filename(_filename)
...
+{
+    bool rg_read = false;
+    // assert(stream)
+    m_replaygain = 0.0;
+    if (stream != nullptr) {
+        size_t replay_gain_size = 0;
+        const AVReplayGain *replay_gain = (AVReplayGain *)av_stream_get_side_data(stream, AV_PKT_DATA_REPLAYGAIN, &replay_gain_size);
+        if (replay_gain_size > 0 && replay_gain != nullptr) {
+            m_replaygain = static_cast<double>(replay_gain->track_gain);
+            m_replaygain /= 100000.0;   // convert from microbels to decibels
```
twollgam, line 331: If you store micro decibel, name the variable `m_replaygainInMicroDecibel`.

twollgam: But I think it should be stored in decibel. Where you need micro decibel convert it.

twollgam (changes requested):
There are some little changes requested.

game/ffmpeg.cc
twollgam: Yes, I am. `duration()` is stl-style and was widely used in performous. But it is expiring and the `get-` and `set-` notation is used.

ooshlablu (approved):
Tests from my side seem good. I definitely notice a difference in the evenness of the output from multi-track oggs. Sometimes one of the parts would overwhelm the others, like drums being way louder than guitars, and this appears to make that more consistent. I also notice a huge difference when playing tracks that are oggs vs mp3. ogg compared to ogg has a more consistent volume when compared to each other.
I can't really say much about the code, but the tests look good to me. :+1:

Baklap4:
Just coming back from a hiates, but in its current state it'll either normalize the audio file or not (if loudness info is missing) for a given stream if i'm correct. Would be good to set this feature behind a config flag so users can enable it if they know *every* file has the loudness info in it. Else it'll trigger the same behavior still: some songs are louder than others

As i'm using different sources (with auto-updates) i'm not going to run `loudgain` on my whole library. I'd rather mix it with my mixer and compressor then to have the software responsible for that.

For most users however this will be a great feature so i think the config item can be enabled by default (as it'll improve most users) but the more advanced users (with external hardware) will have an option to disable this feature entirely

Baklap4:
Blocked by #1005 for ffmpeg build errors osx (and probably windows too)

PR #1001, opened 2024-06-24
Fix module name in `FindCed.cmake`

abompard:
Without this change I'm getting:

```
-- Checking for module 'ced'
--   Package 'ced' not found
CMake Error at /usr/share/cmake-3.29/Modules/FindPkgConfig.cmake:619 (message):
  The following required packages were not found:

   - ced

Call Stack (most recent call first):
  /usr/share/cmake-3.29/Modules/FindPkgConfig.cmake:847 (_pkg_check_modules_internal)
  cmake/Modules/FindCed.cmake:16 (pkg_check_modules)
  game/CMakeLists.txt:138 (find_package)
```


ooshlablu:
I sing-tested this on Ubuntu 24.04 and didn't notice any differences in game play, which I think is the point. I did not test with instruments yet tho.

:+1: from the Linux side testing

Lord-Kamina:
@abompard I haven't touched the CMake code in a while. Walk me a bit through what's going on here and in your setup.
Initially I thought your PR was wrong because why change `ced` only in that line, but then I see that it's a PkgConfig issue.

This is probably me being dumb and not being used to pkg-config but, how does pkg-config even find CED normally, without a `.pc` file?

abompard:
> @abompard I haven't touched the CMake code in a while. Walk me a bit through what's going on here and in your setup. 

Of course, thanks for responding.

> Initially I thought your PR was wrong because why change `ced` only in that line, but then I see that it's a PkgConfig issue.  This is probably me being dumb and not being used to pkg-config but, how does pkg-config even find CED normally, without a `.pc` file?

It's there, it's just that the file is named `CED.pc`, not `ced.pc`:  `/usr/share/pkgconfig/CED.pc`

Lord-Kamina:
> > @abompard I haven't touched the CMake code in a while. Walk me a bit through what's going on here and in your setup.
> 
> Of course, thanks for responding.
> 
> > Initially I thought your PR was wrong because why change `ced` only in that line, but then I see that it's a PkgConfig issue.  This is probably me being dumb and not being used to pkg-config but, how does pkg-config even find CED normally, without a `.pc` file?
> 
> It's there, it's just that the file is named `CED.pc`, not `ced.pc`: `/usr/share/pkgconfig/CED.pc`

I see. What OS/distro are you on? This is not really relevant to the PR per-se, but do you know where that came from? I'm seeing that upstream doesn't have a pkg-config file. Maybe it'd be a good idea to send them a PR.

Lord-Kamina (approved):

abompard:
Huh, now I can't recall where this file is coming from. I thought it was from upstream, but it's not, actually. I'll get back to you shortly.

abompard:
Turns out I had written this file myself and failed to name it properly. Sorry for wasting your time.

PR #1017, opened 2024-11-07
Font paint order

kt--:
### What does this PR do?

In SVG shapes have three components, Fill, Stroke and Markers.  The default is to paint the Fill then the Stroke (then presumably Markers).   However the order of painting is changeable. For example, such that the Stroke is painted first, next the Fill.

This PR loads and parses the "paint_order" value from the SVG, incorporating it into the Theme.  It allows the SVG to decide the paint order, rather than having it hard-coded.

The benefit of this PR is to allow theme fonts to be configured with extra-large outlines, which can be placed _under_ the text, forming a thick border surrounding the text; making it easier to read.

### Closes Issue(s)

n/a

### Motivation

Black and White backgrounds - both video and still make the reading of Karaoke Font text difficult.  Other applications solve this by using a large matte underlying border around the text.   However changing the Theme font in Performous to use a thick Stroke covered up the font body, and made the text _less_ legible.

### Additional Notes

The existing Theme code is a bit weird in how it works.  I endeavoured to make these changes closely model the existing code, despite what I thought of it.

### DIFF
--- a/game/ffmpeg.cc
+++ b/game/ffmpeg.cc
@@ -325,8 +325,6 @@ void FFmpeg::readReplayGain(const AVStream *stream)
 #else
 		size_t replay_gain_size;
 #endif
-
-	if (stream != nullptr) {
 		const AVReplayGain *replay_gain = (AVReplayGain *)av_stream_get_side_data(stream, AV_PKT_DATA_REPLAYGAIN, &replay_gain_size);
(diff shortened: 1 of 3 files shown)


Lord-Kamina:
This is pretty interested. I am under the impression that there was a reason why we were doing text first, so I'd like to thoroughly test it.

Can you see how it looks when the highlighted text has a different stroke color?

PR #1018, opened 2024-11-24
Fix crash in Joystick test mode.

Lord-Kamina:
### What does this PR do?

Fixes a crash (two, really) in the joystick test mode (`--jstest`)

### Motivation

Accidentally stumbled onto this.

### More


ooshlablu (approved):
Can confirm the crash being shown on the version available from current master, and running this version eliminates the crash and shows the black testing window as it should.

PR #1021, opened 2024-12-05
Refactor filesystem code

Lord-Kamina:
This also came about from the logging refactor; I began writing a formatting template for `fs::path` and noticed the mess it was.

Initially was going to include it in the logging PR but that's very big as-is, so I did my best to cherry pick these changes.

### What does this PR do?

1) Reintegrate the unix and windows implementations of the PathCache, since in effect most of the code was duplicated.
2) Remove some unused code (remnants from an ancient `pathCache` instead of `PathCache`)
3) Minor fix to how macOS paths are constructed.
4) Fix part of the code to find locale files never actually executing due to a `goto` statement.
5) Nuke said `goto` to high hell.
6) Reorder some includes and file paths
7) Remove FindFilesystem and support for `std::experimental::filesystem`.
8) We no longer use Boost.Filesystem on any platforms, and thus we no longer need to compile fs separately.
9) Some other minor changes in CMake because the above gave rise to several spurious errors relating to include paths.
10) Normalized some indents

And… I think that’s it.

### Motivation

Code in fs section was NASTY.

### More

### Additional Notes

I moved some files to `game/graphic`, but there are still some files left that would make sense to eventually move to this folder... maybe revisit this in a future smaller PR.


game/fs.cc
```diff
@@ -32,46 +32,218 @@ namespace {
...
+		if (!base.empty()) return;  // Only bootstrap once
+		// Base (e.g. /usr/local), share (src or installed data files) and locale (built or installed .mo files)
+		{
+			char const* root = getenv("PERFORMOUS_ROOT");
+			switch (Platform::currentOS()) {
+				case Platform::HostOS::OS_WIN:
+					base = fs::canonical(root ? root : execname().parent_path());
+					break;
+				case Platform::HostOS::OS_MAC:
+					if (const auto appFolder{execname().parent_path().parent_path().parent_path()}; appFolder.extension() == ".app") {
```
Baklap4, line 50: Would be nice to refactor `base` so it's not kinda arbitrary how many `parent_path()` we need to go up...
Same for Windows and the default case.

game/fs.cc
```diff
@@ -32,46 +32,218 @@ namespace {
...
+	using Lock = std::lock_guard<std::mutex>;
+
+	void PathCache::pathBootstrap() {
+		if (!base.empty()) return;  // Only bootstrap once
+		// Base (e.g. /usr/local), share (src or installed data files) and locale (built or installed .mo files)
+		{
+			char const* root = getenv("PERFORMOUS_ROOT");
+			switch (Platform::currentOS()) {
+				case Platform::HostOS::OS_WIN:
+					base = fs::canonical(root ? root : execname().parent_path());
```
Baklap4, line 47: Is this correct for System Install and for Build folder?

game/fs.cc
```diff
@@ -32,46 +32,218 @@ namespace {
...
+				}
+				if (!share.empty()) {
+					break; // Found
+				}
+				base = base.parent_path();
+			} while (true);
+			// Use locale .mo files from build folder?
+			if (Platform::currentOS() == Platform::HostOS::OS_WIN) {
+				auto folder = base.filename();
+				auto reg = "x\\d{2}-([Dd]ebug|[Rr]elease)(-install)?"; // matches all build folders.
```
Baklap4, line 75: We're missing one case; DebInfo

and maybe a couple more when running on windows but through wsl

Baklap4 (changes requested):
I definitly need to check the windows side of things especially in regard to the environment variable checks/usage.

Overall it looks alright though, but gotta test it quite extensively

game/fs.cc
Lord-Kamina: > Would be nice to refactor `base` so it's not kinda arbitrary how many `parent_path()` we need to go up... Same for Windows and the default case.

Not much can be done about this. Mac apps have the `whatever.app/Contents/MacOS/whatever-exe` structure and that can't really be changed.

The real base I guess would be `Contents` but IMO that looks ugly. In windows, I guess perhaps we could improve it. Also, the purpose here is for it to play nice with something I did in the SpdLog PR:

```c++
std::string formatPath(const fs::path& target) {
	// Normalize the paths by resolving symlinks and removing redundant elements.
	// But, if just a filename, keep it as is.
	fs::path canonicalTarget = target.has_parent_path() ? fs::weakly_canonical(target) : target;
	if (canonicalTarget == cache.base) {
	// Return the absolute path if referring to the base directory.
		return cache.base.string();
	}
	// Check if the target is a descendant of base.
	if (std::mismatch(cache.base.begin(), cache.base.end(), canonicalTarget.begin()).first == cache.base.end()) {
		// If target is a descendant of base, return the relative path starting from base's parent.
		return fs::relative(canonicalTarget, cache.base.parent_path()).string();
	} else {
		// Otherwise, return the absolute path to target.
		return canonicalTarget.string();
	}
}
```

game/fs.cc
Lord-Kamina: Would you mind adding those, since that's more your area?

Lord-Kamina:
The issue @ooshlablu mentioned in Discord should be fixed now. The fix has been squashed into the earlier refactor commit, and I have also removed a now-obsolete preprocessor check for old versions of boost.

Baklap4 (approved):
LGTM

PR #1023, opened 2024-12-09
Pause menu: Move "Quit" back to being the last option.

Lord-Kamina:
### What does this PR do?

Puts "Quit" back at the bottom of the in-game pause menu.

### Motivation

For years, the last option in the "Pause" menu was "Quit", this was changed earlier this year, in a way that I think makes the user more prone to error and causing songs to be accidentally marked as "broken" when trying to quit to the menu.

### More

### Additional Notes

I think the whole "Abort" option should be revisited in the future. I see how it can be useful but maybe we should have some sort of confirmation dialog or similar, to prevent accidentally marking songs as broken.


twollgam (approved):
LGTM

PR #1028, opened 2024-12-31
update docker build documentation, add appimage prerequisites to ubun…

nerdingas-armaitis:
### What does this PR do?
Updates documentation for using docker build, fixes some warnings/best practices in the bash script and adds Appimage builder dependencies to the ubuntu image.

### Motivation
I just set up building for performous first time on my machine and thought docs could be better/clearer.

### DIFF
--- a/docker/Dockerfile.ubuntu
+++ b/docker/Dockerfile.ubuntu
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1
-ARG OS_VERSION
+ARG OS_VERSION="22.04"
 ## Use the official Ubuntu <OS_VERSION> Image from Dockerhub
 FROM docker.io/library/ubuntu:${OS_VERSION}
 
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.ubuntu
```diff
@@ -10,9 +10,11 @@ ARG TZ="America/New_York"
 ENV PACKAGE_TYPE="DEB"
 
 ## Install the deps and create the build directory
+ARG OS_VERSION="22.04"
```
Baklap4, line 13: `OS_VERSION` is now specified twice

Baklap4 (changes requested):
Also pipeline is failing on code-quality check;

docker/Dockerfile.ubuntu
nerdingas-armaitis: Sonarqube is complaining about same things as it would have been complaining about before the change, so it must have been marked as safe before. 

As for multiple ARG of OS_VERSION - This is how docker works.
ARGs defined before FROM statements do not get passed onto container, so it needs to be specified inside too. This is one of the reasons i have added the usual  bash script flags of -euo, so that unbound variables would throw an error.

Lord-Kamina: Wouldn't it be better to just use VERSION_ID, which can be obtained from `/etc/os-release`?

In fact I think I changed that ready in one of my still-open PRs.

PR #1030, opened 2025-01-02
When a song directory is empty, don't bother scanning for songs.

Baklap4:
### What does this PR do?

Fixes a runtime error where performous would crash if you had a folder specified (in config.xml) which didn't contain songs.
It now skips the folder and mentions it in the log.

### More


Baklap4:
Closing pr as build will fail on macos so i can't be merged

Issue #1016, opened 2024-11-06
Translations not working on msvc builds


Lord-Kamina:
Something else I noticed while playing with the logger. MSVC builds are missing the locale folder and thus, translations are not working.

Baklap4:
Seems the pipeline doesn't include the locales since:
Building Performous 1.3.1-1021-2ec02ea-alpha
-- Could NOT find Gettext (missing: GETTEXT_MSGMERGE_EXECUTABLE GETTEXT_MSGFMT_EXECUTABLE) 
-- Localization disabled: Gettext tools (msgfmt) missing

PR #1031, opened 2025-01-02
Try to explicitly install gettext tools

Lord-Kamina:
Probably just wishful thinking but *shrugs*

### More


Lord-Kamina:
Testing needed, but as per the build logs, it appears this does indeed fix #1016

Baklap4 (approved):
LGTM

PR #1035, opened 2025-01-04
Better CI artifact names, do away with individual artifact links.

Lord-Kamina:
### What does this PR do?

1) Finally once and for all, actually use the correct damned SHA for builds.
2) Tweak filenames a bit, by expliciting which is the PR number (where applicable), as well as including the last tag and the commit hash in builds from master as well.
3) Converted the YML files back to LF, as well as normalized the indentation at two spaces because YAML doesn't like tabs (remember that, per the project coding style, we should be using tabs elsewhere)
4) Spread the logic for determining the version over multiple lines. This is one of those cases where readibility > brevity.
5) Removed the individual OS builds from the README, because in the end that that was something else we needed to remember to update when the CI systems change. Instead, provide a link that should be stable, to both the latest builds as well as the latest release.

### More


README.md
```diff
@@ -12,21 +12,8 @@ Discord at https://discord.gg/NS3m3ad
...
-- [Linux - Fedora 36](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_36.rpm.zip)
-- [Linux - Fedora 37](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_37.rpm.zip)
-- [Linux - Fedora 38](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_38.rpm.zip)
-- [Linux - Fedora 39](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_39.rpm.zip)
-- [Linux - Fedora 40](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_40.rpm.zip)
-- [Linux - Fedora 41](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest-fedora_41.rpm.zip)
-- [Linux - AppImage](https://nightly.link/performous/performous/workflows/build_and_release/master/Performous-latest.AppImage.zip)
+- Latest Performous Release: https://github.com/performous/performous/releases/latest
+- Weekly BETA builds from the `master` branch: https://nightly.link/performous/performous/workflows/build_and_release/master?preview
```
ooshlablu, line 19: Nice!

Baklap4 (approved):
LGTM, We might wanna increase builds from weekly to daily, so the cache (for windows) stays pretty much in tact?

Also in regard to daily _or_ weekly builds. We did have a ppa which had the latest build.. is this still accesible or did we lose access? We might wanna add this again

Lord-Kamina:
> LGTM, We might wanna increase builds from weekly to daily, so the cache (for windows) stays pretty much in tact?
> 
> Also in regard to daily _or_ weekly builds. We did have a ppa which had the latest build.. is this still accesible or did we lose access? We might wanna add this again

@Baklap4 We could for sure, but keep in mind we're now using my fork of the cache action, in which I managed to implement "refreshing" the cache even if nothing has changed, to prevent it from expiring so easily.

Lord-Kamina:
I think @yoda-jm was the admin for the ppa and he gave at least some of us access to it a while ago?

PR #1048, opened 2025-04-25
Fix a couple outstanding CI issues

Lord-Kamina:
### What does this PR do?

* On Fedora:

	* Changes `ffmpeg-devel` to `ffmpeg-free-devel`. Why?

		1) Licensing issues.
		2) I don’t think we actually even need libfdk_aac.
		3) Hopefully this gets rid of the constant CI failures due to that package being unavailable.

* On mac:
	* Hopefully set-up a homebrew with a cache, so macOS builds don't take an hour each.

### More

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -88,7 +88,7 @@ jobs:
             brew tap performous/pinned_cmake3
 
       - name: Install dependencies with homebrew.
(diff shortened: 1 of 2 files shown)


Lord-Kamina:
I think I'm done here.

Baklap4 (approved):

PR #1051, opened 2025-05-11
Use the instrumental track on enabling audio/mute_vocals_track when available

GaryCXJk:
### What does this PR do?

Performous stores the instrumental tracks as defined by the most recent UltraStar format specification, however, at the moment, it's not in use. At the same time, the setting `audio/mute_vocals_track` remains unused whenever the UltraStar text files are used. This pull request attempts to rectify this.

### Additional Notes

In order to enable instrumental tracks to be stored on a song, the caching JSON had to be modified to add the instrumental file location. This part should completely be non-breaking, though requires a cache refresh.

In theory, this pull request shouldn't break Frets on Fire and / or StepMania support, this has yet to be tested, but I assume there won't be a problem, as the "Instrumental" track is only used on UltraStar text files.


Baklap4:
@ooshlablu can you give this a test run?

@GaryCXJk Thanks for the PR, Codewise it looks alright but i haven't tested it myself yet

ooshlablu:
@Baklap4 I test on Ubuntu 24.04 with guitar + vocals at the same time and these changes don't interfere with any of the sources I have, which is mostly FoFix/Phaseshift/Clonehero format. I don't have any Ultrastar songs at all tho, even without instruments, so I can't really speak for the changes. Didn't break anything for me tho :-)

ooshlablu (approved):

PR #1052, opened 2025-05-27
Allow a configured minimum value for Player Highscrore

kt--:
### What does this PR do?

Adds a new Game configuration entry for the minimum allowed high score, replacing the hard-coded constant of 2000.
Default 2000
Min 200, Max 8000, step=200

### Closes Issue(s)

n/a

### Motivation

When I first started using Performous for karaoke, I spent many months not scoring much over 500, and never being able to track my appalling high scores diminished the fun.

### More

Is there documentation on this?  None was not updated.

### Additional Notes

None.


ooshlablu (approved):
I tested this with a few instruments + vocals, and set the number all over the place and it seems to work fine. Was able to get a high score on guitar while missing on vocals, the opposite way, etc.

I would appreciate if @Lord-Kamina, @Baklap4, or @twollgam can post a review before I merge tho.

Baklap4 (approved):

PR #1053, opened 2025-05-27
Use audio "loudness" normalisation, if present in audio/video tracks

kt--:
### What does this PR do?

If an audio file (or video file audio track) contains "Replay Gain" data, use it to with the existing `libav` routines to adjust the volume of the playback.
This is particularly useful when audio comes from multiple producers.  Even professionally mastered audio CDs can and do have different base-loudness levels.

A discussion of how Replay Gain works at a technical level is beyond the scope of this MR (and my understanding).   From a users' point of view - it makes overly loud audio softer, and too soft audio louder.  Without this patch I found it necessary to adjust the amplifier volume during use of Performous.

Replay Gain can't pull-off a miracle.  There is still variation in loudness.

The option is turned on/off with a configuration entry: `audio/normalize_songs`.

### Closes Issue(s)

n/a

### Motivation

Audio from different sources comes at different loudness levels.  It's annoying when one audio track is overly loud - the user has to adjust the output volume manually, or ignore it for the next audio track to sound relatively soft.

There are numerous other software platforms using Replay Gain tags to alleviate loudness issues:
https://duckduckgo.com/?t=ffab&q=ReplayGain+2.0&ia=web

### More

n/a 

### Additional Notes

I've now tested this for over a year.  The `libav` implementation is good.

One easy way to add Replay Gain tags to an audio file is to the use `loudgain` application (with patches), but applications like `audacity` can also create them.   LibAV can create them too - but that is a further work.  (Having the Replay Gain calculated by `libav` at load time is theoretically possible - but it might take a second or so.)


ooshlablu (approved):
I tested with the package and the appimage on Ubuntu 24.04 and it makes a noticeable difference for many of my sources. Everything else works as expected :+1:

PR #1059, opened 2025-06-04
Fix formatPath.

Lord-Kamina:
### What does this PR do?

Should properly fix the problem @kt-- mentioned in #1058

### More


Lord-Kamina:
> **As discussed in discord:** It probably doesn't matter a lot, but PathCache::pathBootstrap() should lock the mutex. pathInit() could too, but it's _only_ called by pathBootstrap() anyway.

pathInit is also called from `configuration.cc`
Not locking was a mistake I made earlier. I tried fixing it now.

PR #1061, opened 2025-06-12
Look inside the file to determine the image type.

### LINKED ISSUE #436: Do not rely on extension when reading images

Feature

It would be nice if performous does not rely on filename extension when reading image. Currently if it finds a JPG file named `album.jpg` it tries to read is as PNG and (presumably) gives an error:
```
libpng error: Not a PNG file
image/error: Reading PNG failed
```

kt--:
### What does this PR do?

Previously the code would examine the image filename extension, and then use the image decoder for the matching type.  
This fails when someone on the internet names a PNG image as say JPEG, resulting in a load failure.

This patch looks inside the file, reading the file-header.  Nearly all image types have a "magic number" signature in the top of the file.  Using these few bytes, the patch identifies the file type. This is a much more robust method of determining an image type.

### Closes Issue(s)

Closes #436 https://github.com/performous/performous/issues/436

### Motivation

I had some spare time, and was looking for "low hanging fruit" bugs to fix.
I came, I saw, I fixed.

### More

n/a

### Additional Notes

**Testing:** There should be no visible change from this.   Unless you have images that "won't load", but now they do.

Probably we should add support for loading `.webp` images, as they're becoming more common.  But that's another PR.


Baklap4 (changes requested):

Baklap4 (approved):
LGTM let's get it tested by @ooshlablu and get it to merge afterwards 👍

Lord-Kamina:
If you don't care about flavor of windows but rather just "Windows", please `#include "platform.hh"` and use `BOOST_OS_WINDOWS`

testing/imagetypetest.cc
```diff
@@ -0,0 +1,131 @@
+#include <vector>
+#include <iostream>
+#include <fstream>
+#include <thread>
+#include <fcntl.h>
+#include <cstdio>
+#include <random>
+#include <sys/stat.h>
+
+#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
```
Lord-Kamina, line 10: Use platform.hh and the boost OS predefs.

testing/imagetypetest.cc
```diff
@@ -0,0 +1,131 @@
...
+
+const std::vector<uint8_t> PDF_IMAGE{ 0x25,0x50,0x44,0x46,0x2d,0x31,0x2e,0x35,0x0a,0x25,0xb5,0xed,0xae,0xfb,0x0a,0x34,0x20,0x30,0x20,0x6f,0x62,0x6a,0x0a,0x3c,0x3c,0x20,0x2f,0x4c,0x65,0x6e,0x67,0x74,0x68,0x20,0x35,0x20,0x30,0x20,0x52,0x0a,0x20,0x20,0x20,0x2f,0x46,0x69,0x6c,0x74,0x65,0x72,0x20,0x2f,0x46,0x6c,0x61,0x74,0x65,0x44,0x65,0x63,0x6f,0x64,0x65,0x0a,0x3e,0x3e,0x0a,0x73,0x74,0x72,0x65,0x61,0x6d,0x0a,0x78,0x9c,0x33,0x54,0x30,0x00,0x42,0x5d,0x43,0x20,0x61,0xa8,0x90,0x9c,0xcb,0x55,0xc8,0x65,0xa0,0x67,0x66,0x62,0x6e }; // just the first few bytes, tail truncated
+
+const std::vector<uint8_t> SMALL_FILE{ 0x01, 0x02, 0x03 };
+const std::vector<uint8_t> EMPTY_FILE{ };
+
+std::string getSecureTmpFile()
+{
+#if defined(_MSC_VER) || defined(__MINGW32__) || defined(__MINGW64__)
```
Lord-Kamina, line 36: Same as above.

Lord-Kamina (changes requested):
I haven't reviewed thoroughly but I wanted to leave the request for changes before it gets merged.

Lord-Kamina (approved):

ooshlablu (approved):
I tested this with the appimage and the package on Ubuntu 24.04 and it looks good. I was able to rebuild the cache and everything without issue, and it did pick up a few things that I had not changed the extension from a million years ago.

Issue #1033, opened 2025-01-03
Improve error handling in requesthandler.cc


twollgam:
### Performous & operating system versions

During code review I found a `catch(...)` in requesthandler.cc:
![image](https://github.com/user-attachments/assets/47465977-429b-46aa-b67a-1a37cf690496)
I suggest to use a `catch(std::exception const& e)` and send  the reason of this exception back to the client. Otherwise it should be logged. Only if there could be exceptions that are not derived from `std::exception` use additional a `catch(...)`-block.

### What is the problem?

The reason for the exception is actually lost.

Lord-Kamina:
I propose we integrate this in #1011, which already has a myriad other changes to how errors/exceptions are handled and reported.

Lord-Kamina:
Actually I'm gonna backpedal on this @twollgam because it seems this is mostly a quirk of CppRESTSDK and is very tied to the way some their task manager works.

What I'm trying to say is, it's probably not worth the effort considering we're looking to hopefully get rid of this library sooner rather than later.

PR #1064, opened 2025-06-17
Improved error handling

kt--:
### What does this PR do?

As per issue #1033 - add logging and message-to-client on web server issues.
- Added support for 404 errors
- Added the exception details to the web result
- Logged the exception details at Error level.

### Closes Issue(s)

#1033 

### Motivation

Simply repairing low-hanging-fruit issues.
I Came, I Saw, I Fixed.

### More

n/a

### Additional Notes

I found this difficult to test.  I could readily test the 404 error, and this verified the approach taken with the other handlers.  
But I'm not sure to how trigger the other internal errors.

Note that if you try to use `wget -O - http://127.0.0.1:8000/missing.blah`, that `wget` does _not_ show the actual returned document.  But it was close enough to what we had to be really confusing.

### DIFF
--- a/game/requesthandler.cc
+++ b/game/requesthandler.cc
@@ -35,7 +35,7 @@ void RequestHandler::HandleFile(web::http::http_request request, std::string fil
 	try {
 		fileToSend = findFile(fileName).string();
 	}
-	catch (const std::runtime_error &e) {
+	catch (std::runtime_error const& e) {
 		SpdLogger::error(LogSystem::WEBSERVER, std::string("HandleFile() File Not Found. Client {}. {}"), clientIp, e.what());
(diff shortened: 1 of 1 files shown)


game/requesthandler.cc
```diff
@@ -53,6 +64,7 @@ void RequestHandler::HandleFile(web::http::http_request request, std::string fil
 			content_type = "image/x-icon";
 		}
```
twollgam, line 65: Content type detection should be a function. Don't know if it is only uses here. Cause here is unchanged code, it is only a suggestion.

twollgam (changes requested):
I request some little changes and some optional changes/refactorings.

PR #1065, opened 2025-06-19
Ignore metadata and backup files

kt--:
### What does this PR do?

Bohning complained about #1059, saying that Performous tries to load MacOS meta-data files (which contain icons, etc. (AFAIK)), and then noisily logs an exception.   This patch adds a regex filter on Song file loading, that skips these.   

At the moment, it handles files beginning `._` and `.#`  

The core issue is that (AFAIK) MacOS stores metadata about `songfile.txt` as `._songfile.txt`.  Performous sees a `.txt` file and tries to parse it as a song.   It's not really a bug.

*note:  I haven't used a Mac since they had those cool little monochrome displays.

### Closes Issue(s)

n/a

### Motivation

Just doing whatever I can to get patch #1059 through.

### More

I only know of the MacOS meta-data prefix, and the EMACS backup-file prefix.  Maybe there's more?

### Additional Notes

**Testing**:  Create a `._something.txt` file, containing invalid rubbish, in an existing song directory, and verify that the debug message is logged (saying it's skipped) and/or no exception is logged for that file.


game/songs.cc
```diff
@@ -299,6 +299,11 @@ void Songs::reload_internal(fs::path const& parent, Cache cache) {
 			if (!regex_search(p.filename().string(), expression)) {
 				continue; //if the folder does not contain any of the requested files, ignore it
 			}
+			std::regex ignoredFiles("^(\\._.*|\\.\\#.*)$");  // skip MacOS meta, common backup files
```
Baklap4, line 302: Changes look good on a first glance. I am wondering about the macos specific files like: `.DS_Store` these won't be captured by the regex, but do they really need to be captured? Since a couple of lines above we already filter out most of them?

Baklap4 (approved):

PR #1066, opened 2025-06-24
Webp image support

kt--:
### What does this PR do?

Sometimes songs use WebP images for covers, backgrounds, etc.  Performous cannot load these images.
Using the google WebP library, this PR adds support for WebP format wherever we typically support JPEG.
I only added support for normal images, not webp animations, etc.

**NOTE:** This PR bring a build-time dependency on `libwebp`   (Previous versions of this PR only used webp if it was found)

### Closes Issue(s)

n/a

### Motivation

Tired of converting .webp to .jpeg

### More

I updated all the Dockerfiles, CMakes and what-not. Anywhere libjpeg was being pulled in, I added a libwebp.

### Additional Notes

**Testing:** 
- Converted existing song's .jpg files to .webp, verified images worked the same.  (you need to remove ~/.cache/performous)
- Converted existing song's .jpg files to .webp, verified load failed to blank without the library compiled in.
- Verified Ubuntu 24.04 docker image builds OK
- Verified Debian 12 docker image builds OK
- Verified Fedora 42 docker image builds OK
- (No testing for Windows/Mac - I don't have)

### DIFF
--- a/game/fs.cc
+++ b/game/fs.cc
@@ -309,23 +309,30 @@ std::string formatPath(const fs::path& target) {
 }
 
 const fs::path PathCache::getLogFilename() { Lock l(m_mutex); return cache / "infolog.txt"; }
-const fs::path PathCache::getProfilerLogFilename() { 
+const fs::path PathCache::getProfilerLogFilename() {
 	Lock l(m_mutex);
-	std::string baseName{"profiler.txt"};
-	unsigned logRotate = 5;
-	do {
-		fs::path newPath = cache / fmt::format("profiler.{}.txt", logRotate);
-		fs::path oldPath = cache / (logRotate == 1 ? "profiler.txt" : fmt::format(fmt::runtime("profiler.{}.txt"), logRotate - 1));
-		if (fs::exists(oldPath)) {
-			if (logRotate == 5 && fs::exists(newPath)) {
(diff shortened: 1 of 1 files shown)


game/CMakeLists.txt
```diff
@@ -132,6 +132,21 @@ foreach(lib SDL2 PangoCairo LibRSVG LibXML++ AVFormat SWResample SWScale ZLIB JP
 	add_definitions(${${lib}_DEFINITIONS})
 endforeach(lib)
 
+# Find all the optional libs, and #define HAVE_$lib if found (e.g.: HAVE_WebP)
```
Baklap4, line 135: for other modules like opencv, webserver or midi we use the format `add_definitions("-DUSE_PORTMIDI")` and within code the `USE_PORTMIDI` ifdefs. Might be good to align that _if_ we want to keep it as an optional lib.

Personally since webp is such a big part of the web and libpng and jpeg are already as core libraries defined we might even think of adding this as one of the core libraries instead of making it optional. @Lord-Kamina what do ya think?

Baklap4 (dismissed):
Code wise looks good; i haven't got to try it yet tho, but if @ooshlablu gets there first feel free to merge this :)

PR #1068, opened 2025-07-07
Fixes remote access on Windows

Baklap4:
### What does this PR do?

For some reason cpprestsdk cannot bind on *, but it needs to be binding to +. Not sure what the difference is, but once setup the correct acl's (or run as admin) and setup some firewall rules it all works flawlessly again.

### Closes Issue(s)

None

### Motivation

Webserver remote connections weren't be able to used on windows.

### More

Aside from this code change the end-user has to:
1. Register the ACL (or run as admin)
    * Open up powershell (as admin) 
    * `cmd /c "netsh http add urlacl url=http://+:8000/ sddl=D:(A;;GX;;;WD)"`
        * This will register the acl for every user on the system.
3. Register a firewall rule
    * Open up powershell (as admin)
    *  `New-NetFirewallRule -DisplayName "Performous Allow Port 8000 TCP" -Direction Inbound -Protocol TCP -LocalPort 8000 -Action Allow`

### DIFF
--- a/game/webserver.cc
+++ b/game/webserver.cc
@@ -31,7 +31,7 @@ void WebServer::StartServer(int tried, bool fallbackPortInUse) {
 	} else {
 		if (Platform::currentOS() == Platform::HostOS::OS_WIN)
 		{
-			addr = "http://*:" + portToUse; // Allow Windows to accept all connections. Needs admin privileges though.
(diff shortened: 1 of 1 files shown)


ooshlablu (approved):
I tested it for kicks on Ubuntu and this obviously doesn't cause any issues there. I can't really say much about Windows because I'm allergic to OS's that aren't Unix or Unix-like, but LGTM

PR #1073, opened 2025-07-16
Config for finding the Retro Cult Mods V3 Guitar Adapter

kt--:
### What does this PR do?

One way of using the old Wii Guitar Hero controller is to buy the Retro Cult Mods V3 Wii adapter.
This is a cabled USB device that plugs into this controller in lieu of a wiimote.

Ref: https://shop.retrocultmods.com/products/v3-wii-adapter-for-clone-hero-and-rb4-rock-band-4

Currently Performous does not support this controller, because it's missing the XML configuration.
The MR adds that configuration.

### Closes Issue(s)

n/a

### Motivation

I wanted to use circa 2008 Wii Guitar Hero controller with Performous.

### More

I still wanted to update the doco, but that's not in the repo' (as far as I could tell).

### Additional Notes

**Testing**:   
- I added a custom controller configuration for it `~/.config/performous/controllers.xml`.  
```<performous>
 <entry name="game/instruments" type="string_list">
   <stringvalue>0:GUITAR_RETROCULTMODS</stringvalue>
 </entry>
</performous>
```

- Played with the guitar for a while, verifying:
  - Fret buttons + strumming
  - Whammy bar  (is that he correct term?)
  - Joystick U/D/L/R
  - "God Mode" by tilting the guitar neck up.

I did not test the `(+)` or `(-)` buttons.
I briefly tested "lefty mode" via operator error (it worked).


sanjay900:
If anything it would make more sense to have the mappings be for Santroller in general, being that there are more devices than just RCM adapters using these mappings

sanjay900:
We have a lot of docs on the instruments already, but you really want to be matching on vid and pid (and revision if you want to properly match the santroller / rcm stuff).
That's how all the games worked, besides Xbox because xinput is xinput

PR #1080, opened 2025-08-22
Modify normalisation to support R128_TRACK_GAIN

### LINKED ISSUE #1078: Loudness Normalisation not working for audio files using the Opus codec

Performous: 1.3.1
OS: all

Currently we use `libav` (from FFMPEG) to determine the relative loudness of a tagged audio track by checking for the REPLAYGAIN_TRACK_GAIN  tag, and adjust the volume to normalise the human-perceived loudness.  This makes overly-loud tracks softer, and vice-versa.

However, even when tagged, opus format files use a different tag name (and loudness "LUFS" offset).  So **the existing mechanism does not work for .opus.**

@complexlogic on Discord writes:
> [...] your implementation won't work on Opus files tagged by USDB Syncer, UltraStar Manager, or `loudgain`. The reason is because Opus uses a different metadata tag. For Opus files, FFmpeg doesn't parse the gain values into the `AVReplayGain`stream side data like it does for the other formats. You should check for the existence of `R128_TRACK_GAIN` in `AVStream.metadata` for Opus files.

Indeed the `loudgain` documentation states that:
> Opus adheres to EBU-R128, so the loudness reference is always -23 LUFS. [...] This means players have to add an extra +5 dB to reach the loudness ReplayGain 2.0 prescribes (-18 LUFS). IDJC, foobar2000 and several others already seem to handle this correctly (in newer versions).

So we need to:
- Detect the `R128_TRACK_GAIN` tag in the audio track 
- Read the gain (via libav too?), accounting for the difference in the reference point (+5dB)

This is somewhat important work, since the software USDB Syncer now prefers(?) opus codec (for good reason, performance is best-in-class).  However non-opus formats will continue to normalise correctly without this fixed.

Ref: Replay Gain Spec.: https://wiki.hydrogenaudio.org/index.php?title=ReplayGain_specification
Ref: RFC-7845 Ogg:  https://www.rfc-editor.org/rfc/rfc7845
Ref: loudgain: https://github.com/Moonbase59/loudgain
Ref: USDB Syncer: https://github.com/bohning/usdb_syncer

### What is the problem?

The existing loudness normalisation method does not work for .opus

### Steps to reproduce

Use a .opus audio track with loudness normalisation.  Hear that it's not normalised.

### Additional context

n/a

n/a

kt--:
### What does this PR do?

Adds support for the EBU R 128 gain tag "R128_TRACK_GAIN" to Performous.  Audio files with the opus codec use this tag, and do _not_ use REPLAYGAIN.  So we could not perform normalisation on .opus files.

### Closes Issue(s)

Closes #1078 

### Motivation

Fix normalisation on opus audio.

### More

Additionally I added REPLAYGAIN tags to the menu theme audio (but not the instrument sounds).

### Additional Notes

TESTING:
 - As before, remove any gain tags from an overly loud (or soft) OPUS audio track.  Maybe use Audacity to make it extra-loud.
 - Play the audio in Performous, hear how loud it is.
 - Add the R128 tags to the audio file, I use `loudgain` for this. Ref: https://github.com/Moonbase59/loudgain
 - Play the audio again in Performous, hear how the sound loudness becomes balanced (normalised).


Baklap4:
@kt-- 

They might have gotten tags, but again i don't see them when i parse them with `vorbiscomment`

```
baklap4@DESKTOP-PGG8NVG:/mnt/c/repos/performous/performous$ vorbiscomment -l data/themes/default/menu.ogg
COMMENT=produced by john holowach

vocals & lyrics by \//\\/

composed by rjmarshall as
[ tinderbox ]

remixed by john holowach as
[ the big finale ]
ALBUM=tryad.org
ARTIST=Tryad
TITLE=The Final Rewind
DATE=01.11.2005
COMMENTS=Performous menu music (cut of original)
```
```
baklap4@DESKTOP-PGG8NVG:/mnt/c/repos/performous/performous$ vorbiscomment -l data/themes/default/notice.ogg
```
```
baklap4@DESKTOP-PGG8NVG:/mnt/c/repos/performous/performous$ vorbiscomment -l data/themes/default/practice.ogg
REPLAYGAIN_ALBUM_GAIN=6.19 dB
REPLAYGAIN_ALBUM_PEAK=0.437262
REPLAYGAIN_ALBUM_RANGE=10.01 dB
REPLAYGAIN_REFERENCE_LOUDNESS=-18.00 LUFS
REPLAYGAIN_TRACK_GAIN=6.19 dB
REPLAYGAIN_TRACK_PEAK=0.437262
REPLAYGAIN_TRACK_RANGE=10.01 dB
```

PR #1090, opened 2025-10-11
Set loadStatus to HEADER for songs loaded from cache. Fixes #1088

### LINKED ISSUE #1088: Performous crashes with cannot create std::vector larger than max_size() when trying to play a song

Performous: master branch (d5278e3)
Operating system: Ubuntu 25.04

### What is the problem?

Performous crashes after selecting to start playing a song with the following log:

```
[01:46:11]:::LOGGER / ERROR::: cannot create std::vector larger than max_size()
```

### Steps to reproduce

1. Select a song on the song list
2. Select "play" on the next screen

### Additional context

The issue seems to be triggered when we try to create a `Player` because the frames parameter is `9223372036854775808`, and it tries to create a vector with that many elements, which is not possible.

The reason frames has that value is because `endTime` is never set for the song. The only place where I could find it being set is in `SongParser::finalize`, which is only called from `SongParser::SongParser`, but only if `s.loadStatus == Song::LoadStatus::HEADER`.

I'm not completely sure how `s.loadStatus` is supposed to work, but I found a way to fix the problem for me, so I'll create a pull request and wait for feedback.

[01:46:11]:::LOGGER / ERROR::: cannot create std::vector larger than max_size()

soonick:
### What does this PR do?

Fixes #1088

### Closes Issue(s)

#1088

### Motivation

Issue #1088

### DIFF
--- a/game/song.cc
+++ b/game/song.cc
@@ -55,10 +55,8 @@ Song::Song(nlohmann::json const& song) : dummyVocal(TrackName::VOCAL_LEAD), rand
 	music[TrackName::GUITAR_COOP] = getJsonEntry<std::string>(song, "guitarCoop").value_or("");
 	music[TrackName::GUITAR_RHYTHM] = getJsonEntry<std::string>(song, "guitarRhythm").value_or("");
 
-	// never load loadStatus as FULL, as that is only true after it has been fully parsed
-	// a song loaded from cache only ever has the header information at best and should not be considered
-	// fully parsed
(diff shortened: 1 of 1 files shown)


Baklap4:
For now I've updated your branch to match latest master - we'll still need to do some testing around this especially since a lot of fixes have already been made to the caching/loadstatus in more recent commits than 1.3.1

@ooshlablu can you help out testing this?

PR #1102, opened 2025-12-27
High score columns

kt--:
### What does this PR do?

When text with columns is rendered in Performous, we rely on Pango's handling of tab-stops to nicely align columns.   This somewhat works, but in the default Performous theme, even having a high score with some `1`s in it is enough to break the alignment.

broken

This patch checks for columns in the text, and if some are found, it precisely calculates the tab-stop positions.  The code  then uses Pango's tab-stops API to set the stops to the widest column widths.  This ensures columns line-up nicely, even for scores like `1111` and long player names.

repaired

There is no change to the way Performous gameplay works.

### Closes Issue(s)

n/a

### Motivation

This makes the high scores table easier to read.

### More

n/a

### Additional Notes

Hmm, looking at the screen-shot, I'm now wondering if we should add a `\t` between the date and the time in the timestamp.

## Testing:
- Before testing this patch, scroll through your song list, and find some items that have column display issues.  Note down both songs with and without issues.
- Apply the patch
- Locate the previously found songs, observe _beautifully_ formatted columns.
- Observe the songs that previously had OK columns are still good.


Baklap4:
Within your screenshots i see the 0-padded items - this should've been fixed in latest master. Do you have a screenshot of this aswell?

Baklap4:
Currently only windows is failing with:

```
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(67): error C2676: binary '[': 'std::array<float,16>' does not define this operator or a conversion to a type acceptable to the predefined operator
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(68): error C2676: binary '[': 'std::array<float,16>' does not define this operator or a conversion to a type acceptable to the predefined operator
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(70): error C2027: use of undefined type 'std::array<float,16>'
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.44.35207\include\utility(167): note: see declaration of 'std::array<float,16>'
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(98): error C2079: 'colWidths' uses undefined class 'std::array<float,16>'
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(102): error C2664: 'size_t TextRenderer::measureColumns(const std::string &,const TextStyle &,float,std::array<float,16> &)': cannot convert argument 4 from '<error type>' to 'std::array<float,16> &'
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(49): note: see declaration of 'TextRenderer::measureColumns'
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(102): note: while trying to match the argument list '(const std::string, const TextStyle, float, <error type>)'
D:\a\performous\performous\performous\game\graphic\text_renderer.cc(109): error C2109: subscript requires array or pointer type
```

game/graphic/text_renderer.cc
```diff
@@ -31,6 +31,52 @@ namespace {
...
+	// Previous to this call, we've determined there is some columns in the text string.
+	// Iterate over the text, measuring the pixel-width of each column in each line
+	size_t maxCols = 0;
+	std::istringstream issLines(text);
+	std::string line;
+	while (std::getline(issLines, line)) {
+		// Only consider lines with tab-delimited columns
+		size_t tabCount = static_cast<size_t>(std::count(line.begin(), line.end(), '\t'));
+		if ( tabCount > 0 ) {
+			std::istringstream issCols(line);
```
Baklap4: This is likely a performance hit creating a new `istringstream` everytime

game/graphic/text_renderer.cc
```diff
@@ -31,6 +31,52 @@ namespace {
...
+			std::string colText;
+			size_t colIndex = 0;
+			while (std::getline(issCols, colText, '\t')) {
```
Baklap4: We detect `\t` multiple times: Here and on line 58 and also on 99

Baklap4 (changes requested):
As you mentioned it might also be good to add a tab between date and time

game/graphic/text_renderer.cc
kt--: I've refactored the code to not use `std::istringstream` and generally to absolutely minimise allocations.
IIRC, there is only two `std::string`s left in the columisation code.  Without those, I would have had to drop back to C-type code, and the style of Perfomous is to use modern C++.

game/screen_songs.cc
```diff
@@ -373,11 +381,12 @@ std::string ScreenSongs::getHighScoreText() const {
...
-		fmt::format_to(std::back_inserter(ret), "{}:\n", track);
+		fmt::format_to(std::back_inserter(ret), "{}:\n", track);  // The part of the song played, e.g: "Vocals"
+		// One line for each player score
 		for (auto const& score: scores) {
-			fmt::format_to(std::back_inserter(ret), "{:>10L}\t", score.score);
-			playerFormatter(ret, score.playerid);
-			timeFormatter(ret, score.unixtime);
+			fmt::format_to(std::back_inserter(ret), "{:>10L}\t", score.score);    // score
+			playerFormatter(ret, score.playerid);                                 // player name
+			timeFormatter(ret, score.unixtime);                                   // date-stmap
```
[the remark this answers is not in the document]
kt--: fixed.

kt--:
> As you mentioned it might also be good to add a tab between date and time

I agree, and I've added this in.  It replaces any spaces in the date-time format with tabs.  So it works with all our date formats.  IMNSHO, it looks a lot nicer:

datecols

game/graphic/text_renderer.hh
```diff
@@ -2,12 +2,21 @@
...
 
 class TextRenderer {
 public:
-	OpenGLText render(std::string const&, TextStyle const&, float m);
-	Size measure(std::string const&, TextStyle const&, float m);
+	static const size_t MAX_COLUMNS{32};  // The maximum number of lines/columns we will measure
+
+	OpenGLText render(const std::string &, const TextStyle&, float m);
+	Size measure(const std::string &, const TextStyle&, float m);
+	size_t measureColumnsOLD(const std::string& text, const TextStyle& style, float m, std::array<float,MAX_COLUMNS>& columns);
```
Baklap4: This one can be/should be removed i think?

kt--:
@Baklap4  - Is this waiting on me for anything?

Baklap4 (commented):
Whoops i left the re-review pending. Sorry for that!

game/graphic/text_renderer.hh
kt--: Oops!  :1st_place_medal:

Baklap4 (dismissed):

kt--:
Resynchronised to master.

ooshlablu (dismissed):
LGTM!

Lord-Kamina:
Why are you using std::array?

kt--:
> Why are you using std::array?

(I'm answering in the order of perceived relevance, not trying to be cute)

**I**'m using `std::array` over say `std::vector` for efficiency.  With an array everything will be allocated on the stack (when the stack pointer is moved for the call), rather than (expensively) from the heap once the function starts executing.   This is always a trade-off with vector, since even with a single initial `vector::reserve()` call, usage still needs an underlying allocation and free.  Here the trade-off is having a fixed number of columns.   

I call allocating from the heap "expensive" - it's expensive compared to simply increasing the stack-pointer by a little more. Still, allocation kicks-off reasonably complex chain of code.  The "worst" case is appending to a vector without knowing a limit (and not using `reserve()`, etc.), where additional allocations are made - and std::vector guarantees a contiguous block of memory (IIRC), so this can be a bit of work for the heap too.

The function is supposed to be a quick utility call, and the entire call done with arrays will be faster - of course we're talking micro-second levels here, but faster is faster.  If the array existed for longer, such that the startup cost became meaningless, a vector would have been an arguably valid option.


**I**'m using `std::array` over C-style arrays, just so we can use the `.size()`, and Performous has a heavy "modern C++" style, so it seemed appropriate.


**I**'m using arrays to hold the data so I can easily loop, doing the same thing for each set.  Making the code smaller and simpler.

kt--:
Ok, I've rebased _again_.

Baklap4 (dismissed):

ooshlablu (approved):
LGTM. Just tested with guitar + vocals and no gameplay was affected. Everything looks nice and neat wrt the high scores.

PR #1103, opened 2026-01-06
Force gettext-runtime from source, try canned gcc

Lord-Kamina:
### What does this PR do?

Further try to fix/refine the installation of dependencies for macOS.

### More


Baklap4:
Why are we downgrading clang?

Lord-Kamina:
> clang?

Because at the time of commit, there was no canned binnary gcc package built with clang > 18. 

That version of clang is not being used to compile anything else anyway.

Lord-Kamina:
i.e this clang only gets pulled as a dependency, and using newer version means we have to build gcc15, thus wasting hours of CI

Lord-Kamina:
@Baklap4 could we please get this merged?

Baklap4 (approved):

PR #1106, opened 2026-02-09
Webserver/backend overhaul

GaryCXJk:
### What does this PR do?

I felt like I should separate the front- and backend changes into separate pull requests, making it easier for me to synchronize changes between the various webserver related branches.

Changes:

* Code uses RequestHandler::SongToJsonObject and RequestHandler::SongsToJsonObject as much as possible, as most code retrieving song data would basically do the same with minor changes.
* Both song list retrieval and song search can be paginated paginated (uses query params `offset` and `limit`).
* `/api/getDataBase.json` query string has been parsed, so it's no longer dependent on string order.

New endpoints:

* `GET /api/getCurrentSong`: Retrieve the current song, or the timer when on the playlist screen and autoplay is enabled.
* `POST /api/check`: Check whether the requested song exists / is loaded.
* `POST /api/add/priority`: Add a song at the beginning of the playlist. Requires the same data as `/api/add`.
* `POST /api/add/play`: Immediately play a song instead of queueing it. Requires the same data as `/api/add`.
* `POST /api/play`: Play a certain song, or seek to a certain position (from the start of the song). Either requires `songId` (similar to `/api/remove`, the song at a certain position, where the first song is 0), or `seek` (the position in the song to seek to in seconds). It should be able to accept both, but I don't really know what the use case of that would be.
* `POST /api/skip`: Skip the current song.

### Motivation

* Same backend codebase on the backend for frontend changes, ensuring code wouldn't conflict on rebase.

### Additional Notes

All these new endpoints can be tested using any AJAX request method the `fetch` API on the console. For `/api/getCurrentSong`, you can just use `fetch('/api/getCurrentSong')`. For the POST method endpoints, you need to add the method as well as set the Content-Type header:

```
fetch('<endpoint>', {
    method: 'post',
    header: {'Content-Type': 'application/json'},
    body: JSON.stringify(<JavaScript object with the required data>),
});
```


Lord-Kamina:
I'm busy with other stuff at the moment but I think we should revive the branch fron #545, without the "search decoupling" bit (it was a noble effort but the approach is fundamentally flawed, I think now) and integrate some of this PR's changes into it.

PR #1107, opened 2026-02-18
Fix multiple vocal tracks and make the guitar behaviour like RB/GH

ooshlablu:

### What does this PR do?

For as long as I can remember, the behaviour of the guitars in Performous was that notes would last for 1.5 seconds, and using the whammy bar was necessary to sustain the note. This is not the behaviour in either Rock Band or Guitar Hero, just holding down the fret is enough to sustain the note. This does mean the whammy bar is kind of useless, but it was useless in those games also: the only thing it did was pitch-bend the tracks, it had no influence on scoring or anything.

At some point kind of recently too, there was a bug introduced where songs with multiple vocal tracks show up as such on the song selection screen, but when you entered into the song, the only tracks available were `Vocals` and `dummytrack0`. This restores the tracks for songs with multiple vocals for at least FoFix/PhaseShift/CloneHero formats (basically anything that uses midi to store the notes for the instruments and vocals), but I do not have any other formats to test with. If there is currently an issue with Singstar songs also, then I'm very certain that this fix will not help.

I would appreciate if someone could test this build with sources other than FoFix/PhaseShift/CloneHero. Sources with guitars especially would be greatly appreciated, but any type of source even with a sing test should do fine.

I also fixed an issue where the first and second players were drawn in a random order. Sometimes the first player would be at the top of the screen, sometimes it would be the bottom.

### Motivation

The guitar behaviour was bothering me for a while :-)

I recently fired up Performous to find the issue with the vocal track selection and it didn't behave how it used to. this fix worked for me at least, not sure when or what PR the actual issue was introduced in.

### DIFF
--- a/game/songparser.cc
+++ b/game/songparser.cc
@@ -98,6 +98,10 @@ SongParser::SongParser(Song& s) : m_song(s) {
 				s.m_bpms.clear();
 				addBPM(0, bpm);
 			}
+			// Ensure MIDI filename is known before parsing (for songs that need MIDI)
+			if (s.midifilename.empty()) {
+				guessFiles();
+			}
 			if (s.type == Song::Type::TXT) txtParse();
 			else if (s.type == Song::Type::INI) midParse();  // INI doesn't contain notes, parse those from MIDI
 			else if (s.type == Song::Type::XML) xmlParse();
--- a/game/songparser.cc
+++ b/game/songparser.cc
@@ -91,16 +91,25 @@ SongParser::SongParser(Song& s) : m_song(s) {
 			}
 			m_ss.str(ss);
 		}
-		// Header already parsed?
-		if (s.loadStatus == Song::LoadStatus::HEADER) {
+		// Header always parsed after this point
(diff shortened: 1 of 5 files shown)


game/songparser.cc
kt--: Does it matter that we're doing `guessFiles()` here, and in the previous if() block ?
Maybe the `guessFiles()` can always be done just after the "//Header always parsed".

Does it even matter?  (feel free to just mark "resolved" if not).

EDIT:  Doesn't look like it there's any issues with the double-call, except efficiency.

kt-- (approved):

game/songparser.cc
ooshlablu: Yeah, I'm not really sure either TBH. If you think there's a more efficient way to accomplish this then I'm more than happy to try it out :-)

ooshlablu: I did end up fixing it so there's only 1 call to `guessFiles()`

kt--: I reckon that's better.

Baklap4 (approved):

PR #1108, opened 2026-02-19
Allow adjustable track volume per-difficulty on failure

### LINKED ISSUE #889: When kids mode is selected for an instrument, it should always play the track
### Suggestion

When `Kids` mode is selected for an instrument, the track should play no matter what, and not mute on misses.  I think we all know why :-)
Kids mode is usually just the `Easy` track anyway, so if someone wanted the real game-play experience they could always switch to `Easy`. Making the fail sound is still fine, and it could be interesting to have this be configurable.

### Use case

_No response_

_No response_

ooshlablu:

### What does this PR do?
~~When "Kids" is selected as the difficulty for the instruments, the sound track for that particular instrument continues to play no matter what happens, but the fail sound still plays for off-timed strums and wrong frets, and any wrong conditions still don't add to the score.~~

Originally with this PR, I wanted to make the track play no matter what when a failure condition was met on `Kids` difficulty. I've now re-worked it to allow adjustment of the underlying track volume during a failure condition for each difficulty level.

### Closes Issue(s)

closes https://github.com/performous/performous/issues/889

### Motivation

Want to prevent discouragement while I try to build up some more testers :-)

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -57,6 +57,10 @@ to save the current settings to XML.
 		<short>Difficulty</short>
 		<long>Game difficulty.</long>
 	</entry>
+	<entry name="game/kids_keep_playing" type="bool" value="false">
+		<short>Kids Mode Keep Playing</short>
+		<long>When enabled, the track keeps playing when someone misses a note in Kids difficulty mode, but still plays the failure sound.</long>
+	</entry>
 
 	<entry name="game/Textstyle" type="uint" value="0">
 		<limits>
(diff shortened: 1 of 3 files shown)


kt--:
Does this change _need_ to be tied to kids mode?

From the perspective of my lack of guitar playing ability, the change seems like a worthwhile config option for all modes, not just  `m_level == Difficulty::KIDS` ?

ooshlablu:
> Does this change _need_ to be tied to kids mode?

It doesn't necessarily, and I did actually think about making it for all. I just kind of picked it because this is where I needed to happen at this point lol. I can definitely change it

kt-- (approved):

Baklap4 (approved):

PR #1109, opened 2026-02-19
Fix a bug where 3-note-chords would not be counted in the combo

ooshlablu:

### What does this PR do?
3 note chords on the guitar weren't being counted in the combo. Everything else appeared to work fine though: misses were being triggered correctly, the tracks were playing when they should, and the score would get points added, but the combo would fail every time. 

### Motivation
It was bothering the crap out of my brother. The combo is VERY important to him lol.


kt-- (commented):
Looks ok to me, but I'm not familiar with this part of the code, so I'll leave the review to others.

Baklap4 (approved):

PR #1110, opened 2026-02-21
Make point deductions configurable

ooshlablu:

### What does this PR do?

This creates a toggle for points deductions on instruments. The current behaviour doesn't match RB/GH, but instead of just removing it I figured I'd make it optional.

### Motivation

I was originally going to include this in https://github.com/performous/performous/pull/1108 to go with the original idea of making Kids difficulty a little more friendly, but that ended up changing completely.

### DIFF
--- a/data/config/schema.xml
+++ b/data/config/schema.xml
@@ -57,6 +57,10 @@ to save the current settings to XML.
 		<short>Difficulty</short>
 		<long>Game difficulty.</long>
 	</entry>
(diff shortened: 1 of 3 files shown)


Baklap4 (dismissed):
Thanks for the nice addition!

Baklap4 (approved):

Baklap4:
Please rebase on latest changes to get this merged 👍

PR #1113, opened 2026-03-24
Track song mtime for cache invalidation

### LINKED ISSUE #1112: Updating a Song .txt causes a load error Vs cached copy

Performous master branch, on Debian Linux

### What is the problem?

When the user updates a song file in the song directory, such that there are new/changed resource file(s) the song no longer loads, as the cached version of the song is used, and the previously cached resources may no longer exist.

Performous should realise the song has updated, and invalidate the cache.

**Workarounds**: remove the cache directory to regenerate entire cache or re-scan the songs.

### Steps to reproduce

1. Have an existing, cached song file, that references audio in say `blah.mp3`
2. Play that song in Performous, verify it works OK (and is now cached)
3. Exit Performous
4. Update that song, change the song file to use the audio `blah.opus`, and remove the `blah.mp3` - as if the song was updated externally.
5. Re-run Performous, find the song, observe a load error.

### Additional context

_No response_

_No response_

kt--:
### What does this PR do?

It stores the `mtime` of the song file in the cache, and is used to (in)validate the song cache against the song.
This is the same method used for other cache invalidation activities in Performous.

### Closes Issue(s)

This PR fixes #1112 

### Motivation

Updated some song files, and then they failed to load in Performous, despite there being no actual errors in the song data.

### More

n/a

### Additional Notes

TESTING:
 - Play a song file in Performous, to ensure it works, and is in the cache.
 - Replace the song audio file with a different file, by editing the .TXT file, and removing the old file (important!)
 - Re-start Performous, find the song, and verify that it no longer loads, and the error-log test shows it using the old invalid path from the cache.
 - Apply the patch, repeat the test, verify that the song now loads correctly.
 - Browse the loading logs as songs log, verify the cache operation is preserved correctly - it is not invalidated all the time, or suchlike (Although on the first run, there will be no `mtime` in the cache.)


Baklap4 (approved):
Looking good, this would solve so many issues with the invalidation of the cache.

PR #1116, opened 2026-05-21
Fix MSVC and AppImage builds

ooshlablu:
### What does this PR do?
I ran into the NSIS issue we are seeing when I was working on a bunch of CI fixes the other day. They show up as:
```

Run repolevedavaj/install-nsis@v1.0.3
Run Invoke-WebRequest https://deac-riga.dl.sourceforge.net/project/nsis/NSIS%203/3.11/nsis-3.11-setup.exe?viasf=1 -OutFile C:\WINDOWS\Temp\nsis-3.11-setup.exe
Start-Process: D:\a\_temp\cfe5ab24-deac-4274-8e40-e1efb734f553.ps1:3
Line |
   3 |  $process = Start-Process "C:\WINDOWS\Temp\nsis-3.11-setup.exe" -Argum …
     |             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | This command cannot be run due to the error: An error occurred trying to start process
     | 'C:\WINDOWS\Temp\nsis-3.11-setup.exe' with working directory 'D:\a\performous\performous'. The file or directory
     | is corrupted and unreadable.
Error: Process completed with exit code 1.
```

This PR pulls in the fix for that issue from my buildtacular branch.

IT looks like there may have also been a Widows runner update what was causing an issue with cpprest, so I disabled webserver on the MSVC builds like they are for MinGW. I will see if I can fix this issue in buildtacular.

### Motivation

A bunch of PRs need to be rebased to merge, and the MSVC build will fail without this fix.


Baklap4:
> IT looks like there may have also been a Widows runner update what was causing an issue with cpprest, so I disabled webserver on the MSVC builds like they are for MinGW. I will see if I can fix this issue in buildtacular.

please let's not do this. There are some PR's upstream where this has been fixed / patched.
https://github.com/microsoft/vcpkg/pull/51865
https://github.com/microsoft/cpprestsdk/pull/1844

as you can see it's already fixed on vcpkg's side of things so if we update the hash after that change we should probably be good to go :) Latest hash: `22d9c655dc24ad24c6cac27607005e619ce46740`

ooshlablu:
@Baklap4 ok, I'll try the version bump to see if we can fix cpprestsdk.

The NSIS fix is still required AFAICT tho unfortunately

Lord-Kamina:
The nsis is fixed probably by simply updating the nsis action we were using. Also, you’re removing the cache for no real reason. Most likely the proper CI fixes will be my PR. MSVC is currently failing because I haven’t bumped the revision. I’m waiting because I expect they will make a new release this week that will likely include the fix.

Baklap4 (changes requested):
I'm with @Lord-Kamina in this one:
1. the NSIS action had multiple updates, latest being 1.2.0. This should fix the issue you were seeing since this issue is closed: https://github.com/repolevedavaj/install-nsis/issues/23
2. As for the vcpkg change.. this will cause the cache to be lost. I'm not so sure if we still need @Lord-Kamina version of the `vcpkg-action` github action though since the [action](https://github.com/marketplace/actions/vcpkg-action) does have a cache key built in theses days - which wasn't the case before (or at least it wasn't working correctly). So if we do change this, then i'm probably in favor of running the latest version of that GH-Action. Since the only thing different in Kamina's version is using the binarysource flag which is the same for setting the VCPKG_DEFAULT_BINARY_CACHE variable
3. I noticed that we will default to another windows version somewhere in june - I'm not sure if we need any change in the windows workflow for this...

Lord-Kamina:
The action has has a cache key since forever, but the github cache is still unable to reuse keys. Normally, this means any change in the cache would result in a new cache being created, which soon consumes all the quota.

The vcpkg action now checks and saves cache only if there was a cache miss. So what this means is we'd always have 3-4 hour builds at least once a week.

I've been updating my fork of `actions/cache` periodically ever since I created that hack, because I don't lose hope it might someday get merged. I have also recently updated my fork of vcpkg-action, as it also had several updates recently.

P.S. in case you don't quite remember the hack I'm talking about: My forked cache is able to "refresh" caches by leveraging the API to delete a matched entry before saving it again, which effectively prevents it from expiring due to staleness as long as we build at least once a week.

ooshlablu:
ok, @Baklap4 @Lord-Kamina I did a few things.
* I pulled it back to how it is in the master branch
* Updated the vcpkg version, but it looks like `libsigcpp-3` doesn't exist anymore, so I removed it. I think the current version is that at this point anyway
* Looks like I had to change to `token` from `api-token` in @Lord-Kamina's  cache action
* Updated `repolevedavaj/install-nsis` to `1.2.0`

The appimage runs have been exploding for various inconsistent reasons during each run. They all seem like transient issues, so I will either re-run it later, or the squash should take care of it when I do that in a bit.l

ooshlablu:
I also removed the kitware repo from the appimage build. Not sure if we actually need it anymore, IIRC we needed it to install a usable version of `cmake` when the builder was Ubuntu 20.04, but it's been 24.04 for a while now.

Baklap4:
https://github.com/microsoft/vcpkg/pull/52130

hmmm it seems cpprestsdk is recently archived. It will not have updates anymore, in fact updating msvc with latest master will break probably our build.
If possible we should remove cpprestsdk from GHActions and build it ourselves - for the time being.

@Lord-Kamina you have an ancient PR involving another http library. This definitely now needs to happen lol

Lord-Kamina:
> [microsoft/vcpkg#52130](https://github.com/microsoft/vcpkg/pull/52130)
> 
> hmmm it seems cpprestsdk is recently archived. It will not have updates anymore, in fact updating msvc with latest master will break probably our build. If possible we should remove cpprestsdk from GHActions and build it ourselves - for the time being.
> 
> @Lord-Kamina you have an ancient PR involving another http library. This definitely now needs to happen lol

I do, and yes. One of the things that made that PR more complex than necessary was my trying to decouple the webserver search from the UI (and I tried to solve it by making a copy of the song list to be used for webserver searches, but I didn't consider the case of multiple people browsing at once)

So... what I think needs to happen, is refreshing that PR but nixing the above "solution" and after we merge it, I maintain the best solution would be to move the searching of songs client-side with something like FlexSearch. I don't think that would be that terrible to set-up properly.

Baklap4:
let's close this in favor of #1117 but that one should come with the appimage changes listed here

Baklap4:
closing in favor of #1117

PR #1117, opened 2026-05-22
Try to fix CI properly

Lord-Kamina:
Signed-off-by: Gregorio Litenstein <g.litenstein@gmail.com>

### More


.github/workflows/windows.yml
```diff
@@ -38,16 +38,16 @@ jobs:
...
           triplet: x64-windows-release
           extra-args: '--clean-after-build --overlay-triplets="${{ github.workspace }}\performous\cmake\triplets"'
           cache-key: win64-vcpkg
           debug: true
           # Get the sha from https://github.com/microsoft/vcpkg/releases
-          revision: d2452281016c6d77b6ef199d599d9929eafe4807
-          api-token: ${{ secrets.GITHUB_TOKEN }}
+          revision: 56bb2411609227288b70117ead2c47585ba07713
```
Baklap4: If you update to latest master: `d015e31e90838a4c9dfa3eed45979bc70d9357fc` it'll contain the fixes needed from cpprestsdk. No need to wait on a new 'vcpkg release'

Baklap4 (changes requested):

.github/workflows/windows.yml
Lord-Kamina: I guess using the current head is not much different from using a release, it's just I've had bad experiences in the fast using random commits. They've been pretty rigorously making one release a month for the past several months, so I was planning to wait out the week and go with the current head, should that trend be broken.

Baklap4:
https://github.com/microsoft/vcpkg/pull/52130

hmmm it seems cpprestsdk is recently archived. It will not have updates anymore, in fact updating vcpkg with latest master will break probably our build. Your commit (the latest relase) is still valid.
If possible we should remove cpprestsdk from GHActions and build it ourselves - for the time being.

@Lord-Kamina you have an ancient PR involving another http library. This definitely now needs to happen lol

Lord-Kamina:
@Baklap4 This is done, I think. No idea if @ooshlablu's PR has anything else that should be integrated here.

Baklap4:
if you can include the changes from the app-image aswell, we can close @ooshlablu 's pr, and get this one merged instead - which will also fix the mac builds once again

Subtle note: We should not update the vcpkg commit sha anymore (latest release is still fine), since we'll lose cpprestsdk as a result of that

Baklap4 (approved):

PR #1123, opened 2026-06-17
Start building Ubuntu 26.04 and Fedora 44 + boost:system is header only in 1.69+

ooshlablu:
### What does this PR do?

Add packages for Ubuntu 26.04 and Fedora 44. This required some cmake changes also because `boost::system` is only available as a header in boost 1.69+

### Motivation

They were released so they should be built :-)


Lord-Kamina (approved):
Looks OK to me.

PR #1125, opened 2026-07-20
add missing header

PPN-SD:
### What does this PR do?

Include fstream for std::ifstream / std::ofstream

### Motivation

It fixes compilation w/ gcc-17 that cleaned up transitive header


Lord-Kamina (approved):
LGTM

PR #1126, opened 2026-07-20
add PERFORMOUS_WERROR option and honor CMAKE_COMPILE_WARNING_AS_ERROR

PPN-SD:
### What does this PR do?

Make -Werror optional and honor [CMAKE_COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/prop_tgt/COMPILE_WARNING_AS_ERROR.html#prop_tgt:COMPILE_WARNING_AS_ERROR) if defined (e.g. forced by downstream)

### Motivation

-Werror is usually deactivated by linux distributions


Lord-Kamina:
We originally did it as a way to force us to be a little bit less sloppy, but I can see how people might want to turn it off.

One question, though, is why are you using different defaults for different platforms? I definitely don't mind having an easy way to turn it off, but think it should keep being on as default on every platform, and I'm positive @Baklap4 will agree because he uses MSVC and was the biggest advocate of our current approach.

PPN-SD:
> One question, though, is why are you using different defaults for different platforms?

Only because I didn’t want to change the current behavior.
I'm ok to set ON for all by default if it can be handled with CMAKE_COMPILE_WARNING_AS_ERROR

Lord-Kamina:
> > One question, though, is why are you using different defaults for different platforms?
> 
> Only because I didn’t want to change the current behavior. I'm ok to set ON for all by default if it can be handled with CMAKE_COMPILE_WARNING_AS_ERROR

Wait nevermind then... I thought we had it on in MSVC as well. Maybe it's because MSVC is so idiotic with its warnings?

PPN-SD:
I can't say anything relative to MSVC

Lord-Kamina:
> I can't say anything relative to MSVC

If this doesn't really change the current default behavior, it's ok by me. I would prefer that somebody else takes a look at it, however.

PR #1127, opened 2026-08-17
simplify/unify macos building:

taynaron:
remove Homebrew - didn't reliably work
build everything w/ macos15, to match github runners rely more on cmake, bring cmake build inline with windows/linux style

### What does this PR do?

* unifies, simplifies, and robustifies macOS building
  * unifies everything to macos 15 (inline with github runners) - had been parts at 15, 12, and some dependencies at 10.9
  * remove Homebrew - didn't reliably work, wasn't tested, was under-documented, and I wasn't able to get it working
  * rely more on cmake, bring cmake build inline with windows/linux style
  * a bit more error handling in the bundler
* updates docs, clarifying what works and some potential gotchas along the way

### Closes Issue(s)

None that I'm aware of, I'm just fixing an issue that's bitten me personally off and on since 2022.
Motivation

### Motivation

I've had trouble building Performous on MacOS for a while. Not that it didn't work, just that it was pretty fiddly and the docs weren't up to date with the state of things. I've tried fixing this in the past with just a docs update, but this time I am going for a more complete cleanup.
More

### More

- [x] Added/updated documentation

### Additional Notes


Lord-Kamina:
The updating of documentation to remove outdated gotchas, as well as finally removing the bash script are great.

However, I am right away telling you that there's a few things in here that I am ADAMANT about not merging.

1) I personally HATE homebrew. With a fucking passion you cannot begin to comprehend. That said, a lot of people use it. If you want to find and fix where it's going wrong? Please go ahead and do it; but I don't want to remove it.
2) The deployment target stays at 12.0, or at most can go to 13.0. My mac is running 13.0. What this target does is ensure the oldest version of macOS performous will run on.
3) CMake bundles aren't completely functional anyway. I can't quite remember exactly what issue I had with CPack but trust me I tried it when I made that script.
4) You're hardcoding dependency versions and paths in the CMake preset definition.

I don't want to use `cmake --build`. The intended workflow for debugging is to turn the `macos_bundler.py` with the debug switch, which creates a bundle but skips actually including and fixing all the dependencies.

taynaron:
1. I remember a conversation with you... years ago about hating Homebrew so I figured it'd be fine to get rid of it. Do you know someone who does use Homebrew to compile Performous? Maybe they could give me some pointers on getting it working, because I was unable. I can bring it back, but would sure like to be able to get it running to document how to do so.
1. The GitHub runners were all on macOS 15, so I figured it'd be fine to move forward. It seems reasonable to want to standardize on a single OS version across internal builds and GitHub builds. Do you think that should be 12 or 13? If not a single version, should we target both 12 and 26 and assume in-betweens will be OK?
1. FWIW the cmake build path worked for me just fine, and the builds here worked (the main reason I put up the draft PR - I was hoping to ask in chat before ruffling feathers here, just wanted to make sure it was more stable than "works on my machine). Would you be up for pulling this branch and seeing if it runs on your machine, after changing the two 15.0 strings to 12.0? I'm happy to rewrite the docs to focus on `macos_bundler.py` as the way to build, but it doesn't seem like it hurts to have the cmake options working too.
1. Ah, yeah, I'll fix that. That was more for testing.

Thanks for the feedback! I'm not looking to make big waves here, just clean up an area that I had assumed I was nearly alone in using :)

game/main.cc
```diff
@@ -335,6 +334,7 @@ int main(int argc, char** argv) {
 	}
 	PathCache::pathBootstrap();
 	SpdLogger spdLogger(levelEnum);
+	Platform platform;
```
taynaron, line 337: had to reorder this loading order to deal with a case where platform tried to use the logger in its constructor before it instantiated.

taynaron:
@Lord-Kamina I think we're at another discussion point here.

I have homebrew working! There's even a non-blocking CI check for homebrew working that currently passes.

I have split the CI paths so that macos15 runs on intel with macos12 as the minimum working version (as before), with a new arm64 runner for macos26. My thought is that the arm64 runner tracks with the latest macos until apple starts leaving the M1s behind.

`cmake --build` works, as does `macos_bundler.py`. I'd generally argue that letting cmake handle more, and ad-hoc scripts less, is better for maintainability. In the absence of more data I don't know how cmake has let you down in the past and can't build to fix it. I agree that building the dmg should be entirely in the bundler. Can you please see if this branch works for you, via both cmake and the bundler? Are you happy with the way this branch is now (at least from a code perspective, if not docs), or do you want me to remove some cmake functionality? Is there anything the bundler code does that you'd be willing to hand off to cmake, assuming I can get it working? What would make you convinced that it's working? If this isn't going to be an area of productive discussion, let me know and I'll drop it. :)

Anything else that sticks out to you / feels icky?

Thanks!

game/main.cc
Lord-Kamina: Good catch

Lord-Kamina:
Fuck. I'd written a large reply and lost it by virtue of being dumb.

In a nutshell:

- Target should be 13.0, as I _think_ all  (or MOST) macs that can run 12.0 can also do 13.0.
- I would keep target as the oldest that whichever SDK we're using on the runner supports.
- We don't need to do builds for every version and combination. We can get away with just a single Intel and ARM build targetting the older supported version as per the SDK specification. I need to have a better look at what you changed in the script or Homebrew install to get it to work, but I can run you through some of the hacks I was using.
- As I said, I want to keep homebrew support in the script (even though I personally hate Homebrew), but I don't want Homebrew anywhere near our CI, because that is indeed much more finnicky than the macports version and a successful CI run on homebrew doesn't really guarantee it will work for other environments. Also, it's gratuitous use of cache space and CI time.

I think I'm not leaving anything out.

Oh BTW, I DO like the idea of having the CI end with a run to check the build actually launches fine. I had proposed something similar a while ago in fact.

taynaron:
looks like we'd lose some 2013 and 2014 macs by moving the minimum version up from macos 12 to 13. I'm fine with either way for a minimum.

For arm macs it's fine to use that same minimum, but I figure since no arm machines have been left behind we could have a newer minimum for it. If you disagree we can keep it at 12 too.

Mostly the homebrew changes were to ignore the stated homebrew documentation and bang my head at dependencies for a while, then document the resulting pain. The code itself was fine.

I'm going to mark this as ready for proper review, so let's get it sorted for that. Thanks!

osx-utils/macos-bundler.py
```diff
@@ -80,57 +80,63 @@
...
+	for fmt_version in ["11", "10", "9", "8", "7"]:
+		check_fmt = check_installed_port(f"libfmt{fmt_version}", "fmt-config.cmake")
+		if check_fmt != None:
+			fmt_prefix = str(check_fmt.parent)
+			print(f"--- LibFMT {fmt_version} detected at: " + str(fmt_prefix) + "\n")
+			break
+
+def detect_homebrew():
+	global opencv_prefix, openssl_prefix, ffmpeg_prefix, icu_root, openssl_root
+	for opencv_version in ["4", "3"]:
```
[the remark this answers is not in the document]
taynaron: This is fixed, not sure why it's not marked as fixed here.

taynaron:
FWIW `cmake --build --preset macos-x64-debug && cmake --install build/macos-x64-debug` is way faster than `python macos-bundler.py --prefer-macports --debug` so I've moved to that for compiling when testing. Still need `macos-bundler` for final inspection, CI, and actually building and distributing, but cmake is speeding up the workflow significantly.

taynaron:
@Lord-Kamina sorry to be noisy, but is there anything you'd like changed here before an approval? Thanks!

.github/CONTRIBUTING.md
```diff
@@ -73,141 +73,138 @@ libwebp-devel fmt-devel
...
- -Wl,-headerpad_max_install_names" configure.ldflags="-m64 -arch x86_64 -stdlib=libc++ \ 
- -headerpad_max_install_names -mmacosx-version-min=10.9"
+sudo port install boost cairo cmake cpprestsdk dylibbundler ffmpeg7 \
+   fftw-3-single libfmt11 fontconfig freetype glm help2man icu libepoxy \
+   librsvg libsdl2 libxmlxx5 nlohmann-json opencv4 openssl pango \
+   portaudio portmidi
 ```
 
-and also these, in `~/.macports/macports.conf` (largely they're the same things as the commandline parameters, but Macports is sometimes... well, special, in deciding which things can be configured via command-line and which cannot.
+`ffmpeg7`, `libfmt11`, and `opencv4` are MacPorts' current major-version-suffixed ports for those
```
Lord-Kamina: Macports (and presumably homebrew as well) actually have ffmpeg8 and libfmt 12, as well.

.github/CONTRIBUTING.md
```diff
@@ -73,141 +73,138 @@ libwebp-devel fmt-devel
...
+libraries (MacPorts allows several ABI-incompatible major versions to coexist, so there's no
+version-less alias to install "whichever is newest"). If MacPorts has newer versions available, install whatever's current instead — CMake auto-detects whichever versioned port is installed (see `cmake/Modules/DarwinVersionedPortPrefixes.cmake`), so nothing else needs to change to match.
+
+Some of these (e.g. `ffmpeg7`) have no prebuilt archive for every macOS version and get compiled from source by MacPorts, which can pull in further source builds transitively. If one of those fails with errors like `fatal error: 'memory' file not found`/`'cstdint' file not found` for standard C/C++ headers, your Command Line Tools installation has stale leftover headers (a known MacPorts/CLT interaction — MacPorts prints its own warning about this during `clean`). In this case, just reinstall the CommandLineTools:
```
Lord-Kamina: Please make the following explicit: While using built-binaries is definitely the user's prerogative (and possibly homebrew might be better for that), the CI purposefully takes great care in building all of performous run-time dependencies from source, because that is the most fool-proof way to ensure a truly portable bundle.

.github/CONTRIBUTING.md
```diff
@@ -73,141 +73,138 @@ libwebp-devel fmt-devel
...
+cd osx-utils
+python3 -m venv ./bundler-venv
+source ./bundler-venv/bin/activate
+pip3 install -r ./macos-bundler-requirements.txt
+python3 ./macos-bundler.py --prefer-homebrew --debug
 ```
 
-Then follow the usual Performous build instructions. No special flags are required and you can do make install without sudo.
+One known Homebrew-specific issue this project works around: Homebrew's `fmt` package is often newer than this project's vendored `spdlog` expects, which used to cause `-Werror`/`-Wdeprecated-declarations` build failures. This is handled (`cmake/Modules/FindSpdlog.cmake` marks the vendored spdlog headers as system headers), so it shouldn't need any manual workaround. If you hit a similar deprecation error building against a very new Homebrew library, that's the class of issue to look for.
+
```
Lord-Kamina, line 204: As long as there aren't any breaking changes, feel free to update the vendored version of spdlog.

osx-utils/CMakeLists.txt
```diff
@@ -3,12 +3,32 @@ cmake_minimum_required(VERSION 3.19)
...
+	set(PERFORMOUS_MIN_OS_VERSION "${CMAKE_OSX_DEPLOYMENT_TARGET}")
+else()
+	message(WARNING "CMAKE_OSX_DEPLOYMENT_TARGET is not set; Info.plist's LSMinimumSystemVersion will be empty.")
+	set(PERFORMOUS_MIN_OS_VERSION "")
+endif()
+
+file(READ "${CMAKE_CURRENT_LIST_DIR}/resources/Info.plist.in" PERFORMOUS_INFO_PLIST_TEMPLATE)
+string(REPLACE "__PERFORMOUS_MIN_OS_VERSION__" "${PERFORMOUS_MIN_OS_VERSION}"
+	PERFORMOUS_INFO_PLIST_TEMPLATE "${PERFORMOUS_INFO_PLIST_TEMPLATE}")
+file(GENERATE
```
Lord-Kamina, line 22: I'm not sure I understand the different between `file(WRITE)` and `file(GENERATE)`, could you explain?

osx-utils/macos-bundler.py
```diff
@@ -232,18 +245,100 @@ def create_dmg(fancy: bool = True):
 		detach_retries=10
 	)
 
+def dedupe_rpaths():
+	# dylibbundler can write the same LC_RPATH entry into a dylib more than once when
+	# it's a shared transitive dependency of several bundled libraries. Recent dyld
+	# versions hard-reject duplicate LC_RPATH entries at load time ("Library not
```
Lord-Kamina, line 251: I had no idea about this, when did this begin happening?

osx-utils/macos-bundler.py
```diff
@@ -232,18 +245,100 @@ def create_dmg(fancy: bool = True):
...
 	execute(fr"""
 		dylibbundler -od -b \
 		-x "{performous_out_dir}/Performous.app/Contents/MacOS/Performous" \
 		-d "{performous_out_dir}/Performous.app/Contents/Resources/lib" \
 		-p @executable_path/../Resources/lib/
 	""")
+	dedupe_rpaths()
 	return
 
+def verify_bundle():
```
Lord-Kamina, line 284: Have any of those things actually happened to you? In several years, the only thing I'd occasionally had issues with was possible stale files. I appreciate the effort into adding all these checks, but personally I don't think they should be tied to the debug mode. Rather, put it under its own command-line flag, and just run it during CI if you consider it important.

CMakePresets.json
```diff
@@ -51,6 +51,28 @@
...
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "condition": {
+        "type": "equals",
+        "lhs": "${hostSystemName}",
+        "rhs": "Darwin"
+      },
+      "cacheVariables": {
+        "CMAKE_PREFIX_PATH": "/opt/local",
+        "PKG_CONFIG_USE_CMAKE_PREFIX_PATH": "ON",
+        "CMAKE_OSX_DEPLOYMENT_TARGET": "15.0",
```
Lord-Kamina: I'd take have the deployment target for both the intel and arm64 builds match, at 13.0
Granted, it's very much an edge-case for someone on arm64 to not want to update to the latest macOS version, but there's really no tangible advantage to gatekeeping it with a higher target.

Lord-Kamina (changes requested):
A couple things still.

osx-utils/CMakeLists.txt
taynaron: `file(WRITE` always makes a new file from scratch, `file(GENERATE` only makes a new one if cmake detects changes. It helps prevent rebuilding targets with no changes, speeding up compiling.

.github/CONTRIBUTING.md
taynaron: The most recent version of spdlog still has the isue that was causing the warning. Let me tweak the verbiage here to clarify.

osx-utils/macos-bundler.py
taynaron: I only encountered it when running macos26. I expect it's related to apple's quest to make all its tools fussier about security stuff. Regardless, I actually hit this (both the rpaths and the signing issues) when working on this.

osx-utils/macos-bundler.py
taynaron: I did have these happen to me locally while I was working on this, which is why I added them. I agree that they're too heavy for debug mode, but it seems reasonable to me that anyone who is building the app for real may want it to be portable, and therefore we should maximize the likelihood that it will be. Since I actually hit these issues others could too.

If you still want it behind a separate flag, I'm happy to do that, but I don't see the harm in the extra checks, and they did help me out.

.github/CONTRIBUTING.md
taynaron: updated, thanks!

.github/CONTRIBUTING.md
taynaron: added!

CMakePresets.json
taynaron: fair enough. Fixed.

osx-utils/CMakeLists.txt
Lord-Kamina: Ahh. I see!

osx-utils/macos-bundler.py
Lord-Kamina: I'd rather still have it as a separate flag, because I do occasionally compile local release builds on the rare occasions when I actually want to play.

taynaron: updated!

taynaron:
@Lord-Kamina all comments addressed, take another look please?

Lord-Kamina (approved):

taynaron:
Thanks @Lord-Kamina ! Can you please also diskiss the [gh-bot comment](https://github.com/performous/performous/pull/1127/changes#r3808898990) ? It mentioned the issue twice, and only closed one of them when I fixed it.

PR #538, opened 2020-04-16
let game to memorize player's names selection

allcreater:
### What does this PR do?
matches player controller's name with player name at high scores screen

### Closes Issue(s)
#61 

### Motivation
To save players from routine searching his/her name at names list
Two or more players usually uses their favorite controllers so that game should remember player's name selection for that controller.

### More

- [ ] small refactoring: SongItems::lookup(std::shared_ptr<Song> song) removed because it responsibilities could be taken by SongItems::lookup(Song& song) 

### Additional Notes

one of my experimental crafts, not sure is it good way to do so but will be glad to discuss :)

Issue #61, opened 2013-08-26
Remember player settings


Tronic:
Player names should be remembered per controller so that if the same group of people plays multiple songs in succession, correct player names are already chosen by default for the high score.

PR #1129, opened 2026-09-09
Remember player high-score settings

taynaron:
### What does this PR do?

This cleans up and updates https://github.com/performous/performous/pull/538 which was a great idea, just never fully merged. Devices should remember the last-used player on the high-score screen.

Along the way I found that user creation and scrolling was actually broken for new players, and perhaps all players. Bad null check.

### Closes Issue(s)

https://github.com/performous/performous/issues/61
PR https://github.com/performous/performous/pull/538

### Motivation

Fix the bug that was preventing high scores from being entered for new players
Make high-scores a bit easier to deal with

### DIFF
--- a/game/database.cc
+++ b/game/database.cc
@@ -68,16 +68,17 @@ void Database::addHiscore(std::shared_ptr<Song> s) {
 		return;
 	}
 	auto playerid = maybe_playerid.value();
-	unsigned score = scores.front().score;
-	std::string track = scores.front().track;
+	ScoreItem const& hiscore = scores.front();
 	const auto songid = m_songs.lookup(s);
 	if(!songid.has_value()) {
 		SpdLogger::error(LogSystem::DATABASE, "Invalid song ID for song: artist={}, title={}", s->artist, s->title);
 		return;
 	}
 	unsigned short level = config["game/difficulty"].ui();
-	m_hiscores.addHiscore(score, playerid, songid.value(), level, track);
-	SpdLogger::info(LogSystem::DATABASE, "Added new hiscore. Score={} on track={} for song id={}, on level={}", score, track, songid.value(), level);
+	m_hiscores.addHiscore(hiscore.score, playerid, songid.value(), level, hiscore.track);
+	// Remember which player was selected for this score's source device, so it can be pre-selected next time.
+	if (!hiscore.player_id.empty()) playersByDevices[hiscore.player_id] = playerid;
+	SpdLogger::info(LogSystem::DATABASE, "Added new hiscore. Score={} on track={} for song id={}, on level={}", hiscore.score, hiscore.track, songid.value(), level);
 }
 
 bool Database::reachedHiscore(std::shared_ptr<Song> s) const {
@@ -142,4 +143,3 @@ std::vector<HiscoreItem> Database::getHiscores(SongPtr const& s) const {
 		throw;
 	}
 }
-
--- a/game/database.cc
+++ b/game/database.cc
@@ -81,6 +81,15 @@ void Database::addHiscore(std::shared_ptr<Song> s) {
 	SpdLogger::info(LogSystem::DATABASE, "Added new hiscore. Score={} on track={} for song id={}, on level={}", hiscore.score, hiscore.track, songid.value(), level);
 }
 
+std::optional<PlayerId> Database::rememberedPlayerForDevice(std::string const& deviceId) const {
+	if (deviceId.empty()) return std::nullopt;
+
+	auto const it = playersByDevices.find(deviceId);
+	if (it == playersByDevices.end()) return std::nullopt;
+
+	return it->second;
+}
+
 bool Database::reachedHiscore(std::shared_ptr<Song> s) const {
 	unsigned score = scores.front().score;
 	const auto track = scores.front().track;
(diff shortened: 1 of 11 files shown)


game/players.cc
```diff
@@ -96,10 +123,10 @@ void Players::setFilter(std::string const& val) {
 
 PlayerId Players::assign_id_internal() {
 	const auto it = std::max_element(m_players.begin(),m_players.end());
-	
-	if (it != m_players.end() && it->id)
```
taynaron, line 100: This null check was failing us because if players exist, the first player is id 0. Therefore this check always failed.

taynaron:
The windows failure seems unrelated to the code changes, it's a dl failure

game/players.cc
kt--: How does this bug trigger?  I was able to add new player names OK.

Sure the null check passing through is an issue, I'm not questioning that.

But is it safe to simply remove the ID check?  I'm sitting here wondering if there's another issue around the player ID being `0` as well.  Maybe we _need_ that ID check too? 

Remembering the player's controller would be a good feature to have.

taynaron: I had a setup with one pre-existing player when starting performous. At the end of a song, I tried to add another player and was unable to do so. If you can't replicate this I'm happy to look into it more. 

The bug was introduced back in 2022 with the PR "Use std::optional instead of -1 in database lookups" - Performous used to use -1 as a special "empty DB" variable state, that PR moved to a better model, but introduced this bug in 2 places, one of which was caught almost immediately. There's nothing special about ID 0, and no reason to keep this check. It was worth checking though.

kt--: Sounds good.

I'm a bit torn on std::optional.  Sure it's better than having a raw pointer, but underneath it's still a pointer  with all the pointer pitfalls.  But when a "no record", or "enabled -> false" is removed and replaced with a std::optional that _adds_ a potential null pointer - I don't think it's something to casually introduce.

taynaron: I'm still catching up on c++ - the last time I seriously coded in it was 2013. std::optional seemed like an improvement over bare pointers. If you disagree, then what would you suggest as an improvement here?

kt--: Heh - I don't have one.  I agree `std::optional` _is_ an improvement over bare pointers.

Whether it's best practice to replace an existing and working "integer result of -1" Vs "std::optional null pointer" is what I'm undecided about.  But that's not part of this change anyway.

game/database.cc
```diff
@@ -62,22 +62,39 @@ void Database::addSong(std::shared_ptr<Song> s) {
 }
 
 void Database::addHiscore(std::shared_ptr<Song> s) {
-	auto maybe_playerid = m_players.lookup(m_players.current().name);
+	std::string const& currentName = m_players.current().name;
+	if (currentName.empty()) {
```
kt--, line 67: I think adding a trim of leading and following space here, before checking for blank would be good.

One of the high-scores on my TV is owned by _blank_.  So I have an axe to grind ;)

game/database.cc
```diff
@@ -62,22 +62,39 @@ void Database::addSong(std::shared_ptr<Song> s) {
...
 	if(!songid.has_value()) {
 		SpdLogger::error(LogSystem::DATABASE, "Invalid song ID for song: artist={}, title={}", s->artist, s->title);
 		return;
 	}
 	unsigned short level = config["game/difficulty"].ui();
-	m_hiscores.addHiscore(score, playerid, songid.value(), level, track);
-	SpdLogger::info(LogSystem::DATABASE, "Added new hiscore. Score={} on track={} for song id={}, on level={}", score, track, songid.value(), level);
+	m_hiscores.addHiscore(hiscore.score, playerid, songid.value(), level, hiscore.track);
+	// Remember which player was selected for this score's source device, so it can be pre-selected next time.
+	if (!hiscore.player_id.empty()) playersByDevices[hiscore.player_id] = playerid;
```
kt--, line 88: I love this.

game/database.hh
```diff
@@ -61,11 +61,14 @@ public:
 private: // will be bypassed by above friend declaration
 	typedef std::list<Player> cur_players_t;
 	typedef std::list<ScoreItem> cur_scores_t;
+	typedef std::map <std::string, PlayerId> players_devices_t;
```
kt--, line 64: Nitpick:  A bunch of static analysis tools will whine about adding `_t` to your own structs & classes.
Not an issue for me though.  I don't remember if this is a Performous convention.

game/players.cc
```diff
@@ -165,8 +192,13 @@ void Players::advance(std::ptrdiff_t diff) {
 	math_cover.setTarget(current, count());
 }
 
+void Players::advanceToId(PlayerId id) {
```
kt--, line 197: Please add a quick comment on this, it's not immediately clear what it's doing.

kt-- (changes requested):
I'm generally happy with this change.

Please add some notes on how to test & verify to the PR description.  I want to be able to observe the existing issue, apply the patch, and observe it's gone.

game/database.hh
taynaron: official or not, it's all throughout the codebase, so I followed suit.

game/database.cc
taynaron: Elsewhere in this PR, truly empty names are converted to "Player {id}". Trim added.

game/players.cc
taynaron: done

kt-- (approved):

